PR #42, opened 2018-08-24
Extended Puzzle Props

tabascq:
I'm removing a couple of puzzle properties that can't be cheap to
compute (they modify scoring based on the number of solves at the time
of the solve - we have not used these in years) and I'm adding several
properties that used to be in the extended bucket, as well as a
prerequisites table that we can use for dependencies.


.vs/config/applicationhost.config
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="UTF-8"?>
+<?xml version="1.0" encoding="UTF-8"?>
```
lambertwang: This file should be in the .gitignore.

tabascq: Strangely, the entire .vs folder *is* in the .gitignore. I suppose I can pull this file from the review anyways but it would be great if .gitignore would, you know, ignore.

tabascq: Related: Is there any reason we are not using the Visual Studio .gitignore that comes with a new GitHub repo? It's way more detailed.

asyasky: Because it wasn't generated when I created the project? Let's try the VS one and see if that solves the problem, since I've been seeing the same thing and I just keep having to stage changes around it.

asyasky: Oh, I see, this isn't a new check in, it's a change. So first we need to delete this file from the repo. That will likely just fix the problem.

tabascq: Yep, just added a commit to do that.

Data/DataModel/Prerequisites.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations.Schema;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace ServerCore.DataModel
+{
+    public class Prerequisites
```
asyasky: Can you add comments to this? I had to stop and think about what a prerequisite would mean and what this is, which makes a good candidate for comments. Xml comments are best since then Intellisense knows all the things. (I'm aware that a lot of the data model doesn't have comments yet, it's a work in progress.)

Data/DataModel/Puzzle.cs
```diff
@@ -13,9 +13,34 @@ public class Puzzle
...
+        /// <summary>
+        /// Reward if solved: Sometimes displayed publicly, sometimes used internally by meta engine
+        /// </summary>
+        public string Token { get; set; }
+
+        /// <summary>
+        /// Puzzlehunt: module
+        /// Puzzleday: pregame
+        /// </summary>
+        public string Group { get; set; }
```
asyasky: I'd love to have this be an enum of some kind so it can be a drop down in the UI (and avoid string comparisons), maybe pulled from the database. I'll look into what it takes to get a list more strongly typed here (while still having it vary event to event).

Data/DataModel/Prerequisites.cs
tabascq: Sure, will do.

Data/DataModel/Puzzle.cs
tabascq: My comment must be unclear - for Puzzlehunt the Group would be the *name of the module*. So it can't be an enum, though it could be a link to a Groups table if we want to make one. I felt that was overkill though.

tabascq:
Happy to help - but I think it is a stretch to say I added the unlocking *logic* here...

Data/DataModel/Puzzle.cs
tabascq: Updated comment to clarify.

tabascq:
I am not sure how I am supposed to say this is ready to re-review other than this:

"This is ready for re-review"

Data/DataModel/Puzzle.cs
asyasky: Oh, no, it was clear. I'm wondering if we can pull the list of module names from the db or something similar so that every author doesn't have to put in their module name by hand for every puzzle. And I like things to be strongly typed, but thinking about it more I think filling a collection of strings from the db is plenty to avoid the problems I want to avoid.

PR #47, opened 2018-08-29
Event-based Puzzles

tabascq:
Making puzzles be owned by events, as a way to play with object
relationships.

The "CLICK HERE TO CHECK YOUR DATABASE SETUP" link now points to Events
instead of Puzzles, and the Events index has a new link to access that
event's puzzle list.

I've needed to pass the event ID around, and the event ID should
probably be part of some global context, but this is a helpful microcosm
of issues we'll have when doing other features so I still think it's a
good start.

### DIFF
--- a/ServerCore/Pages/Events/Index.cshtml
+++ b/ServerCore/Pages/Events/Index.cshtml
@@ -19,6 +19,7 @@
                 <th>
                     @Html.DisplayNameFor(model => model.Event[0].UrlString)
                 </th>
+            <!--
                 <th>
                     @Html.DisplayNameFor(model => model.Event[0].MaxNumberOfTeams)
                 </th>
@@ -28,9 +29,11 @@
                 <th>
                     @Html.DisplayNameFor(model => model.Event[0].MaxExternalsPerTeam)
                 </th>
+            -->
                 <th>
                     @Html.DisplayNameFor(model => model.Event[0].IsInternEvent)
                 </th>
+            <!--
                 <th>
                     @Html.DisplayNameFor(model => model.Event[0].TeamRegistrationBegin)
                 </th>
@@ -70,6 +73,7 @@
                 <th>
                     @Html.DisplayNameFor(model => model.Event[0].AllowFeedback)
                 </th>
+            -->
             <th></th>
         </tr>
     </thead>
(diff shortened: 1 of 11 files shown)


ServerCore/Pages/Events/Index.cshtml
```diff
@@ -19,6 +19,7 @@
                 <th>
                     @Html.DisplayNameFor(model => model.Event[0].UrlString)
                 </th>
+            <!--
```
morganbr: Please remove commented out code.

ServerCore/Pages/Puzzles/Create.cshtml.cs
```diff
@@ -27,17 +28,21 @@ public IActionResult OnGet()
         [BindProperty]
         public Puzzle Puzzle { get; set; }
 
-        public async Task<IActionResult> OnPostAsync()
+        public async Task<IActionResult> OnPostAsync(int? eventid)
```
morganbr: Should we require having an event id? We probably don't want puzzles that aren't in events

ServerCore/Pages/Puzzles/Create.cshtml.cs
```diff
@@ -27,17 +28,21 @@ public IActionResult OnGet()
...
 
+            if (eventid != null)
+            {
+                Puzzle.Event = await _context.Event.SingleOrDefaultAsync(m => m.ID == eventid);
+            }
             _context.Puzzle.Add(Puzzle);
             await _context.SaveChangesAsync();
 
-            return RedirectToPage("./Index");
+            return RedirectToPage("./Index", new { eventid = eventid });
```
morganbr: FYI @asyasky, take a look at how we should be doing this (but no need to block on this)

ServerCore/Pages/Puzzles/Index.cshtml.cs
```diff
@@ -19,11 +19,21 @@ public IndexModel(ServerCore.Models.PuzzleServerContext context)
...
-        public IList<Puzzle> Puzzle { get;set; }
+        public IList<Puzzle> Puzzle { get; set; }
 
-        public async Task OnGetAsync()
+        public int? EventId { get; set; }
+
+        public async Task OnGetAsync(int? eventid)
         {
-            Puzzle = await _context.Puzzle.ToListAsync();
+            if (eventid != null)
```
morganbr: Please file an issue tracking removing the null tolerance since we don't want people accidentally wandering into all puzzles in all events

tabascq:
Changes made, thanks!

PR #57, opened 2018-09-07
Routing and event awareness cleanup

tabascq:
I've shrunk the work required for event awareness down to the very
small.

1. When declaring event-specific pages, derive page models from
ServerCore.ModelBases.EventSpecificModelBase. This adds an Event
property of type Event, and the menus will use it. Example:

public class DetailsModel : EventSpecificModelBase

2. Change the @page attribute to provide a specific routing path, e.g.
this one for puzzle details:

@page "{eventId}/Puzzles/Details/{id}"

[Note: we can probably remove this eventually, in favor of routing
conventions.]

3. When navigating from a page not in the event specific zone, add an
asp-route-eventId parameter, like this one in Events/Index that uses the
ID of the chosen event:

<a asp-page="/Puzzles/Index" asp-route-eventId="@item.ID">

That is currently the only place that asp-route-eventId is needed,
so these should be rare.

### DIFF
--- a/ServerCore/ModelBases/EventSpecificPageModel.cs
+++ b/ServerCore/ModelBases/EventSpecificPageModel.cs
@@ -1,0 +1,38 @@
+﻿using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+using Microsoft.AspNetCore.Mvc.RazorPages;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.DependencyInjection;
+using ServerCore.DataModel;
+using ServerCore.Models;
+
+namespace ServerCore.ModelBases
+{
+    public abstract class EventSpecificPageModel : PageModel
+    {
+        [FromRoute]
+        [ModelBinder(typeof(EventBinder))]
+        public Event Event { get; set; }
+
+        public class EventBinder : IModelBinder
+        {
+            public async Task BindModelAsync(ModelBindingContext bindingContext)
+            {
+                string eventIdAsString = bindingContext.ActionContext.RouteData.Values["eventId"] as string;
+
+                if (int.TryParse(eventIdAsString, out int eventId))
+                {
+                    var puzzleServerContext = bindingContext.HttpContext.RequestServices.GetService<PuzzleServerContext>();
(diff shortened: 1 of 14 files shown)


ServerCore/ModelBases/EventSpecificPageModel.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        [FromRoute]
+        [ModelBinder(typeof(EventBinder))]
+        public Event Event { get; set; }
+
+        public class EventBinder : IModelBinder
+        {
+            public async Task BindModelAsync(ModelBindingContext bindingContext)
+            {
+                string eventIdAsString = bindingContext.ActionContext.RouteData.Values["eventId"] as string;
```
morganbr: Should we consider adding logic to translate from a string like "PuzzleHunt2019" to the id here? That would give us friendly urls instead of having to find/translate database ids.

ServerCore/ModelBases/EventSpecificPageModel.cs
```diff
@@ -0,0 +1,38 @@
...
+
+                if (int.TryParse(eventIdAsString, out int eventId))
+                {
+                    var puzzleServerContext = bindingContext.HttpContext.RequestServices.GetService<PuzzleServerContext>();
+                    Event eventObj = await puzzleServerContext.Events.Where(e => e.ID == eventId).FirstOrDefaultAsync();
+
+                    if (eventObj != null)
```
morganbr: If the lookup failed, what happens? I guess I'd hope for a 404

ServerCore/ModelBases/EventSpecificPageModel.cs
tabascq: Reading my mind here. I want to add another property to Event like "ShortName" and key off that instead of ID. Shortname examples our existing site uses: 19, 2018.5, etc.

I just didn't want to add a database change to this PR as well.

ServerCore/ModelBases/EventSpecificPageModel.cs
tabascq: So far no such luck. I can't figure out how to have the property be required.

tabascq: By which I mean that if I set the event ID to be out of scope, it just merrily continues along with Event == null. The binding result is {Failed}, but the model doesn't care.

PR #58, opened 2018-09-08
Add Teams Pages

tabascq:
Adding pages for Teams, to prepare for linking puzzles to per-team
state. Only needed to tweak the @page attribute and the base class in
order to get it to be Event-specific.


ServerCore/Pages/Teams/Create.cshtml
```diff
@@ -0,0 +1,64 @@
...
+    <div class="col-md-4">
+        <form method="post">
+            <div asp-validation-summary="ModelOnly" class="text-danger"></div>
+            <div class="form-group">
+                <label asp-for="Team.Name" class="control-label"></label>
+                <input asp-for="Team.Name" class="form-control" />
+                <span asp-validation-for="Team.Name" class="text-danger"></span>
+            </div>
+            <div class="form-group">
+                <label asp-for="Team.Password" class="control-label"></label>
```
morganbr: @asyasky, we're not supposed to have team passwords, right?

ServerCore/Pages/Teams/Create.cshtml.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        public async Task<IActionResult> OnPostAsync()
+        {
+            if (!ModelState.IsValid)
+            {
+                return Page();
+            }
+
+            Team.Event = Event;
+            Team.PuzzleCacheLastUpdated = DateTime.UtcNow;
```
morganbr: Puzzle cache is a somewhat suspect holdover from the old system. I'd suggest ditching it.

ServerCore/Pages/Teams/Delete.cshtml
```diff
@@ -0,0 +1,64 @@
...
+    <hr />
+    <dl class="dl-horizontal">
+        <dt>
+            @Html.DisplayNameFor(model => model.Team.Name)
+        </dt>
+        <dd>
+            @Html.DisplayFor(model => model.Team.Name)
+        </dd>
+        <dt>
+            @Html.DisplayNameFor(model => model.Team.Password)
```
morganbr: Do we actually need to display anything other than name on this page?

ServerCore/Pages/Teams/Create.cshtml.cs
```diff
@@ -0,0 +1,42 @@
...
+        public Team Team { get; set; }
+
+        public async Task<IActionResult> OnPostAsync()
+        {
+            if (!ModelState.IsValid)
+            {
+                return Page();
+            }
+
+            Team.Event = Event;
```
morganbr: It might be good to prevent multiple teams from having the same name in the same event. This could be a good chance to set the pattern for input validation.

ServerCore/Pages/Teams/Delete.cshtml.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            _context = context;
+        }
+
+        [BindProperty]
+        public Team Team { get; set; }
+
+        public async Task<IActionResult> OnGetAsync(int? id)
+        {
+            if (id == null)
```
morganbr: if you make id non-nullable, will the NotFound get handled automatically by the path not matching?

Data/DataModel/PuzzleStatePerTeam.cs
```diff
@@ -16,7 +16,45 @@ public class PuzzleStatePerTeam
 
         public virtual Puzzle Puzzle { get; set; }
         public virtual Team Team { get; set; }
-        public virtual State State { get; set; }
-        public DateTime StateChanged { get; set; }
+
+        /// <summary>
```
morganbr: @asyasky, what motivated you to have this stuff in a separate table?

tabascq:
OK, will wait for @asyasky to review in order to keep the number of migrations low.

ServerCore/Pages/Teams/Create.cshtml.cs
tabascq: This is beyond my current skill set. I mean I know how to tell that it's a duplicate but I don't know how to communicate that on the web page through validation.

ServerCore/Pages/Teams/Delete.cshtml
tabascq: The CRUD wizard generates all of this and I decided to leave it in for now.

ServerCore/Pages/Teams/Create.cshtml.cs
tabascq: If @asyasky has another plan here I am all for it.

ServerCore/Pages/Teams/Delete.cshtml.cs
tabascq: Yes - the CRUD wizard adds this, and we can clean it up.

ServerCore/Pages/Teams/Create.cshtml.cs
asyasky: What exactly is it caching? (puzzle list, puzzle state, something else?)

Data/DataModel/PuzzleStatePerTeam.cs
asyasky: Originally PuzzleStatePerTeam was a join table with a timestamp. I was keeping that pattern when I re-created it. Let me thing about what makes the most sense in this case and then we can have a brief design discussion.

ServerCore/Pages/Teams/Create.cshtml.cs
asyasky: https://docs.microsoft.com/en-us/ef/core/modeling/relational/unique-constraints

morganbr: Ok, don't treat this as blocking. @asyasky, maybe you can think about how we should deal with this? A unique constraint would keep the DB from accepting it, but wouldn't' be user-friendly (it would probably just throw an exception when you try to put the data in)

ServerCore/Pages/Teams/Create.cshtml.cs
morganbr: I think it was caching something puzzle list and state related

Data/DataModel/PuzzleStatePerTeam.cs
tabascq: I feel like the PuzzleStatePerTeam and State tables had a 1:1 mapping, and if so I don't see the advantage in having separate tables.

ServerCore/Pages/Teams/Create.cshtml.cs
tabascq: Sounds like there are no plans for this. I will remove when the rest of the issues are resolved.

ServerCore/Pages/Teams/Create.cshtml
tabascq: OK, will remove when the rest of the issues are resolved.

ServerCore/Pages/Teams/Create.cshtml.cs
tabascq: The code sample in that link does not appear to follow the conventions described immediately above it - either that or I just don't understand the terms well enough. I'm going to pass on this for now, we can discuss in person.

Data/DataModel/PuzzleStatePerTeam.cs
tabascq: Just so we're clear: I am blocked on this issue. What I've done makes sense to me but I'm not a DB person...

tabascq:
All planned changes made - any further comments or are we done here?

PR #60, opened 2018-09-11
Jessica/puzzle answer submissions

digit01Wave:
Add response and submission tables. 

Made It so submission actions are done in only one page.
Submission page does the following:
1. Allow for answer submissions to puzzles
2.  Submissions are responded to with the appropriate response
3. Answers are locked after the correct answer gets inputted


ServerCore/Pages/Responses/Create.cshtml.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.RazorPages;
+using Microsoft.EntityFrameworkCore;
+using ServerCore.DataModel;
+
+namespace ServerCore.Pages.Responses
+{
+    public class CreateModel : PageModel
```
tabascq: You're going to want to make this page (and all pages you add) Event-aware; it will standardize things and get the menu working properly.

https://github.com/PuzzleServer/mainpuzzleserver/wiki/Making-a-page-Event-aware

Data/DataModel/Response.cs
```diff
@@ -1,18 +1,53 @@
...
+        /// </summary>
         public bool IsSolution { get; set; }
-        public string SubmittedText { get; set; }
+
+        /// <summary>
+        /// The submitted text
+        /// </summary>
+        public string SubmittedText
+        {
+            get { return this.submittedText; }
```
tabascq: This pattern doesn't cause EF to freak out? [Note: I have no EF experience, just trying to learn what the boundaries are]

ServerCore/Pages/Puzzles/Index.cshtml
```diff
@@ -98,9 +98,11 @@
                 @Html.DisplayFor(modelItem => item.MaterialsUrlString)
             </td>
             <td>
-                <a asp-Page="./Edit" asp-route-id="@item.ID">Edit</a> |
-                <a asp-Page="./Details" asp-route-id="@item.ID">Details</a> |
-                <a asp-Page="./Delete" asp-route-id="@item.ID">Delete</a>
+                <a asp-page="./Edit" asp-route-id="@item.ID">Edit</a> |
+                <a asp-page="./Details" asp-route-id="@item.ID">Details</a> |
+                <a asp-page="./Delete" asp-route-id="@item.ID">Delete</a> |
+                <a asp-page="/Responses/Index" asp-route-puzzleid="@item.ID" asp-route-eventid="@item.Event.ID">Responses</a> |
```
tabascq: When you make the Responses pages Event-aware (see link in another comment), you shouldn't need to respecify the eventid here. It should transfer just fine.

ServerCore/Pages/Responses/Delete.cshtml.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        public DeleteModel(ServerCore.Models.PuzzleServerContext context)
+        {
+            _context = context;
+        }
+
+        [BindProperty]
+        public Response PuzzleResponse { get; set; }
+
+        public async Task<IActionResult> OnGetAsync(int? id)
```
tabascq: We're modifying these pages to have the parameters not be optional, unless they really really are. The NotFound case on line 27 will then be automatically handled.

Data/DataModel/Response.cs
```diff
@@ -1,18 +1,53 @@
...
+            set { this.submittedText = FormatSubmission(value); }
+        }
+
```
tabascq: Where is the actual response? "Correct", "Keep going", etc.

ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -0,0 +1,84 @@
...
+            {
+                <tr>
+                    <td>
+                        @Html.DisplayFor(modelItem => item.TimeSubmitted)
+                    </td>
+                    <td>
+                        @Html.DisplayFor(modelItem => item.SubmissionText)
+                    </td>
+                    <td>
+                        @(item.Response != null ? item.Response.Note : "Incorrect")
```
tabascq: This isn't the response, at least historically it isn't. It's extra notes that the author can tag things with, for author/admin use only. If we are going to have this be the actual response let's name it better.


Note that I have never seen the Note field actually used in the old DB, so it's probably OK to repurpose this field - let's just pick a name that better matches the purpose.

ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -0,0 +1,79 @@
...
+            List<Response> responses = await _context.Responses.Where(r => r.Puzzle.ID == puzzleId && Submission.SubmissionText == r.SubmittedText).ToListAsync();
+            Submission.Response = responses.Count == 1 ? responses.First() : null;
+
+            _context.Submissions.Add(Submission);
+            await _context.SaveChangesAsync();
+
+            return RedirectToPage("./Index", new { puzzleid = puzzleId, eventid = Submission.Puzzle.Event.ID });
+        }
+
+        public async Task OnGetAsync(int? puzzleId, int? eventId)
```
tabascq: This page needs some team awareness. Maybe link to it from the Teams index so you can supply a team ID?

ServerCore/Pages/Responses/Create.cshtml.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            _context = context;
+        }
+
+        [BindProperty]
+        public Response PuzzleResponse { get; set; }
+
+        public int PuzzleId { get; set; }
+
+        public int EventId { get; set; }
```
tabascq: Note: Event-aware pages have an Event property you can use.

Data/DataModel/Response.cs
digit01Wave: The response seems to be the Note

Data/DataModel/Response.cs
digit01Wave: Not that I know of.

Data/DataModel/Response.cs
tabascq: Compare with the old site at https://puzzleday.research.microsoft.com/2018/Author/SubmissionResponses.aspx

You will see columns named "Submission" and "Response" with, well, the submission and response, and another column named "Note" that is empty.

ServerCore/Pages/Responses/Create.cshtml.cs
```diff
@@ -20,28 +21,20 @@ public CreateModel(ServerCore.Models.PuzzleServerContext context)
...
         public IActionResult OnGet(int puzzleId, int eventId)
         {
             this.PuzzleId = puzzleId;
-            this.EventId = eventId;
             return Page();
         }
 
         public async Task<IActionResult> OnPostAsync(int puzzleId)
         {
-            if (!ModelState.IsValid)
```
tabascq: Other Create pages keep this validity check - why did you remove it here?

ServerCore/Pages/Responses/Delete.cshtml
```diff
@@ -31,10 +31,10 @@
             @Html.DisplayFor(model => model.PuzzleResponse.Note)
         </dd>
     </dl>
-    
+
     <form method="post">
         <input type="hidden" asp-for="PuzzleResponse.ID" />
         <input type="submit" value="Delete" class="btn btn-default" /> |
-        <a asp-page="./Index">Back to List</a>
+        <a asp-page="./Index">Back to Puzzle List</a>
```
tabascq: But this appears to go back to the list of responses? Doesn't it navigate to /{eventId}/Responses/Index?

ServerCore/Pages/Responses/Delete.cshtml.cs
```diff
@@ -21,15 +17,13 @@ public DeleteModel(ServerCore.Models.PuzzleServerContext context)
 
         [BindProperty]
         public Response PuzzleResponse { get; set; }
+        
+        public int PuzzleId { get; set; }
 
-        public async Task<IActionResult> OnGetAsync(int? id)
+        public async Task<IActionResult> OnGetAsync(int id, int puzzleId)
```
tabascq: Why do you need a puzzle ID here? It's in Response.Puzzle.ID.

tabascq: Ditto other pages where this is now a parameter.

ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -23,19 +23,12 @@ public IndexModel(ServerCore.Models.PuzzleServerContext context)
...
-        public int? PuzzleId { get; set; }
-
-        public int? EventId { get; set; }
+        public int PuzzleId { get; set; }
 
         public string AnswerToken { get; set; }
 
         public async Task<IActionResult> OnPostAsync(int puzzleId)
         {
-            if (!ModelState.IsValid)
```
tabascq: Still not sure why validation checks are being removed

Data/DataModel/Response.cs
morganbr: @asyasky says the standard for fields should be camel case with no 'this.' In this case, that would be: ```get { return submittedText; }```

ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        public string AnswerToken { get; set; }
+
+        public async Task<IActionResult> OnPostAsync(int puzzleId)
+        {
+            Submission.TimeSubmitted = DateTime.Now;
+            Submission.Puzzle = await _context.Puzzles.SingleOrDefaultAsync(p => p.ID == puzzleId);
+
+            List<Response> responses = await _context.Responses.Where(r => r.Puzzle.ID == puzzleId && Submission.SubmissionText == r.SubmittedText).ToListAsync();
+            Submission.Response = responses.Count == 1 ? responses.First() : null;
```
morganbr: You can use responses.FirstOrDefault for this

ServerCore/Pages/Responses/Create.cshtml.cs
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        [BindProperty]
+        public Response PuzzleResponse { get; set; }
+
+        public int PuzzleId { get; set; }
+
+        public IActionResult OnGet(int puzzleId, int eventId)
+        {
+            this.PuzzleId = puzzleId;
```
morganbr: Similarly, please remove 'this.' from property accesses

ServerCore/Pages/Submissions/Index.cshtml
morganbr: As we discussed in the meeting, please keep Note for authors to take notes on responses and use a separate field

Data/DataModel/Response.cs
asyasky: Per discussion at the meeting - there should be a response string property in this class, I must have missed it in the port from the old site.

ServerCore/Pages/Responses/Create.cshtml.cs
asyasky: We want to keep the validity checks in place - they  let us do things like add attributes to the properties in the data model and have those attributes be enforced all the way through the stack.

ServerCore/Pages/Responses/Create.cshtml.cs
asyasky: I have a PR that'll go in soon (don't want to break everything currently out for review) and it'll set up automatic coding conventions for the code base. It won't apply to existing code, but I'll do a formatting pass as part of that check-in, so don't worry about these for the moment.

ServerCore/Pages/Responses/Delete.cshtml.cs
```diff
@@ -0,0 +1,59 @@
...
+    public class DeleteModel : PageModel
+    {
+        private readonly ServerCore.Models.PuzzleServerContext _context;
+
+        public DeleteModel(ServerCore.Models.PuzzleServerContext context)
+        {
+            _context = context;
+        }
+
+        [BindProperty]
```
asyasky: I'm not familiar with BindProperty - what does it do in this context?

ServerCore/Pages/Responses/Create.cshtml.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            _context = context;
+        }
+
+        [BindProperty]
+        public Response PuzzleResponse { get; set; }
+
+        public int PuzzleId { get; set; }
+
+        public IActionResult OnGet(int puzzleId, int eventId)
```
tabascq: eventId should be unnecessary

ServerCore/Pages/Responses/Delete.cshtml
tabascq: Please change the text of the link to "Back to Response List"

tabascq: Or even better, port this from the Details page: `<a asp-page="./Index" asp-route-puzzleid="@Model.PuzzleId">Back to List</a>`

ServerCore/Pages/Teams/Play.cshtml
```diff
@@ -55,25 +55,24 @@
...
             </td>
-    }
+            @if (showAnswers) {
+                <td>
+                    @Html.DisplayFor(modelItem => item.Puzzle.AnswerFile.ShortName)
+                </td>
+            }
             <td>
-                @* TODO: link to Jessica's answer submission page here *@
-                <a asp-page="./Play" asp-route-id="@item.Puzzle.ID">
```
tabascq: Information seems to be jumping between table columns here, without a matching change to the table headers. Is this a merge issue perhaps? I would have expected you to move the /Submissions/Index link to this line and remove the comment but you seem to have made far more extensive changes than that.

ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -0,0 +1,74 @@
...
+            Submission.Puzzle = await _context.Puzzles.SingleOrDefaultAsync(p => p.ID == puzzleId);
+            Submission.Team = await _context.Teams.Where((t) => t.ID == teamId).FirstOrDefaultAsync();
+
+            List<Response> responses = await _context.Responses.Where(r => r.Puzzle.ID == puzzleId && Submission.SubmissionText == r.SubmittedText).ToListAsync();
+            Submission.Response = responses.FirstOrDefault();
+
+            // Update puzzle state if submission was correct
+            if (Submission.Response != null && Submission.Response.IsSolution)
+            {
+                PuzzleStatePerTeam puzzleState = await _context.PuzzleStatePerTeam.Where(state => state.TeamID == teamId && state.PuzzleID == puzzleId).FirstOrDefaultAsync();
```
tabascq: Oops, missed this earlier - please use `PuzzleStateHelper.GetFullReadWriteQueryAsync` for this call, to match other uses. It's not strictly required in this case, but I'm trying to keep all the state code as similar as I can.

PR #62, opened 2018-09-12
Simple file upload capabilities

### LINKED ISSUE #46: File upload and access
We need authors to be able to upload puzzle files, and the system to deliver those files to players.

morganbr:
Adds backend for puzzle file management, storing the files in Azure blob storage. Currently, the only way to add a file is to set it on the puzzle's edit page, but we can add a full file management ui in a subsequent change. This also includes a controller that redirects users to the blobs for a file. 

Fixes #46


Data/DataModel/ContentFile.cs
```diff
@@ -0,0 +1,96 @@
...
+
+        /// <summary>
+        /// Identifies how this content should be handled
+        /// </summary>
+        public ContentFileType FileType { get; set; }
+
+        /// <summary>
+        /// Required for indexing. Use <see cref="Event"/> instead.
+        /// </summary>
+        public int EventID { get; set; }
```
tabascq: Why do both EventID and Event properties appear? We do not do this anywhere else. Should we?

Data/DataModel/Puzzle.cs
```diff
@@ -43,34 +46,70 @@ public class Puzzle
...
+        /// All of the content files associated with this puzzle
+        /// </summary>
+        public virtual ICollection<ContentFile> Contents { get; set; }
 
+        /// <summary>
+        /// File for the main puzzle (typically a PDF containing the puzzle)
+        /// </summary>
         [NotMapped]
-        public Uri PuzzleUrl
+        public ContentFile PuzzlePDF
```
tabascq: What if it's not a PDF? Why be specific about that?

tabascq: Ditto AnswerPDF of course.

Data/DataModel/Puzzle.cs
```diff
@@ -43,34 +46,70 @@ public class Puzzle
...
 
-        [DataType(DataType.Url)]
-        public string MaterialsUrlString { get; set; }
 
+        /// <summary>
+        /// Files associated with the puzzle, such as media
+        /// </summary>
         [NotMapped]
-        public Uri MaterialsUrl
+        public IEnumerable<ContentFile> Materials
```
tabascq: Just curious - can you build a rudimentary website out of Materials? Does that work? I know that you can't possibly have backend code, really I'm just asking about whether you can make a simple HTML page that links in lots of images/audio.

Data/DataModel/Puzzle.cs
```diff
@@ -43,34 +46,70 @@ public class Puzzle
...
+                       where contentFile.FileType == ContentFileType.PuzzleMaterial
+                       select contentFile;
+            }
         }
+
+        /// <summary>
+        /// Files unlocked by solving this puzzle, often for metapuzzles
+        /// </summary>
+        [NotMapped]
+        public IEnumerable<ContentFile> SolveTokens
```
tabascq: I think this is a great capability and support it 100% - but since tokens are often just alternate answer words that we build a meta out of, I think we need to be careful about naming. Maybe if all these were postfixed with "File" it would be clear - PuzzleFile, AnswerFile, SupportingFiles, TokenFiles...

ServerCore/Pages/FilesController.cs
```diff
@@ -0,0 +1,43 @@
...
+    public class FilesController : Controller
+    {
+        private readonly PuzzleServerContext _context;
+
+        public FilesController(PuzzleServerContext context)
+        {
+            _context = context;
+        }
+
+        [Route("{eventID}/Files/{filename}")]
```
tabascq: Rightly or wrongly, we've been using eventId, not eventID. I am fine with either but we should standardize.

Data/DataModel/ContentFile.cs
morganbr: We normally don't need to (EF will typically generate these properties on its own), but because I'm setting up an explicit index using the ID, it needs to be referenceable. EF recognizes the Foo and FooID pattern to mean that Foo is a navigation property backed by FooID.

Data/DataModel/Puzzle.cs
morganbr: Will change to PuzzleFile/AnswerFile

Data/DataModel/Puzzle.cs
morganbr: We should allow zip files that get unpacked into storage blobs (as well allowing ones that don't). I'll think about a way to do that in a future change. (#65)

ServerCore/Pages/FilesController.cs
morganbr: I'll switch to eventId

Data/DataModel/Puzzle.cs
morganbr: I want to include "solve" to make the unlocking part clear, but I'm fine with "SolveTokenFiles"

Issue #53, opened 2018-08-30
Top level menus


tabascq:
Provide a place for people to start making pages and hooking them up.

Strawman proposal follows.

[Note: ~ links only available when the event is running]
[Note: * links should work for authors AND for admins, just with a different puzzle set]
[Note: ! links are admins only]

NAME OF EVENT

Puzzles
  ~ Solved Puzzles
  ~ Unsolved Puzzles
  * Administer Puzzles
  * Import Puzzles/Responses (P2 but really really helpful)

N Updates (if any) (aka Tasks and response updates - look ma, no mailer)
  each update listed here in menu

Results
  ~ Standings
  ~ Fastest Solves
  * Puzzle State Map
  * Submissions
  * Puzzles in lockout

Registration
  My team (if known)
  Matchmaking Zone (teams ISO players, players ISO teams)
  List of Teams
  List of Players (full details for admins)
  ! Authors
  ! Admins

Resources
  FAQ
  Rules
  Encodings
  ! Event Properties
  My Events
  ! Events Dashboard

Sign up/Log in

jenetlan:
https://github.com/PuzzleServer/mainpuzzleserver/pull/136
https://github.com/PuzzleServer/mainpuzzleserver/pull/143
https://github.com/PuzzleServer/mainpuzzleserver/pull/163

jenetlan:
Done

PR #64, opened 2018-09-13
Small additions to top level menus for event pages

lambertwang:
Related: #53


Data/DataModel/Event.cs
```diff
@@ -29,19 +29,70 @@ public Uri URL
...
         public DateTime TeamRegistrationBegin { get; set; }
         public DateTime TeamRegistrationEnd { get; set; }
+
+        /// <summary>
+        /// Returns whether or not team registration is active.
+        /// </summary>
+        /// <returns>True if the current date is between the team registration begin and end times.</returns>
+        public bool IsTeamRegistrationActive()
+        {
+            return DateTime.Now.CompareTo(TeamRegistrationBegin) > 0 &&
```
tabascq: We should be using UtcNow instead of Now.

Data/DataModel/Event.cs
```diff
@@ -29,19 +29,70 @@ public Uri URL
         public bool IsInternEvent { get; set; }
         public DateTime TeamRegistrationBegin { get; set; }
         public DateTime TeamRegistrationEnd { get; set; }
+
+        /// <summary>
+        /// Returns whether or not team registration is active.
+        /// </summary>
+        /// <returns>True if the current date is between the team registration begin and end times.</returns>
+        public bool IsTeamRegistrationActive()
```
tabascq: Can we make these into standard gettable properties instead? Note that you will need the [NotMapped] attribute if they become properties.

Data/DataModel/Event.cs
```diff
@@ -29,19 +29,70 @@ public Uri URL
...
+                DateTime.Now.CompareTo(AnswerSubmissionEnd) < 0;
+        }
+
         public DateTime AnswersAvailableBegin { get; set; }
 
+        /// <summary>
+        /// Returns whether or not the puzzle answers should be available now.
+        /// </summary>
+        /// <returns>True if the current date is after the AnswersAvailableBegin time</returns>
+        public bool IsAnswersAvailableNow()
```
tabascq: My inner grammarian wants this to be "are" not "is" :-)

Data/DataModel/Event.cs
asyasky: The long term plan is to handle this differently, but for now go with UtcNow so that's consistent.

Data/DataModel/Event.cs
```diff
@@ -29,19 +29,70 @@ public Uri URL
...
         public DateTime TeamMiscDataChangeEnd { get; set; }
         public DateTime TeamDeleteEnd { get; set; }
         public DateTime EventBegin { get; set; }
+
+        /// <summary>
+        /// Returns whether or not the event has started. Does not necessarily indicate the event
+        /// is currently active.
+        /// </summary>
+        /// <returns>True if the current date is after the EventBegin time</returns>
+        public bool IsEventStarted()
```
asyasky: nit: EventHasStarted
How is "active" defined for an event if not based on the start time?

lambertwang: This only compares the current time to the event begin time. So, after the event has started, this will return true. When the event ends, the event is not 'active' but this will continue to return true.

PR #71, opened 2018-09-26
Add file management page for puzzles

morganbr:
Adds a page for uploading/deleting all of the kinds of files that can be associated with a puzzle


ServerCore/Pages/Puzzles/FileManagement.cshtml
```diff
@@ -0,0 +1,82 @@
+﻿@page  "/{eventId}/Puzzles/FileManagement/{id}/{handler?}"
```
tabascq: Is "handler" used anywhere?

morganbr: Yes, that's matched (by convention) with OnPostDeleteAsync. That makes the handler "Delete".

PR #75, opened 2018-09-27
Added footer with privacy policy and event Contact Email

lambertwang:
Added footer with privacy policy and event Contact Email

If an event does not have a ContactEmail set, then the link will default to "puzzhunt@microsoft.com"

### DIFF
--- a/Data/DataModel/PuzzleStatePerTeam.cs
+++ b/Data/DataModel/PuzzleStatePerTeam.cs
@@ -5,11 +5,7 @@ namespace ServerCore.DataModel
 {
     public class PuzzleStatePerTeam
     {
-        /// <summary>
-        /// Row ID
-        /// </summary>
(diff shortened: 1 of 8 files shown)


Data/Migrations/20180927022504_AddEventContactEmail.cs
```diff
@@ -0,0 +1,96 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Data.Migrations
+{
+    public partial class AddEventContactEmail : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
```
tabascq: The size of this migration suggests you have a problem with your DB config. Your Up/Down should only be affecting the Events table. I would not expect as much as this. Please have @morganbr or @asyasky review this part.

PR #76, opened 2018-09-27
Puzzle State Map

tabascq:
Provides the majority of the puzzle state map support from the old site.
It's missing the ability to group the results, and can't link to the
submissions page because it doesn't exist yet, but that's about it.

### DIFF
--- a/ServerCore/Pages/Events/FastestSolves.cshtml.cs
+++ b/ServerCore/Pages/Events/FastestSolves.cshtml.cs
@@ -1,0 +1,83 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.RazorPages;
+using Microsoft.EntityFrameworkCore;
+using ServerCore.DataModel;
+using ServerCore.ModelBases;
+using ServerCore.Models;
+
+namespace ServerCore.Pages.Events
+{
+    public class FastestSolvesModel : EventSpecificPageModel
+    {
+        private readonly ServerCore.Models.PuzzleServerContext _context;
+
+        public List<PuzzleStats> Puzzles { get; private set; }
+
+        public FastestSolvesModel(ServerCore.Models.PuzzleServerContext context)
+        {
+            _context = context;
+        }
+
+        // TODO: This bears striking similarities to the puzzle state map and maybe key logic could be shared. But I am also trying to keep things efficient, so maybe not.
+        // This one might also be writable in a far shorter query by someone who knows what they are doing (not me).
+        public async Task OnGetAsync()
+        {
+            // get the puzzles and teams
+            // Unlike other cases, filter to just real puzzles
+            var puzzles = await _context.Puzzles.Where(p => p.Event == this.Event && p.IsPuzzle).Select(p => new PuzzleStats() { Puzzle = p }).ToListAsync();
+            var teams = await _context.Teams.Where(t => t.Event == this.Event).ToListAsync();
+
+            // build an ID-based lookup for puzzles and teams
+            var puzzleLookup = new Dictionary<int, PuzzleStats>();
+            puzzles.ForEach(p => puzzleLookup[p.Puzzle.ID] = p);
+
+            var teamLookup = new Dictionary<int, Team>();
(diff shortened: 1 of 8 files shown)


tabascq:
I'm pretty happy with where this is right now. Not planning any further changes unless reviewers request any.

ServerCore/Pages/Events/FastestSolves.cshtml.cs
```diff
@@ -0,0 +1,88 @@
...
+            {
+                var data = puzzlesData[i];
+                var stats = new PuzzleStats()
+                {
+                    Puzzle = data.Puzzle,
+                    SolveCount = data.SolveCount,
+                    SortOrder = i
+                };
+
+                var fastEnum = data.Fastest.GetEnumerator();
```
asyasky: Is there a reason to use GetEnumerator() instead of a for loop? C# best practices typically recommend considering the Enumerator as an internal implementation detail whenever possible.

asyasky: If you switch to a for loop over a collection it would also let us change the number of fastest solves we want to display in the future. Anything that we can leave flexible is likely to simplify future changes.

ServerCore/Pages/Events/Map.cshtml
```diff
@@ -0,0 +1,83 @@
...
+                    <td>
+                        <a asp-page="/Teams/Status" asp-route-id="@team.Team.ID">@(team.Team.Name)</a>
+                    </td>
+                    <td>
+                        @Html.DisplayFor(modelItem => team.SolveCount)
+                    </td>
+                    <td>
+                        @Html.DisplayFor(modelItem => team.Score)
+                    </td>
+            @for (int x = 0; x < Model.Puzzles.Count; x++)
```
asyasky: Can this be moved into the cs backing file? The more we can keep logic in the cs file and UI in the html file the easier it will be for people to pick up just the changes they want to make.

ServerCore/Pages/Events/Map.cshtml.cs
```diff
@@ -0,0 +1,121 @@
...
+            // Build the map
+            var stateMap = new StateStats[puzzles.Count, teams.Count];
+            stateList.ForEach(state => stateMap[state.Puzzle.SortOrder, state.Team.SortOrder] = state);
+
+            this.Puzzles = puzzles;
+            this.Teams = teams;
+            this.StateMap = stateMap;
+        }
+
+        public class PuzzleStats
```
asyasky: I'd prefer one class per file, but might be persuaded to have these in a PuzzleStateMapClasses file until/unless they have more logic added. Defining the classes outside of the pages folder would make them easier to find in case they're reusable on other pages or for other logic.

ServerCore/Pages/Events/FastestSolves.cshtml.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        public List<PuzzleStats> Puzzles { get; private set; }
+
+        public FastestSolvesModel(ServerCore.Models.PuzzleServerContext context)
+        {
+            _context = context;
+        }
+
+        public async Task OnGetAsync()
+        {
```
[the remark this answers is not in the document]
tabascq: It returns all the data needed to display the fastest solves. Not sure why that needs a comment, but I can do it this evening if you like.

ServerCore/Pages/Events/Map.cshtml
tabascq: Perhaps but I do not know how at present :-)

ServerCore/Pages/Events/Map.cshtml.cs
tabascq: I do not want to reuse these as they contain data specific to the page they display on.

tabascq: And I think that is important, for performance reasons.

ServerCore/Pages/Events/FastestSolves.cshtml.cs
tabascq: Replaced it with a version that probably uses more memory but uses the "best practices" APIs.

ServerCore/Pages/Events/Map.cshtml
tabascq: OK I figured it out

tabascq:
Thanks for the review!

PR #78, opened 2018-09-27
Fix Play Page

tabascq:


tabascq:
@jenetlan - Note this is fixed! Thanks Morgan for all the help.

PR #97, opened 2018-10-23
Added some default and required fields

### LINKED ISSUE #73: Make IsPuzzle default to true
Most puzzle-like objects are puzzles, so we should set it to true by default to avoid mistakes

lambertwang:
Fixed #73

### DIFF
--- a/Data/DataModel/Event.cs
+++ b/Data/DataModel/Event.cs
@@ -8,6 +8,8 @@ namespace ServerCore.DataModel
     {
         [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
         public int ID { get; set; }
+
+        [Required]
         public string Name { get; set; }
 
         [DataType(DataType.Url)]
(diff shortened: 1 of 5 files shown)


ServerCore/Pages/Events/Create.cshtml.cs
```diff
@@ -9,19 +10,32 @@ public class CreateModel : PageModel
...
+        public Event Event { get; set; }
+
         public CreateModel(PuzzleServerContext context)
         {
             _context = context;
         }
 
         public IActionResult OnGet()
         {
+            // Ppoulate default fields
```
jenetlan: Sp: Populate

Data/DataModel/Event.cs
```diff
@@ -8,6 +8,8 @@ public class Event
     {
         [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
         public int ID { get; set; }
+
```
jenetlan: Do you need a database migration for all of these?

Issue #94, opened 2018-10-23
Submitting a null string for an answer results in NullReferenceException


lambertwang:
Caused by PR #60

Issue #52, opened 2018-08-30
Add cooldown/lockout to prevent answer spamming


morganbr:
When teams are submitting too fast, the current system allows X unlimited submissions, followed by an increasing time backoff for Y further submissions, followed by putting the puzzle in email mode -- telling the team to email an author and the author can unlock them. We should have some combination of those.

lambertwang:
Blocked by #54

PR #101, opened 2018-10-24
Implemented puzzle lockout times

lambertwang:
#94 and #52

### DIFF
--- a/Data/DataModel/PuzzleStatePerTeam.cs
+++ b/Data/DataModel/PuzzleStatePerTeam.cs
@@ -64,5 +64,53 @@ namespace ServerCore.DataModel
         /// Notes input by the team for this puzzle
         /// </summary>
         public string Notes { get; set; }
+
+        /// <summary>
+        /// If set, represents the time when a team can begin submitting answers for the
+        /// puzzle again. If a team spams answers to a puzzle, they will be locked-out
+        /// from submitting additional answers for a brief period of time.
+        /// </summary>
+        public DateTime? LockoutTime { get; set; }
+
+        /// <summary>
+        /// Returns whether or not the team is currently locked out of a puzzle.
+        /// </summary>
+        [NotMapped]
+        public bool IsTeamLockedOut
+        {
+            get { return LockoutTime == null ? false : DateTime.UtcNow.CompareTo(LockoutTime) < 0; }
+        }
+
+        /// <summary>
+        /// Returns how much lockout time is left.
+        /// </summary>
+        [NotMapped]
+        public TimeSpan? LockoutTimeRemaining
+        {
+            get { return LockoutTime == null ? null : LockoutTime - DateTime.UtcNow; }
+        }
+
+        /// <summary>
+        /// Represents the duration in minutes of the next lockout period for the team on
+        /// this puzzle. This should decrease over time to 0 if a team hasn't been locked
+        /// out in a while.
+        /// </summary>
+        public double LockoutStage { get; set; }
+
+        /// <summary>
+        /// If the team submits too many wrong answers, they become permanently locked out
+        /// of the puzzle and must email the admins.
+        /// </summary>
+        public bool IsEmailOnlyMode { get; set; }
+
+        /// <summary>
+        /// Sets the puzzle state lockout fields.
+        /// </summary>
+        /// <param name="lockoutTime">The number of minutes this puzzle should be locked out.</param>
+        public void SetPuzzleLockout (double lockoutTime)
+        {
+            LockoutStage = lockoutTime;
+            LockoutTime = DateTime.UtcNow.AddMinutes(lockoutTime);
+        }
     }
 }
--- a/Data/DataModel/Response.cs
+++ b/Data/DataModel/Response.cs
@@ -61,6 +61,11 @@ namespace ServerCore.DataModel
         /// <returns>The formatted submission</returns>
         public static string FormatSubmission(string submission)
         {
+            if (submission == null)
+            {
+                return "";
+            }
+
             return Regex.Replace(submission, @"[^a-zA-Z\d]", string.Empty).ToUpper();
         }
     }
--- a/ServerCore/Pages/Submissions/Index.cshtml
+++ b/ServerCore/Pages/Submissions/Index.cshtml
@@ -12,31 +12,40 @@
 </div>
 <br />
 <div class="row">
+    <div class="col-md-4">
     @if (!string.IsNullOrEmpty(Model.AnswerToken))
     {
-        <div class="col-md-4">
-            <div class="alert alert-success" role="alert">
-                You got it! Correct answer: @Model.AnswerToken
-            </div>
+        <div class="alert alert-success" role="alert">
+            You got it! Correct answer: @Model.AnswerToken
+        </div>
+    }
+    else if (Model.PuzzleState.IsEmailOnlyMode)
+    {
+        <div class="alert alert-danger" role="alert">
+            You have submitted too many wrong responses. Please email your solutions to <a href="mailto:@(Model.Event?.ContactEmail ?? "puzzhunt@microsoft.com")">the admins</a>.
+        </div>
+    }
+    else if (Model.PuzzleState.IsTeamLockedOut)
+    {
+        <div class="alert alert-danger" role="alert">
+           You are locked out. Please wait @Model.PuzzleState.LockoutTimeRemaining?.ToString(@"mm\:ss") before attempting another submission.
         </div>
     }
     else
     {
-        <div class="col-md-4">
-            <form method="post">
-                <div asp-validation-summary="ModelOnly" class="text-danger"></div>
-                <div class="form-group">
-                    <label asp-for="Submission.SubmissionText" class="control-label">Answer</label>
-                    <input asp-for="Submission.SubmissionText" class="form-control" />
-                    <span asp-validation-for="Submission.SubmissionText" class="text-danger"></span>
-                </div>
-                <div class="form-group">
-                    <input type="submit" value="Submit" class="btn btn-default" />
-                </div>
-            </form>
-        </div>
+        <form method="post">
+            <div asp-validation-summary="ModelOnly" class="text-danger"></div>
+            <div class="form-group">
+                <label asp-for="Submission.SubmissionText" class="control-label">Answer</label>
+                <input asp-for="Submission.SubmissionText" class="form-control" />
+                <span asp-validation-for="Submission.SubmissionText" class="text-danger"></span>
+            </div>
+            <div class="form-group">
+                <input type="submit" value="Submit" class="btn btn-default" />
+            </div>
+        </form>
     }
-
+    </div>
 </div>
 
 <div>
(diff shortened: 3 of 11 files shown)


Data/DataModel/PuzzleStatePerTeam.cs
```diff
@@ -64,5 +64,53 @@ public bool IsSolved
         /// Notes input by the team for this puzzle
         /// </summary>
         public string Notes { get; set; }
+
+        /// <summary>
+        /// If set, represents the time when a team can begin submitting answers for the
+        /// puzzle again. If a team spams answers to a puzzle, they will be locked-out
+        /// from submitting additional answers for a brief period of time.
+        /// </summary>
+        public DateTime? LockoutTime { get; set; }
```
tabascq: Suggest "LockoutExpiryTime" - I misread this as when the lockout was applied.

Data/DataModel/PuzzleStatePerTeam.cs
```diff
@@ -64,5 +64,53 @@ public bool IsSolved
...
+            get { return LockoutTime == null ? false : DateTime.UtcNow.CompareTo(LockoutTime) < 0; }
+        }
+
+        /// <summary>
+        /// Returns how much lockout time is left.
+        /// </summary>
+        [NotMapped]
+        public TimeSpan? LockoutTimeRemaining
+        {
+            get { return LockoutTime == null ? null : LockoutTime - DateTime.UtcNow; }
```
tabascq: Perhaps make this a true TimeSpan and return TimeSpan.Zero, and also clipping negative times to TimeSpan.Zero

Data/DataModel/PuzzleStatePerTeam.cs
```diff
@@ -64,5 +64,53 @@ public bool IsSolved
...
+        }
+
+        /// <summary>
+        /// Represents the duration in minutes of the next lockout period for the team on
+        /// this puzzle. This should decrease over time to 0 if a team hasn't been locked
+        /// out in a while.
+        /// </summary>
+        public double LockoutStage { get; set; }
```
tabascq: These could be ints, as discussed

Data/DataModel/Response.cs
```diff
@@ -61,6 +61,11 @@ public string SubmittedText
         /// <returns>The formatted submission</returns>
         public static string FormatSubmission(string submission)
         {
+            if (submission == null)
+            {
+                return "";
```
tabascq: Personally I prefer string.Empty

Data/DataModel/PuzzleStatePerTeam.cs
```diff
@@ -64,5 +64,53 @@ public bool IsSolved
         /// Notes input by the team for this puzzle
         /// </summary>
         public string Notes { get; set; }
+
```
tabascq: I just want to record what the old algorithm was for this, in case we get performance issues with this implementation (though I am hopeful we won't). It computed the time penalty solely by looking at how many wrong guesses were in the previous N minutes (data that the submissions page already has).

        public static async Task<TimeSpan> GetRemainingTemporaryBanDurationAsync(SqlConnection connection, int teamId, int puzzleId)
        {
            DataSet ds = await GetTeamPuzzleSubmissionsAsync(connection, teamId, puzzleId, true);
            DataTable dtIncorrectSubmissions = ds.Tables[0];
            DataView dvIncorrectSubmissions = new DataView(dtIncorrectSubmissions);
            dvIncorrectSubmissions.RowFilter = "[DateTimeSubmitted] >= #" + DateTime.UtcNow.AddMinutes(-1 * TemporaryBanIncorrectGuessPeriod).ToString() + "#";
            dvIncorrectSubmissions.Sort = "[DateTimeSubmitted] DESC";
            if (dvIncorrectSubmissions.Count > 0)
            {
                int overGuesses = dvIncorrectSubmissions.Count - TemporaryBanIncorrectGuessLimit;
                DateTime lastGuess = (DateTime)dvIncorrectSubmissions[0]["DateTimeSubmitted"];
                var penaltyFree = lastGuess.AddMinutes(overGuesses*1);
                var penalty = penaltyFree - DateTime.UtcNow;

                if (penalty > TimeSpan.Zero)
                {
                    return penalty;
                }
            }
            return TimeSpan.Zero;
        }

ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -12,31 +12,40 @@
...
-                    <label asp-for="Submission.SubmissionText" class="control-label">Answer</label>
-                    <input asp-for="Submission.SubmissionText" class="form-control" />
-                    <span asp-validation-for="Submission.SubmissionText" class="text-danger"></span>
-                </div>
-                <div class="form-group">
-                    <input type="submit" value="Submit" class="btn btn-default" />
-                </div>
-            </form>
+        <div class="alert alert-danger" role="alert">
+            You have submitted too many wrong responses. Please email your solutions to <a href="mailto:@(Model.Event?.ContactEmail ?? "puzzhunt@microsoft.com")">the admins</a>.
```
tabascq: Instead of "the admins", just repeat the address (without the mailto: of course). Then if we ever reroute the address to authors, the text will be fine.

ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -63,7 +72,7 @@
                         @Html.DisplayFor(modelItem => item.TimeSubmitted)
                     </td>
                     <td>
-                        @Html.DisplayFor(modelItem => item.SubmissionText)
+                        "@Html.DisplayFor(modelItem => item.SubmissionText)"
```
tabascq: "no quotes"

ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -31,26 +32,56 @@ public IndexModel(ServerCore.DataModel.PuzzleServerContext context)
 
         public async Task<IActionResult> OnPostAsync(int puzzleId, int teamId)
         {
+
+            // Don't allow submissions after the answer has been found.
+            if (AnswerToken != null)
```
tabascq: AnswerToken is a weird thing to key off. Should be the PuzzleStatePerTeam.SolvedTime property..

ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -31,26 +32,56 @@ public IndexModel(ServerCore.DataModel.PuzzleServerContext context)
...
+            Submissions.Add(Submission);
 
             // Update puzzle state if submission was correct
             if (Submission.Response != null && Submission.Response.IsSolution)
             {
-                var statesQ = await PuzzleStateHelper.GetFullReadWriteQueryAsync(_context, this.Event, Submission.Puzzle, Submission.Team);
-                PuzzleStatePerTeam puzzleState = await statesQ.FirstOrDefaultAsync();
-                puzzleState.IsSolved = true;
-                Submission.TimeSubmitted = (DateTime)puzzleState.SolvedTime;
+                PuzzleState.IsSolved = true;
```
tabascq: This line is redundant and I am removing the property anyway in my next PR - please remove

ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -31,26 +32,56 @@ public IndexModel(ServerCore.DataModel.PuzzleServerContext context)
...
+                // we will do the lockout computations now.
+                double? lockoutTime = ComputeLockoutTime(Event,
+                                                         Submissions,
+                                                         PuzzleState);
+                if (lockoutTime != null)
+                {
+                    PuzzleState.SetPuzzleLockout((double)lockoutTime);
+                }
+                // We also determine if the puzzle should be set to email-only mode.
+                if (IsPuzzleSubmissionLimitReached(Event, Submissions))
```
tabascq: Compute this first - possible bad vibes if a single submission creates a lockout and also puts you in email mode? No need for a double penalty in this case.

ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -59,17 +90,129 @@ public async Task<IActionResult> OnPostAsync(int puzzleId, int teamId)
...
+        public async Task<IActionResult> OnGetAsync(int puzzleId, int teamId)
         {
-            Submissions = await _context.Submissions.Where((s) => s.Team != null && s.Team.ID == teamId && s.Puzzle != null && s.Puzzle.ID == puzzleId).ToListAsync();
-            PuzzleId = puzzleId;
-            TeamId = teamId;
+            // TODO: Once auth exists, we need to check if the team has access
+            // to this puzzle.
+
+            await SetupContext(puzzleId, teamId);
+            Submission correctSubmission = this.Submissions?.Where(
```
tabascq: Really it's only the last submission that needs to be checked, right?

ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -59,17 +90,129 @@ public async Task<IActionResult> OnPostAsync(int puzzleId, int teamId)
...
+                 * If this is the team's first lockout for the puzzle or If enough times has passed
+                 * since the team's last lockout, we give them the minimum lockout duration, 1 minute.
+                 */
+                if (puzzleState.LockoutTime == null ||
+                    (DateTime.UtcNow - puzzleState.LockoutTime)?.TotalMinutes >= ev.LockoutForgivenessTime)
+                {
+                    return 1.0;
+                }
+
+                return puzzleState.LockoutStage * ev.LockoutDurationMultiplier;
```
tabascq: The way I read this, the lockout increases forever even if I put a puzzle away for hours and then come back to it. It should not do that, the multiplier should cool down. Consider borrowing some tricks from the lockout computation in the code snippet I provided and removing the LockoutStage property from PuzzleStatePerTeam - though I agree that storing the lockout expiry time in the state is a good performance win.

ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -86,4 +95,30 @@
...
+                if (timer === null) {
+                    return;
+                }
+                var timeDeltaSeconds = Math.floor((unlockTime - Date.now()) / 1000);
+                if (timeDeltaSeconds <= 0) {
+                    window.location.reload(false); 
+                }
+                var minutesLeft = Math.floor(timeDeltaSeconds / 60);
+                var secondsLeft = timeDeltaSeconds % 60;
+                timer.innerText = minutesLeft + ":" + secondsLeft;
```
tabascq: nit: what about leading zeroes, e.g. 5:09 is preferable to 5:9...

ServerCore/PuzzleStateHelper.cs
```diff
@@ -193,8 +208,90 @@ public static async Task SetSolveStateAsync(PuzzleServerContext context, Event e
...
+        /// </param>
+        /// <param name="team">
+        ///     The team; if null, get all the teams in the event.
+        /// </param>
+        /// <param name="value">The new email only state for the puzzle</param>
+        /// <param name="author"></param>
+        /// <returns>
+        ///     A task that can be awaited for the lockout operation
+        /// </returns>
+        public static async Task SetEmailOnlyModeAsync(
```
tabascq: We should have a "teams in email mode" page that authors/admins can use to kick teams out. Doesn't have to be part of this PR but let's not lose sight of it.

PR #103, opened 2018-10-24
per-Event landing page and list of events on main page

Spartan97:
I'll demo this today, but in addition to the "view all events" page, each event has its own home page that can be linked back to from elsewhere in the event (or you can go to the main page).

I am not sure what links are need on this event page, but I can add those pretty quickly tonight after some review/discussion.

Also it looks like the pull requesting is weird trying to include all my old commits in this PR), so I might have to learn how to use git branches :)


Spartan97:
Whoops -- looks like tihs was already done. That's what I get for leaving town for a few weeks... will sync and get back on track tonight.

Issue #100, opened 2018-10-23
Add a quick way to populate the database with a mock event for testing


lambertwang:
See title

PR #108, opened 2018-10-29
Demo Event Creator

tabascq:
Create a demo event that exercises most features that currently exist. Others should feel free to expand by adding things I've missed!

Implements #100.

### DIFF
--- a/ServerCore/Pages/Teams/Play.cshtml
+++ b/ServerCore/Pages/Teams/Play.cshtml
@@ -50,7 +50,7 @@
                 @Html.DisplayFor(modelItem => item.Puzzle.Group)
             </td>
             <td>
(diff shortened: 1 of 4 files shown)


ServerCore/Pages/Teams/Play.cshtml
```diff
@@ -50,7 +50,7 @@
                 @Html.DisplayFor(modelItem => item.Puzzle.Group)
             </td>
             <td>
-                @Html.DisplayFor(modelItem => item.Puzzle.OrderInGroup)
+                @(item.Puzzle.OrderInGroup < 90 ? item.Puzzle.OrderInGroup.ToString() : "Meta")
```
morganbr: Should this just use the IsMetapuzzle property? I'm not sure what 90 would mean

PR #109, opened 2018-11-05
Prevent team creation if event team registration is closed.

Spartan97:
Whoops forgot to make a PR for this. Should be a simple change -- is there anything else I need to lock down if team registration isn't active?

### DIFF
--- a/ServerCore/Pages/Teams/Create.cshtml.cs
+++ b/ServerCore/Pages/Teams/Create.cshtml.cs
@@ -30,6 +30,11 @@ namespace ServerCore.Pages.Teams
 
         public async Task<IActionResult> OnPostAsync()
         {
+            if (!Event.IsTeamRegistrationActive)
+            {
+                return NotFound();
+            }
+
             if (!ModelState.IsValid)
             {
                 return Page();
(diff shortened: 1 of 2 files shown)


morganbr:
It would be good to add the check in the backing OnPostAsync as well so we have a server-side check

ServerCore/Pages/Teams/Create.cshtml.cs
```diff
@@ -16,7 +16,12 @@ public CreateModel(PuzzleServerContext context)
         }
 
         public IActionResult OnGet()
-        {
+        {        
+            if (!Event.IsTeamRegistrationActive)
```
morganbr: It would be good to check this in OnPostAsync too to avoid issues like a team loading the page before the deadline and submitting afterward

Spartan97: Good point. I'll just need to check it in Create's OnPostAsync, right? Not anywhere else?

morganbr: I think that covers it

PR #110, opened 2018-11-06
Users/jehoang/view all submissions

digit01Wave:
Allow you to view all the puzzle submissions for an event


ServerCore/Pages/Submissions/AuthorIndex.cshtml
```diff
@@ -0,0 +1,60 @@
...
+                    @Html.DisplayFor(modelItem => item.Team.Name)
+                </td>
+                <td>
+                    @Html.DisplayFor(modelItem => item.Response.ResponseText)
+                </td>
+                <td>
+                    @Html.DisplayFor(modelItem => item.SubmissionText)
+                </td>
+                <td>
+                    @Html.DisplayFor(modelItem => item.TimeSubmitted)
```
morganbr: I think somebody figured out how to display this in local time. Is it possible to use that?

ServerCore/Pages/Submissions/AuthorIndex.cshtml.cs
```diff
@@ -0,0 +1,35 @@
...
+                        
+        public async Task OnGetAsync(int? puzzleId)
+        {
+            if (puzzleId == null)
+            {
+                Submissions = await _context.Submissions.ToListAsync();
+            }
+            else
+            {
+                Submissions = await _context.Submissions.Where((s) => s.Puzzle != null && s.Puzzle.ID == puzzleId).ToListAsync();
```
morganbr: Should this be sorted (perhaps by submission time)?

digit01Wave: Aren't these automatically sorted by time since they get added in order?

ServerCore/Pages/Submissions/AuthorIndex.cshtml
digit01Wave: We have not yet decided how to display these items in local time

PR #111, opened 2018-11-06
Add IsAdmin & IsAuthor checks

asyasky:
Added checked to unblock progress on author & admin filtering. This check will be replaced by a site wide policy soon (but use these methods in the meantime and I'll take on the cleanup that results).


Data/DataModel/EventAdmins.cs
```diff
@@ -11,7 +11,7 @@ public class EventAdmins
         /// Foreign key - event table
         /// </summary
         [ForeignKey("Event.ID")]
-        public virtual EventOwners Event { get; set; }
+        public virtual Event Event { get; set; }
```
tabascq: Needs a migration

Data/DataModel/PuzzleUser.cs
```diff
@@ -32,5 +33,35 @@ public static PuzzleUser GetPuzzleUser(string identityUserId, PuzzleServerContex
...
+        {
+            return puzzleServerContext.EventAuthors.Where(a => a.Author.ID == ID && a.Event.ID == thisEvent.ID).Any();
+        }
+
+        /// <summary>
+        /// Returns a list of the author's puzzles for the given event
+        /// </summary>
+        /// <param name="thisEvent">The event that's being checked</param>
+        /// <param name="puzzleServerContext">Current PuzzleServerContext</param>
+        public List<Puzzle> GetPuzzlesForAuthorAndEvent(PuzzleServerContext dbContext, Event thisEvent)
```
morganbr: Why would this be in this class (or this change?)

PR #118, opened 2018-12-05
Azure deployment glue

asyasky:
Adds/updates the things we need to do an Azure Deployment using Azure DevOps Continuous Deployment. Sets up the framework for using Storage & KeyVault.


ServerCore/Properties/PublishProfiles/PuzzleServerTestDeploy - Web Deploy.pubxml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8"?>
```
morganbr: You seem to have two of these. Do we actually need both?

ServerCore/Properties/launchSettings.json
```diff
@@ -12,16 +12,20 @@
...
+        "ASPNETCORE_HOSTINGSTARTUP__KEYVAULT__CONFIGURATIONENABLED": "true",
+        "ASPNETCORE_HOSTINGSTARTUP__KEYVAULT__CONFIGURATIONVAULT": "https://PuzzleServerTestKeyVault.vault.azure.net"
       }
     },
     "ServerCore": {
       "commandName": "Project",
       "launchBrowser": true,
       "environmentVariables": {
-        "ASPNETCORE_ENVIRONMENT": "Development"
+        "ASPNETCORE_ENVIRONMENT": "Development",
```
morganbr: Do you need to leave this on for now? If so, please file an issue to turn it off

ServerCore/Startup.cs
```diff
@@ -14,6 +15,17 @@ public Startup(IConfiguration configuration)
             Configuration = configuration;
         }
 
+
```
morganbr: Extra white space

ServerCore/Startup.cs
```diff
@@ -28,28 +40,41 @@ public void ConfigureServices(IServiceCollection services)
                     options.Conventions.AuthorizeFolder("/Teams");
                 });
 
-            services.AddDbContext<PuzzleServerContext>
-                (options => options.UseLazyLoadingProxies()
-                    .UseSqlServer(Configuration.GetConnectionString("PuzzleServerContext")));
+            // Use SQL Database if in Azure, otherwise, use localdb
```
morganbr: This looks like duplication of Program.cs. Which one do we actually need?

ServerCore/Startup.cs
```diff
@@ -14,6 +15,17 @@ public Startup(IConfiguration configuration)
             Configuration = configuration;
         }
 
+
+        public Startup(IHostingEnvironment env)
```
morganbr: Also looks like a dupe of Program.cs

ServerCore/Startup.cs
```diff
@@ -28,28 +40,41 @@ public void ConfigureServices(IServiceCollection services)
...
+
+            if (env.IsDevelopment() || Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT") == "Development")
             {
                 app.UseBrowserLink();
                 app.UseDeveloperExceptionPage();
-                PuzzleServerContext.UpdateDatabase(app);
             }
             else
             {
+                app.UseDeveloperExceptionPage();
```
morganbr: Please remove this or file an issue to do so

ServerCore/Startup.cs
```diff
@@ -28,28 +40,41 @@ public void ConfigureServices(IServiceCollection services)
...
+                microsoftOptions.ClientSecret = Configuration["Authentication-Microsoft-Password"];
+
             });
         }
 
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
         public void Configure(IApplicationBuilder app, IHostingEnvironment env)
         {
-            if (env.IsDevelopment())
+            PuzzleServerContext.UpdateDatabase(app);
```
morganbr: Is this how we actually want production db updates to happen?

ServerCore/Areas/Identity/IdentityHostingStartup.cs
```diff
@@ -14,9 +15,28 @@ public void Configure(IWebHostBuilder builder)
...
+                    IConfigurationBuilder configBuilder = new ConfigurationBuilder()
+                        .SetBasePath(env.ContentRootPath)
+                        .AddJsonFile("appsettings.json", optional: false, reloadOnChange: true)
+                        .AddJsonFile($"appsettings.{env.EnvironmentName}.json", optional: true)
+                        .AddEnvironmentVariables();
+                    context.Configuration = configBuilder.Build();
+
+                    services.AddDbContext<PuzzleServerContext>
+                        (options => options.UseLazyLoadingProxies()
+                            .UseSqlServer(context.Configuration.GetConnectionString("PuzzleServerSQLConnectionString")));
```
morganbr: Please document where our secrets are and how to set them up as part of this so we can set up real production if we need it

ServerCore/Properties/launchSettings.json
asyasky: This is the only way to get useful error messages right now. I'm looking into other ways for the long term (supposedly I can turn on logging in the web.config, but the web.config is generated during Publish & I haven't traced that back to what I actually need to do). I'll add an issue.

ServerCore/Startup.cs
asyasky: That is an excellent question and I don't have an answer right now. This is the simplest way to take care of the database, but it does prevent us from having full control once we have real data.

ServerCore/Areas/Identity/IdentityHostingStartup.cs
asyasky: The secrets store isn't set up yet, right now we're using the Azure Application Settings - a future change will turn on KeyVault access and I'll add docs for using that.

ServerCore/Properties/PublishProfiles/PuzzleServerTestDeploy - Web Deploy.pubxml
asyasky: Nope, good catch, thanks! :)

ServerCore/Startup.cs
asyasky: Removed (was for debugging)

PR #124, opened 2018-12-07
Add hinting system

morganbr:
Lets authors create hints for puzzles with a description, contents and a cost. Players can currently get hint coins by solving puzzles (in the future, there will also be timed and manual additions). The play page links players to a hint page that they can use to unlock and view hints.


ServerCore/Pages/Hints/Edit.cshtml.cs
```diff
@@ -0,0 +1,76 @@
...
+            if (Hint == null)
+            {
+                return NotFound();
+            }
+            return Page();
+        }
+
+        public async Task<IActionResult> OnPostAsync()
+        {
+            ModelState.Remove("Hint.Puzzle");
```
tabascq: What is this?

ServerCore/Pages/Hints/Edit.cshtml.cs
```diff
@@ -0,0 +1,76 @@
...
+
+            _context.Attach(Hint).State = EntityState.Modified;
+
+            try
+            {
+                await _context.SaveChangesAsync();
+            }
+            catch (DbUpdateConcurrencyException)
+            {
+                if (!HintExists(Hint.Id))
```
tabascq: Is all this handling actually needed? How bad would it be if you just removed the try/catch entirely?

ServerCore/Pages/Hints/Index.cshtml.cs
```diff
@@ -0,0 +1,35 @@
...
+    public class IndexModel : EventSpecificPageModel
+    {
+        private readonly ServerCore.DataModel.PuzzleServerContext _context;
+
+        public IndexModel(ServerCore.DataModel.PuzzleServerContext context)
+        {
+            _context = context;
+        }
+
+        public IList<Hint> Hint { get;set; }
```
tabascq: Hints?

ServerCore/PuzzleStateHelper.cs
```diff
@@ -168,6 +168,25 @@ public static async Task SetSolveStateAsync(PuzzleServerContext context, Event e
                 states[i].SolvedTime = value;
             }
 
+            // Award hint coins
+            if (value != null)
```
tabascq: "&& puzzle.HintCoinsForSolve != 0", for perf

ServerCore/Pages/Hints/Edit.cshtml.cs
morganbr: Added a comment (this is because ASP.NET doesn't round-trip the Puzzle property and so validation fails because it thinks it's null. This line removes the check for it)

ServerCore/Pages/Hints/Edit.cshtml.cs
morganbr: This is scaffolded boilerplate and makes error handling marginally nicer

ServerCore/Pages/Hints/Index.cshtml.cs
morganbr: Fixed

ServerCore/PuzzleStateHelper.cs
morganbr: Good idea. Done.

morganbr:
Thanks for the review!

PR #131, opened 2018-12-11
Fix Azure Pipelines deploy

asyasky:
Change from script to task in the yaml, which fixes the deployment problem.


asyasky:
No longer needed - currently checked in yaml is correct.

PR #133, opened 2018-12-12
Add authorization policy

asyasky:
Implements an authorization policy for authors, admins, and global admins.


ServerCore/Startup.cs
```diff
@@ -34,9 +36,21 @@ public void ConfigureServices(IServiceCollection services)
 
             services.AddAuthentication().AddMicrosoftAccount(microsoftOptions =>
             {
-                microsoftOptions.ClientId = Configuration["Authentication:Microsoft:ApplicationId"];
-                microsoftOptions.ClientSecret = Configuration["Authentication:Microsoft:Password"];
+                microsoftOptions.ClientId = Configuration["Authentication-Microsoft-ApplicationId"];
```
morganbr: Make sure to update the wiki and tell everyone about this change

ServerCore/Startup.cs
```diff
@@ -34,9 +36,21 @@ public void ConfigureServices(IServiceCollection services)
...
             });
+
+            services.AddAuthorization(options =>
+            {
+                options.AddPolicy("IsAuthor", policy => policy.Requirements.Add(new IsAuthorInEventRequirement()));
+                options.AddPolicy("IsAdmin", policy => policy.Requirements.Add(new IsAdminInEventRequirement()));
+                options.AddPolicy("IsGlobalAdmin", policy => policy.Requirements.Add(new IsGlobalAdminRequirement()));
+            });
+
+            services.AddTransient<IAuthorizationHandler, IsAuthorInEventHandler>();
```
morganbr: Would Scoped or Singleton work? It seems like we could end up creating a ton of objects

ServerCore/Helpers/PlayerEventHelper.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// <summary>
+        /// Returns the the team for the given player
+        /// </summary>
+        /// <param name="dbContext">Current PuzzleServerContext</param>
+        /// <param name="thisEvent">The event that's being checked</param>
+        /// <param name="user">The user being checked</param>
+        /// <returns>The user's team for this event</returns>
+        public Team GetTeamForPlayer(PuzzleServerContext dbContext, Event thisEvent, PuzzleUser user)
+        {
+            return dbContext.TeamMembers.Where(t => t.Member.ID == user.ID && t.Team.Event.ID == thisEvent.ID).Select(t => t.Team).First();
```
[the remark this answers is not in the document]
asyasky: Changed it to FirstOrDefault based on feedback from the other PR that introduced this.

ServerCore/Pages/Events/Create.cshtml.cs
```diff
@@ -1,52 +1,54 @@
 ﻿using System;
-using System.Threading.Tasks;
```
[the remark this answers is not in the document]
asyasky: Nope. All I did was add an attribute at the top of the page.

ServerCore/Startup.cs
asyasky: Singleton absolutely does not work. I'm looking at scoped now (that might be what the dbContext is by default anyway, so it could work).

ServerCore/Helpers/PlayerEventHelper.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Security.Claims;
+using Microsoft.AspNetCore.Identity;
+using ServerCore.DataModel;
+
+namespace ServerCore.Helpers
+{
+    public class PlayerEventHelper
```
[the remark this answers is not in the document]
asyasky: Good point, it's for users in general. Also added a comment to clarify.

ServerCore/Startup.cs
asyasky: Wiki updated, email will go out in a few minutes & I'll tell people tonight.

ServerCore/Startup.cs
asyasky: Scoped works, going with that & it shouldn't cause any weirdness because it's the default for PuzzleServerContext.

PR #134, opened 2018-12-12
Make first user admin in development envrionments

asyasky:
If there are no GlobalAdmins in a development environment the next user to register becomes a GlobalAdmin - this is to prevent us from getting locked out.

### DIFF
--- a/ServerCore/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
+++ b/ServerCore/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
@@ -1,7 +1,9 @@
-﻿using System.Linq;
+﻿using System;
+using System.Linq;
 using System.Security.Claims;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authorization;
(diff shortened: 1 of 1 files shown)


ServerCore/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
```diff
@@ -120,12 +122,19 @@ public async Task<IActionResult> OnPostConfirmationAsync(string returnUrl = null
 
                         if (ModelState.IsValid)
                         {
+                            // If there are no GlobalAdmins make them the GlobalAdmin
+                            if ((Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT") == EnvironmentName.Development) && !_context.PuzzleUsers.Where(u => u.IsGlobalAdmin).Any())
```
morganbr: Is there any chance of a random bot becoming global admin of our devtest site?

asyasky: Only if they can sign in with MSA

PR #135, opened 2018-12-13
Importing Events

tabascq:
Lets you import the contents of one event into another.


ServerCore/Pages/Events/Import.cshtml.cs
```diff
@@ -0,0 +1,155 @@
...
+        public async Task<IActionResult> OnGetAsync(int id)
+        {
+            Event = await _context.Events.SingleOrDefaultAsync(m => m.ID == id);
+
+            if (Event == null)
+            {
+                return NotFound();
+            }
+
+            Events = await _context.Events.Where(e => e != Event).ToListAsync();
```
morganbr: It might be nice to put in an IsAdminForEvent check for the other events to avoid picking illegal ones

ServerCore/Pages/Events/Import.cshtml.cs
```diff
@@ -0,0 +1,155 @@
...
+        {
+            // the BindProperty only binds the event ID, let's get the rest
+            Event = await _context.Events.Where((e) => e.ID == Event.ID).FirstOrDefaultAsync();
+
+            if (Event == null || await _context.Events.Where((e) => e.ID == ImportEventID).FirstOrDefaultAsync() == null)
+            {
+                return NotFound();
+            }
+
+            // TODO: validate that you are an admin of both events!!!
```
morganbr: I think this is now possible

ServerCore/Pages/Events/Import.cshtml.cs
```diff
@@ -0,0 +1,155 @@
...
+                Dictionary<int, Puzzle> puzzleCloneMap = new Dictionary<int, Puzzle>();
+
+                foreach (var sourcePuzzle in sourcePuzzles)
+                {
+                    // delete the puzzle if it exists
+                    if (deletePuzzleIfPresent)
+                    {
+                        foreach (Puzzle p in _context.Puzzles.Where((p) => p.Event == Event && p.Name == sourcePuzzle.Name))
+                        {
+                            // TODO why do I need a null check here? Does that mean something is improperly wired up?
```
morganbr: This might have been fixed in #144. I'm not sure if it's null if there are no contents

PR #136, opened 2018-12-13
Layout things

jenetlan:
Updating so that pages are event role aware

### DIFF
--- a/ServerCore/ModelBases/EventSpecificPageModel.cs
+++ b/ServerCore/ModelBases/EventSpecificPageModel.cs
@@ -17,6 +17,10 @@ namespace ServerCore.ModelBases
         [ModelBinder(typeof(EventBinder))]
         public Event Event { get; set; }
 
+        [FromRoute]
+        [ModelBinder(typeof(RoleBinder))]
+        public EventRole EventRole { get; set; }
+
         private PuzzleUser loggedInUser;
 
         public PuzzleUser LoggedInUser
@@ -46,10 +50,6 @@ namespace ServerCore.ModelBases
             userManager = manager;
         }
 
-        [FromRoute]
-        [ModelBinder(typeof(RoleBinder))]
-        public EventRole EventRole { get; set; }
-
         public class EventBinder : IModelBinder
         {
             public async Task BindModelAsync(ModelBindingContext bindingContext)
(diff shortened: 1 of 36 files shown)


ServerCore/ModelBases/EventSpecificPageModel.cs
```diff
@@ -63,5 +68,25 @@ public async Task BindModelAsync(ModelBindingContext bindingContext)
...
+                if (eventRoleAsString == null)
+                {
+                    eventRoleAsString = ModelBases.EventRole.play.ToString();
+                }
+                // TODO: Add auth check to make sure the user has permissions for the given eventRole
+                eventRoleAsString = eventRoleAsString.ToLower();
+
+                if (Enum.IsDefined(typeof(EventRole), eventRoleAsString))
+                {
+                    bindingContext.Result = ModelBindingResult.Success(Enum.Parse(typeof(EventRole), eventRoleAsString));
```
morganbr: Should this fail if the role isn't specified or isn't a real role?

ServerCore/Pages/Shared/_Layout.cshtml
```diff
@@ -1,189 +1,333 @@
...
+                        <ul class="dropdown-menu">
+                            <li><a asp-page="/." asp-route-eventRole="@ModelBases.EventRole.admin">Admin</a></li>
+                            <li><a asp-page="/." asp-route-eventRole="@ModelBases.EventRole.author">Author</a></li>
+                            <li><a asp-page="/." asp-route-eventRole="@ModelBases.EventRole.play">Player</a></li>
+                        </ul>
+                        -->
+                    </div>
+                    <div class="navbar-collapse collapse">
+                        <ul class="nav navbar-nav">
+                            <li><a asp-page="/Events/Details" asp-route-id="@Event.ID" asp-route-eventRole="@ModelBases.EventRole.admin">Event</a></li>
```
morganbr: Does relative navigation work without having to add in the two route parameters?

jenetlan: Yes - relative navigation works since the parameters show up before that point in the URL

ServerCore/ModelBases/EventSpecificPageModel.cs
jenetlan: Will add a ToDo to fail this

PR #142, opened 2018-12-13
Authors Per Puzzle

tabascq:
Allow the list of authors per puzzle to be edited, as a prelude to actual filtering.

Note that EventAuthors still needs to be edited directly in SQL, but that is more rarely needed and still needs some design thinking.

### DIFF
--- a/ServerCore/Pages/Puzzles/Edit.cshtml.cs
+++ b/ServerCore/Pages/Puzzles/Edit.cshtml.cs
@@ -24,10 +24,17 @@ namespace ServerCore.Pages.Puzzles
         public Puzzle Puzzle { get; set; }
 
         [BindProperty]
+        public int NewAuthorID { get; set; }
+
+        [BindProperty]
         public int NewPrerequisiteID { get; set; }
 
         [BindProperty]
         public int NewPrerequisiteOfID { get; set; }
+
+        public List<PuzzleUser> PotentialAuthors { get; set; }
+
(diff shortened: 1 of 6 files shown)


Data/Migrations/PuzzleServerContextModelSnapshot.cs
```diff
@@ -799,13 +799,15 @@ protected override void BuildModel(ModelBuilder modelBuilder)
 
             modelBuilder.Entity("ServerCore.DataModel.PuzzleAuthors", b =>
                 {
-                    b.HasOne("ServerCore.DataModel.Puzzle", "Puzzle")
+                    b.HasOne("ServerCore.DataModel.PuzzleUser", "Author")
```
morganbr: If you add [Required] to the Puzzle and Author properties, you probably don't need the customization in this method.

ServerCore/Pages/Puzzles/Edit.cshtml.cs
```diff
@@ -89,6 +102,22 @@ public async Task<IActionResult> OnPostAsync()
             return RedirectToPage("./Index");
         }
 
+        public async Task<IActionResult> OnPostAddAuthorAsync()
+        {
+            if (!(await _context.EventAuthors.Select(m => m.Author.ID == NewAuthorID).AnyAsync()))
```
morganbr: This should also probably check that the author is in the current event

PR #144, opened 2018-12-13
Add file controller authorization

morganbr:
Add authorization for navigating to files. Also fixes lazy loading proxies and file display to players

### DIFF
--- a/ServerCore/Pages/FilesController.cs
+++ b/ServerCore/Pages/FilesController.cs
@@ -1,28 +1,37 @@
-﻿using System.Linq;
+﻿using System;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Identity;
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.EntityFrameworkCore;
 using ServerCore.DataModel;
+using ServerCore.Helpers;
 
 namespace ServerCore.Pages
 {
     /// <summary>
     /// Controller to redirect users to files in blob storage
     /// </summary>
+    [Authorize]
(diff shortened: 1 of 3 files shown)


ServerCore/Pages/FilesController.cs
```diff
@@ -35,8 +44,83 @@ public IActionResult Index(int eventId, string filename)
...
+            PuzzleUser user = PuzzleUser.GetPuzzleUserForCurrentUser(context, User, userManager);
+
+            // Admins can see all files
+            if (user.IsAdminForEvent(context, currentEvent))
+            {
+                return true;
+            }
+
+            // Authors can see all files attached to their puzzles
+            if (await UserEventHelper.IsAuthorOfPuzzle(context, puzzle, user))
```
asyasky: Can you just return this directly?

ServerCore/Pages/FilesController.cs
```diff
@@ -35,8 +44,83 @@ public IActionResult Index(int eventId, string filename)
...
+            }
+
+            Team team = await UserEventHelper.GetTeamForPlayer(context, currentEvent, user);
+            if (team == null)
+            {
+                return false;
+            }
+
+            // Once answers are available, so are all other files
+            if (currentEvent.AreAnswersAvailableNow)
```
asyasky: Ditto

ServerCore/Pages/FilesController.cs
morganbr: Nope, as discussed

ServerCore/Pages/FilesController.cs
morganbr: Ditto

PR #147, opened 2018-12-22
Feedback -- role awareness

Spartan97:
A few changes here:
1) Updated the VScode project files so that the debugger actually builds and runs the correct project
2) Trying to view feedback for a puzzle routes you to the appropriate page (Submit or View) based on your role. /play/ routes to submit feedback while /author/ and /admin/ route to view feedback. Authors can only view feedback for puzzles they created.
3) Updated the feedback to track the user that submitted it and aggregate all of the fun/difficulty ratings at the top of the view page.
4) When a player that has already submitted feedback on a puzzle tries to submit more feedback, it overwrites the ratings and appends the written feedback. Probably could do with a warning or something here, but for now its good enough.

### DIFF
--- a/ServerCore/Pages/Puzzles/Feedback.cshtml
+++ b/ServerCore/Pages/Puzzles/Feedback.cshtml
@@ -4,52 +4,79 @@
 @{
     ViewData["Title"] = "Feedback";
     var id = ViewContext.RouteData.Values["id"];
+    var role = ViewContext.RouteData.Values["eventRole"].ToString();
 }
 
-<h2>Feedback for @Html.DisplayFor(model => model.Puzzle.Name)</h2>
+@if (role == "admin" || (role == "author" && Model.IsCurrentUserPuzzleAuthor))
+{
+    <h2>Feedback for @Html.DisplayFor(model => model.Puzzle.Name)</h2>
 
-<table class="table">
-    <thead>
-        <tr>
-            <th>
-                @Html.DisplayNameFor(model => model.Feedbacks[0].Submitter)
-            </th>
-            <th>
-                @Html.DisplayNameFor(model => model.Feedbacks[0].SubmissionTime)
-            </th>
-            <th>
-                @Html.DisplayNameFor(model => model.Feedbacks[0].WrittenFeedback)
-            </th>
-            <th>
-                @Html.DisplayNameFor(model => model.Feedbacks[0].Difficulty)
-            </th>
-            <th>
-                @Html.DisplayNameFor(model => model.Feedbacks[0].Fun)
-            </th>
-        </tr>
-    </thead>
-    <tbody>
-@foreach (var item in Model.Feedbacks) {
-        <tr>
-            <td>
-                @Html.DisplayFor(modelItem => item.Submitter.Name)
-            </td>
-            <td>
-                @Html.DisplayFor(modelItem => item.SubmissionTime)
-            </td>
-            <td>
-                @Html.DisplayFor(modelItem => item.WrittenFeedback)
-            </td>
-            <td>
-                @Html.DisplayFor(modelItem => item.Difficulty)
-            </td>
-            <td>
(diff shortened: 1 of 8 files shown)


Spartan97:
I can demo this at our first meeting back from the holidays :)

ServerCore/Pages/Puzzles/Feedback.cshtml
```diff
@@ -4,52 +4,79 @@
 @{
     ViewData["Title"] = "Feedback";
     var id = ViewContext.RouteData.Values["id"];
+    var role = ViewContext.RouteData.Values["eventRole"].ToString();
```
tabascq: You don't need to compute the role and do this if/else here. Instead you can put an attribute on the page, and in fact we've already put it on because we did a pass over all the pages. Merge and you will see it in the class file:

    [Authorize(Policy = "IsEventAdminOrAuthorOfPuzzle")]

ServerCore/Pages/Puzzles/Feedback.cshtml.cs
```diff
@@ -16,27 +17,30 @@ public class FeedbackModel : EventSpecificPageModel
...
-            Feedbacks = await _context.Feedback.Where((f) => f.Puzzle.ID == id).ToListAsync();
+            Feedbacks = await _context.Feedback.Where((f) => f.Puzzle.ID == id).Include("Submitter").ToListAsync();
             Puzzle = await _context.Puzzles.Where((p) => p.ID == id).FirstOrDefaultAsync();
 
             if (Puzzle == null) 
             { 
                 return NotFound(); 
             } 
 
+            IsCurrentUserPuzzleAuthor = (await _context.PuzzleAuthors.Where((a) => a.Puzzle == Puzzle && a.Author == LoggedInUser).FirstOrDefaultAsync() != null);
```
tabascq: If you're just using this for auth, you should be able to do it with attributes. If it serves another purpose, OK to keep.

ServerCore/Pages/Puzzles/Index.cshtml
```diff
@@ -91,8 +93,10 @@
                 <a asp-Page="./Details" asp-route-id="@item.ID">Details</a> |
                 <a asp-Page="./Delete" asp-route-id="@item.ID">Delete</a> |
                 <a asp-page="./Status" asp-route-id="@item.ID">Status</a> |
-                <a asp-Page="./Feedback" asp-route-id="@item.ID">Admin View Feedback</a> |
-                <a asp-Page="./SubmitFeedback" asp-route-id="@item.ID">User Submit Feedback</a> |
+                @{
+                    var feedbackLink = (role == "play") ? "./SubmitFeedback" : "./Feedback";
```
tabascq: Puzzles/Index is currently unused in the play role. We use Teams/Play instead. [Maybe these pages should be merged at some point, maybe not.]

ServerCore/Pages/Puzzles/SubmitFeedback.cshtml
```diff
@@ -4,55 +4,63 @@
 @{
     ViewData["Title"] = "Submit Feedback";
     var id = ViewContext.RouteData.Values["id"];
+    var role = ViewContext.RouteData.Values["eventRole"].ToString();
 }
 
-<h2>Submit Feedback for @Html.DisplayFor(model => model.Puzzle.Name)</h2> 
+@if (role == "play")
```
tabascq: We have an attribute for this too.

Spartan97:
@tabascq The merge got a bit ugly (as expected), so I'm going to rebranch from master start a new PR. Thanks for the comments -- it looks like it'll be pretty easy. Is there a page that completely uses the new auth stuff that I can take a look at to wrap my head around it?

tabascq:
Puzzles/Index is a fine example that uses an attribute for the auth part and then provides slightly different data as well.

PR #155, opened 2019-01-13
Team navigation & partial auth

jenetlan:


ServerCore/Pages/Teams/Index.cshtml.cs
```diff
@@ -1,13 +1,15 @@
...
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Identity;
 using Microsoft.EntityFrameworkCore;
 using ServerCore.DataModel;
 using ServerCore.ModelBases;
 
 namespace ServerCore.Pages.Teams
 {
+    [Authorize(Policy = "IsEventAdminOrEventAuthor")]
```
morganbr: Since this is becoming an admin view, would it make sense to rename it and make the index be the player page?

ServerCore/Pages/Teams/Index.cshtml
```diff
@@ -8,7 +8,7 @@
 <h2>All Teams</h2>
 
 <p>
-    @if (Model.Event.IsTeamRegistrationActive) // Or current user is admin
+    @if (Model.Event.IsTeamRegistrationActive && Model.EventRole == ModelBases.EventRole.admin)
```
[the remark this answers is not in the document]
jenetlan: There was a change since the comment - this page is only going to be for authors & admins, so if you're not an admin you can't create a team (since authors can't create teams)

ServerCore/Pages/Teams/Index.cshtml.cs
jenetlan: I think that causes a lot of issues with the navigation we have in place (getting to / from teams list) - for now if it's ok I'll leave it

PR #156, opened 2019-01-13
Event Admin fixups

tabascq:
Event-specific Event pages now use EventSpecificPageModel so that their auth works properly.

The only event pages that require global admin rights are Create, CreateDemo, and the Index.

### DIFF
--- a/ServerCore/Pages/Events/Index.cshtml
+++ b/ServerCore/Pages/Events/Index.cshtml
@@ -43,14 +43,14 @@
             <td>
                 <a asp-page="./Edit" asp-route-eventId="@item.EventID" asp-route-eventRole="admin">Edit</a> |
                 <a asp-page="./Details" asp-route-eventId="@item.EventID" asp-route-eventRole="admin">Details</a> |
-                <a asp-page="/Events/Delete" asp-route-eventId="@item.EventID" asp-route-eventRole="admin">Delete</a> |
+                <a asp-page="./Delete" asp-route-eventId="@item.EventID" asp-route-eventRole="admin">Delete</a> |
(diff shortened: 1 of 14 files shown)


ServerCore/Pages/Events/Index.cshtml
```diff
@@ -41,10 +41,10 @@
...
             </td>
             <td>
-                <a asp-page="./Edit" asp-route-id="@item.ID">Edit</a> |
-                <a asp-page="./Details" asp-route-id="@item.ID">Details</a> |
-                <a asp-page="./Delete" asp-route-id="@item.ID">Delete</a> |
-                <a asp-page="./Import" asp-route-id="@item.ID" asp-route-eventRole="admin">Import</a> |
+                <a asp-page="./Edit" asp-route-eventId="@item.EventID" asp-route-eventRole="admin">Edit</a> |
+                <a asp-page="./Details" asp-route-eventId="@item.EventID" asp-route-eventRole="admin">Details</a> |
+                <a asp-page="/Events/Delete" asp-route-eventId="@item.EventID" asp-route-eventRole="admin">Delete</a> |
+                <a asp-page="./Import" asp-route-eventId="@item.EventID" asp-route-eventRole="admin">Import</a> |
```
jenetlan: Did you add "Events" as part of the route for "Edit", "Details",  and "Import" as well? Is there a reason they aren't updated here?
On that note, I don't think you have to include "Events" in the route at all - any reason we need that word to show up for these pages? Is this something we should be doing? (I'll end up in the same spot when I add FAQs - wondering if the url should be "/{eventId}/{eventRole}/Events/FAQ" or just "/{eventId}/{eventRole}/FAQ"

tabascq: I standardized the links here - "Events" isn't required.

PR #158, opened 2019-01-14
Team applications

morganbr:
Give users a way to apply to a team and only allow teams to add users who have applied. Admins can add anyone to a team. Also includes fixing up auth for some of the pages involved.

### DIFF
--- a/ServerCore/Pages/Teams/Apply.cshtml
+++ b/ServerCore/Pages/Teams/Apply.cshtml
@@ -10,7 +10,7 @@
 <hr />
 <div class="row">
     <div class="col-md-4">
-        You have applied to <strong>@{@Model.Team.Name}</strong>. Once they accept you, you will be on the team.
+        You have applied to <strong>@{@Model.Team.Name}</strong>. Once they accept you, you will be on the team (make sure to contact somebody on the team and ask them to add you!). If you've previously requested to join to another team, that request has been removed.
     </div>
 </div>
 
--- a/ServerCore/Pages/Teams/Apply.cshtml
+++ b/ServerCore/Pages/Teams/Apply.cshtml
@@ -1,27 +1,21 @@
-﻿@page
+﻿@page  "/{eventId}/{eventRole}/Teams/Apply"
 @model ServerCore.Pages.Teams.ApplyModel
 
 @{
     ViewData["Title"] = "Apply";
 }
 
-<h2>Apply</h2>
+<h2>Apply to a Team</h2>
 
-<h4>TeamApplication</h4>
 <hr />
 <div class="row">
     <div class="col-md-4">
-        <form method="post">
-            <div asp-validation-summary="ModelOnly" class="text-danger"></div>
-            <div class="form-group">
-                <input type="submit" value="Create" class="btn btn-default" />
-            </div>
-        </form>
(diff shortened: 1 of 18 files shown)


ServerCore/Pages/Teams/Apply.cshtml
```diff
@@ -0,0 +1,23 @@
...
+@{
+    ViewData["Title"] = "Apply";
+}
+
+<h2>Apply to a Team</h2>
+
+<hr />
+<div class="row">
+    <div class="col-md-4">
+        You have applied to <strong>@{@Model.Team.Name}</strong>. Once they accept you, you will be on the team.
```
tabascq: I would include a couple more points here:

1. If you had previously applied to a different team, that application has been removed as you can only apply to one team at a time.
2. Your team will not be notified of your application, and it is recommended that you follow up with email to close the loop.

ServerCore/Pages/Teams/Create.cshtml.cs
```diff
@@ -1,23 +1,41 @@
...
+            {
+                return Challenge();
+            }
+
+            if (EventRole != EventRole.play && EventRole != EventRole.admin)
+            {
+                return NotFound();
+            }
+
+            if (EventRole == EventRole.admin && !await LoggedInUser.IsAdminForEvent(_context, Event))
```
tabascq: Is this necessary? I would hope that this would automatically be authenticated earlier.

tabascq: Actually, it looks like AuthorizationHelper methods would allow these to slip through, but judicious use of authContext.Fail() should close that gap.

ServerCore/Pages/Teams/List.cshtml
```diff
@@ -0,0 +1,50 @@
...
+        </div>
+    }
+</p>
+<table class="table">
+    <thead>
+        <tr>
+            <th>
+                @Html.DisplayNameFor(model => model.Teams[0].Name)
+            </th>
+            @if (Model.EventRole == ModelBases.EventRole.play && !Model.UserOnTeam && Model.Event.IsTeamMembershipChangeActive)
```
tabascq: Maybe a column for team size?

ServerCore/Pages/Teams/List.cshtml.cs
```diff
@@ -0,0 +1,56 @@
...
+
+namespace ServerCore.Pages.Teams
+{
+    /// <summary>
+    /// Player-facing list of all teams
+    /// </summary>
+    [AllowAnonymous]
+    public class ListModel : EventSpecificPageModel
+    {
+
```
tabascq: unnecessary blank line

ServerCore/Pages/Teams/Create.cshtml.cs
morganbr: AllowAnonymous turns off the auth checks, which I'm using because the base class would otherwise require players to be on a team to create a team.

ServerCore/Pages/Teams/List.cshtml
morganbr: Done. We'll have to keep an eye on perf for it (since getting a count involves querying all of the players in the event)

PR #160, opened 2019-01-15
Timed Unlocks

tabascq:
Implementing timed unlocks the way the old site did them, which is not nearly as cool as a wave-based approach but is better than poking the puzzle status manually during all betas and the live event.

The unlocking computation has a 10-minute throttle that is applied per team.

### DIFF
--- a/ServerCore/Areas/Identity/UserAuthorizationPolicy/AuthorizationHelper.cs
+++ b/ServerCore/Areas/Identity/UserAuthorizationPolicy/AuthorizationHelper.cs
@@ -96,6 +96,8 @@ namespace ServerCore.Areas.Identity
             {
                 authContext.Succeed(requirement);
             }
+
+            dbContext.Entry(puzzle).State = EntityState.Detached;
         }
 
         public static async Task IsEventAuthorCheck(AuthorizationHandlerContext authContext, PuzzleServerContext dbContext, UserManager<IdentityUser> userManager, IAuthorizationRequirement requirement)
--- a/ServerCore/Pages/Teams/Play.cshtml.cs
+++ b/ServerCore/Pages/Teams/Play.cshtml.cs
@@ -33,6 +33,8 @@ namespace ServerCore.Pages.Teams
             this.TeamID = teamId;
             this.Sort = sort;
 
+            await PuzzleStateHelper.CheckForTimedUnlocksAsync(_context, Event, teamId);
+
             // all puzzles for this event that are real puzzles
             var puzzlesInEventQ = _context.Puzzles.Where(puzzle => puzzle.Event.ID == this.Event.ID && puzzle.IsPuzzle);
 
--- a/ServerCore/Pages/Teams/Play.cshtml.cs
+++ b/ServerCore/Pages/Teams/Play.cshtml.cs
@@ -37,14 +37,13 @@ namespace ServerCore.Pages.Teams
             if (myTeam != null)
             {
                 this.TeamID = myTeam.ID;
+                await PuzzleStateHelper.CheckForTimedUnlocksAsync(_context, Event, myTeam);
             }
             else
             {
                 throw new Exception("Not currently registered for a team");
             }
             this.Sort = sort;
-
-            await PuzzleStateHelper.CheckForTimedUnlocksAsync(_context, Event, teamId);
 
             // all puzzles for this event that are real puzzles
             var puzzlesInEventQ = _context.Puzzles.Where(puzzle => puzzle.Event.ID == this.Event.ID && puzzle.IsPuzzle);
--- a/ServerCore/PuzzleStateHelper.cs
+++ b/ServerCore/PuzzleStateHelper.cs
@@ -304,19 +304,17 @@ namespace ServerCore
         public static async Task CheckForTimedUnlocksAsync(
             PuzzleServerContext context,
             Event eventObj,
-            int teamId)
+            Team team)
         {
             // throttle this by an expiry interval before we do anything even remotely expensive
-            if (TimedUnlockExpiryCache.TryGetValue(teamId, out DateTime expiry) && expiry >= DateTime.UtcNow)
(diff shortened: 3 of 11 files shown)


tabascq:
Note: I haven't been able to run this code yet because I can't get the player experience to work at all post-auth. So I won't promise it is 100% correct, though the rough shape is still worth a look.

ServerCore/Areas/Identity/UserAuthorizationPolicy/AuthorizationHelper.cs
```diff
@@ -96,6 +96,8 @@ public static async Task IsPuzzleAuthorCheck(AuthorizationHandlerContext authCon
             {
                 authContext.Succeed(requirement);
             }
+
+            dbContext.Entry(puzzle).State = EntityState.Detached;
```
jenetlan: What is the goal of this line? It seems like it would only run if the auth check failed - I can't figure out what it's supposed to do

ServerCore/Pages/Teams/Play.cshtml.cs
```diff
@@ -33,6 +33,8 @@ public async Task OnGetAsync(int teamId, SortOrder? sort)
             this.TeamID = teamId;
             this.Sort = sort;
 
+            await PuzzleStateHelper.CheckForTimedUnlocksAsync(_context, Event, teamId);
```
jenetlan: Does this generate a task that auto-runs or just one that runs every time the user refreshes the page?

ServerCore/Areas/Identity/UserAuthorizationPolicy/AuthorizationHelper.cs
tabascq: The "Succeed" line above is not a return - so this line actually runs in all cases. This is something I discovered this evening, where auth prevents a puzzle from being edited because it is keeping an entity attached.

ServerCore/Pages/Teams/Play.cshtml.cs
tabascq: It runs every time the user refreshes the page - but CheckForTimedUnlocks won't do its job any more frequently than once per 10 minutes (per team).

jenetlan: Got it - thanks :)

ServerCore/Areas/Identity/UserAuthorizationPolicy/AuthorizationHelper.cs
jenetlan: What does the line do then? I thought the only purpose of the call was to set succeeded / failed, which doesn't seem related to this line? Is it detaching a puzzle, and if so what does that mean and why?

tabascq: The auth check is keeping some sort of lock on the puzzle that prevents it from being edited. This line removes that lock. As to why we don't have similar locking issues all over the codebase, I can't answer that. I only know that without this line, I can't edit a puzzle, and with this line, I can.

tabascq:
After merging with Jenna's work, I was able to navigate my way through the Play experience and saw this working properly (after I fixed a key bug, that is).

ServerCore/PuzzleStateHelper.cs
```diff
@@ -294,6 +294,38 @@ public static async Task SetUnlockStateAsync(PuzzleServerContext context, Event
...
+                .Where(state => state.SolvedTime == null && state.UnlockedTime != null && state.Puzzle.MinutesToAutomaticallySolve != null && state.UnlockedTime.Value + TimeSpan.FromMinutes(state.Puzzle.MinutesToAutomaticallySolve.Value) <= now)
+                .Select(state => state.Puzzle)
+                .ToListAsync();
+
+            foreach (Puzzle puzzle in puzzlesToSolveByTime)
+            {
+                await PuzzleStateHelper.SetSolveStateAsync(context, eventObj, puzzle, team, DateTime.UtcNow);
+            }
+
+            TimedUnlockExpiryCache[team.ID] = DateTime.UtcNow + TimedUnlockExpiryInterval;
```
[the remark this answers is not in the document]
tabascq: I took your concerns to heart and rewrote the expiry scheduling. The per-team expiry throttle is gone, but globally, the requests are throttled to be at least 2s apart.

ServerCore/PuzzleStateHelper.cs
```diff
@@ -294,6 +294,38 @@ public static async Task SetUnlockStateAsync(PuzzleServerContext context, Event
             await context.SaveChangesAsync();
         }
 
+        private static Dictionary<int, DateTime> TimedUnlockExpiryCache = new Dictionary<int, DateTime>();
+#if DEBUG
+        private static readonly TimeSpan TimedUnlockExpiryInterval = TimeSpan.FromMinutes(1);
+#else
+        private static readonly TimeSpan TimedUnlockExpiryInterval = TimeSpan.FromMinutes(10);
```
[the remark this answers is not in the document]
tabascq: I took your concerns to heart and rewrote the expiry scheduling. The per-team expiry throttle is gone, but globally, the requests are throttled to be at least 2s apart.

ServerCore/PuzzleStateHelper.cs
```diff
@@ -294,6 +294,38 @@ public static async Task SetUnlockStateAsync(PuzzleServerContext context, Event
             await context.SaveChangesAsync();
         }
 
+        private static Dictionary<int, DateTime> TimedUnlockExpiryCache = new Dictionary<int, DateTime>();
```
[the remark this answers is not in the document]
tabascq: Good point. But since my newer design has a global throttle that is outside the dictionary, I'm going with a lock so I can keep it all consistent. The expensive stuff happens outside the lock, so I think it'll be OK.

tabascq:
@morganbr , I know you approved this, but can you give it another look to confirm that I've got a good concurrency approach here?

PR #161, opened 2019-01-15
Role-switching & improved page-to-page navigation

jenetlan:


ServerCore/Pages/EventSpecific/FAQ.cshtml
```diff
@@ -0,0 +1,13 @@
+﻿@page "/{eventId}/{eventRole}/faq"
+@model ServerCore.Pages.EventSpecific.FAQModel
+@{
+    ViewData["Title"] = "FAQ";
+    ViewData["AdminRoute"] = "/EventSpecific/FAQ";
```
tabascq: I don't understand why these use different route info but maybe it will become clear when I get to the page that uses them.

ServerCore/Pages/EventSpecific/Index.cshtml
```diff
@@ -1,7 +1,23 @@
-﻿@page "/{eventId}"
+﻿@page "/{eventId}/{eventRole}"
 @model ServerCore.Pages.EventSpecific.IndexModel
 @{
     ViewData["Title"] = Model.Event.Name;
+    ViewData["AdminRoute"] = "/EventSpecific/Index";
```
tabascq: Maybe the absence of the data means go to the index? You'd have fewer records that way.

ServerCore/Pages/Hints/Create.cshtml
```diff
@@ -2,10 +2,14 @@
 @model ServerCore.Pages.Hints.CreateModel
 
 @{
-    ViewData["Title"] = "Create";
+    ViewData["Title"] = "Create hint";
+    //Needs route data - ViewData["AdminRoute"] = "/Hints/Create";
```
tabascq: Maybe go somewhere like the puzzles list in these cases? Not sure.

ServerCore/Pages/Shared/_Layout.cshtml
```diff
@@ -28,6 +28,15 @@
     @{
         var Event = (Model as ServerCore.ModelBases.EventSpecificPageModel)?.Event;
         var EventRole = (Model as ServerCore.ModelBases.EventSpecificPageModel)?.EventRole;
+        var AdminRoute = "/EventSpecific/Index";
```
tabascq: Behold the fancy "??" operator that condenses each of these to one line:

var AdminRoute = ViewData["AdminRoute"] as string ?? "/EventSpecific/Index";

"??" means "use the thing to my left if it is non-null, otherwise use the thing to my right".

ServerCore/Pages/Shared/_Layout.cshtml
```diff
@@ -166,26 +175,46 @@
                             <span class="icon-bar"></span>
                             <span class="icon-bar"></span>
                         </button>
-                        <a asp-page="/." class="navbar-brand" data-toggle="dropdown">
+                        <a class="navbar-brand" data-toggle="dropdown" href="#">
                             @Event.Name - Admin
                         </a>
+                        <!-- TODO: Check for permissions / make partial -->
```
tabascq: Put IsEventAuthor and IsEventAdmin methods on EventSpecificPageModel?

ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -11,7 +11,7 @@
 
 namespace ServerCore.Pages.Submissions
 {
-    [Authorize(Policy = "IsEventAdminOrAuthorOfPuzzle")]
+    //TODO: Not working [Authorize(Policy = "PlayerCanSeePuzzle")]
```
tabascq: This page is used by all three roles?

ServerCore/ServerCore.csproj
```diff
@@ -43,4 +43,15 @@
   <ItemGroup>
     <WCFMetadata Include="Connected Services" />
   </ItemGroup>
+  <ItemGroup>
+    <Content Update="Pages\EventSpecific\FAQ.cshtml">
```
tabascq: What's up with these? Most pages don't need this sort of thing. Are you sure it's necessary?

jenetlan: Odd - I don't think I touched those. It might have been something automatic?

ServerCore/Pages/Submissions/Index.cshtml.cs
jenetlan: AuthorIndex is the one that looks like it should be used for Authors / Admins, although I think Jeremy has a PR out that might change that.

ServerCore/Pages/Shared/_Layout.cshtml
jenetlan: Good idea - made a note to myself so I can try it out :)

ServerCore/Pages/Hints/Create.cshtml
jenetlan: I'm hoping I can figure this out soon enough that we won't have to deal with it for long - if I can't get it working I can look to switch to a different backup (right now they all go to the event home page

ServerCore/Pages/EventSpecific/Index.cshtml
jenetlan: As in remove the word "Index" or as in remove this line entirely since it is the default? I included the line so that I wouldn't think I had forgotten to add an admin route and to keep this consistent in case the default changes, but I could clean up the word "Index"

ServerCore/Pages/EventSpecific/FAQ.cshtml
jenetlan: I need to set all three routes so that no matter which role the user selects I'll have a non-default value for that role. It just so happens that for this page they're all the same :)

ServerCore/Pages/EventSpecific/Index.cshtml
jenetlan: Will do in separate cleanup if needed

ServerCore/Pages/Shared/_Layout.cshtml
jenetlan: Oh my goodness this is awesome - so much cleaner :D Updated :)

PR #163, opened 2019-01-17
Navigation for puzzle role switching and team signup + unregistered user header

jenetlan:


ServerCore/Pages/Teams/Create.cshtml.cs
```diff
@@ -111,6 +111,10 @@ public async Task<IActionResult> OnPostAsync()
                 transaction.Commit();
             }
 
+            if(EventRole == ModelBases.EventRole.play)
```
tabascq: Why not just do this all the time?

PR #183, opened 2019-01-17
Feedback pages improvements

Spartan97:
- If a user has already submitted feedback it will seed the page with what they submitted last time.
- When they resubmit it will update instead of duplicating.
- Shows user that submitted each piece of feedback in author view
- Aggregates all feedback in author view

### DIFF
--- a/ServerCore/DataModel/Feedback.cs
+++ b/ServerCore/DataModel/Feedback.cs
@@ -11,7 +11,7 @@ namespace ServerCore.DataModel
         [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
         public int ID { get; set; }
         public Puzzle Puzzle { get; set; }
-        public User Submitter {get; set;}
+        public PuzzleUser Submitter {get; set;}
         public DateTime SubmissionTime { get; set; }
         public string WrittenFeedback { get; set; }
         public int Difficulty { get; set; }
--- a/ServerCore/Pages/Puzzles/Feedback.cshtml
+++ b/ServerCore/Pages/Puzzles/Feedback.cshtml
@@ -8,25 +8,14 @@
     ViewData["PlayRoute"] = "../Puzzles/SubmitFeedback";
     ViewData["RoutingPuzzleId"] = ViewContext.RouteData.Values["puzzleId"];
     Layout = "_puzzleManagementLayout.cshtml";
-
-    var sumFun = 0.0;
-    var sumDiff = 0.0;
-    var count = 0;
-    foreach (var item in Model.Feedbacks) {
-        sumFun += item.Fun;
-        sumDiff += item.Difficulty;
-        count++;
-    }
-    sumFun /= count;
-    sumDiff /= count;
 }
 
 <h2>@Html.DisplayFor(model => model.Puzzle.Name): Feedback</h2>
 
 <div>
-    <h3><b>Avg Fun Rating:</b> @sumFun</h3>
+    <h3><b>Avg Fun Rating:</b> @Model.FunScore</h3>
     <br>
-    <h3><b>Avg Difficulty Rating:</b> @sumDiff</h3>
+    <h3><b>Avg Difficulty Rating:</b> @Model.DiffScore</h3>
 </div>
 
 <br>
--- a/ServerCore/Pages/Puzzles/Feedback.cshtml
+++ b/ServerCore/Pages/Puzzles/Feedback.cshtml
@@ -8,10 +8,28 @@
     ViewData["PlayRoute"] = "../Puzzles/SubmitFeedback";
     ViewData["RoutingPuzzleId"] = ViewContext.RouteData.Values["puzzleId"];
     Layout = "_puzzleManagementLayout.cshtml";
-    var id = ViewContext.RouteData.Values["id"];
+
+    var sumFun = 0.0;
+    var sumDiff = 0.0;
+    var count = 0;
+    foreach (var item in Model.Feedbacks) {
+        sumFun += item.Fun;
+        sumDiff += item.Difficulty;
+        count++;
+    }
+    sumFun /= count;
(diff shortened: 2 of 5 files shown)


ServerCore/DataModel/Feedback.cs
```diff
@@ -11,7 +11,7 @@ public class Feedback
         [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
         public int ID { get; set; }
         public Puzzle Puzzle { get; set; }
-        public User Submitter {get; set;}
+        public PuzzleUser Submitter {get; set;}
```
tabascq: Do we need a DB migration for this?

Spartan97: I'm not entirely sure. Adding Morgan and Megan because hopefully one of them knows and can tell me what to do if the answer is yes.

Spartan97:
@PuzzleServer/adminteam Can someone take a look at this so its ready for checkin after I demo tomorrow?

ServerCore/Pages/Puzzles/Feedback.cshtml
```diff
@@ -8,11 +8,29 @@
...
+    var sumDiff = 0.0;
+    var count = 0;
+    foreach (var item in Model.Feedbacks) {
+        sumFun += item.Fun;
+        sumDiff += item.Difficulty;
+        count++;
+    }
+    sumFun /= count;
+    sumDiff /= count;
 }
```
jenetlan: This feels a bit out of place here - is this something that should be in the model instead? I'm not sure what the performance implications but usually we do the calculation in the model instead. Any reason you chose to have it here?

ServerCore/Pages/Puzzles/Feedback.cshtml
```diff
@@ -8,11 +8,29 @@
...
 
 <h2>@Html.DisplayFor(model => model.Puzzle.Name): Feedback</h2>
 
+<div>
+    <h3><b>Avg Fun Rating:</b> @sumFun</h3>
```
jenetlan: Very small nit - maybe call this sumFunScore / sumDiffScore instead? It's pretty easy to mis-read these as sumFunction and sumDifference instead :P

ServerCore/Pages/Puzzles/SubmitFeedback.cshtml
```diff
@@ -26,7 +26,7 @@
             </div>
             <div class="form-group">
                 <label asp-for="Feedback.Difficulty" class="control-label"></label>
-                <input asp-for="Feedback.Difficulty" type="range" min="0" max="10" value="5" step="1" class="slider" id="difficultySlider">
+                <input asp-for="Feedback.Difficulty" type="range" min="0" max="10" step="1" class="slider" id="difficultySlider">
```
jenetlan: Nit - should these be constants somewhere so we can enforce them on the model side as well?

ServerCore/Pages/Puzzles/SubmitFeedback.cshtml.cs
```diff
@@ -44,15 +55,33 @@ public async Task<IActionResult> OnPostAsync(int puzzleId)
...
+            {
+                editableFeedback.SubmissionTime = DateTime.UtcNow;
+                editableFeedback.Submitter = LoggedInUser;
+                editableFeedback.Difficulty = Feedback.Difficulty;
+                editableFeedback.Fun = Feedback.Fun;
+                editableFeedback.WrittenFeedback = Feedback.WrittenFeedback;
+                editableFeedback.Puzzle = await _context.Puzzles.Where(m => m.ID == puzzleId).FirstOrDefaultAsync();
+                if (editableFeedback.Puzzle == null)
+                {
+                    return NotFound();
```
jenetlan: Nit: Is it possible to add text to the NotFound constructor to indicate what was not found?

ServerCore/Pages/Puzzles/SubmitFeedback.cshtml.cs
```diff
@@ -44,15 +55,33 @@ public async Task<IActionResult> OnPostAsync(int puzzleId)
...
-            } 
+            Feedback editableFeedback = await _context.Feedback.Where((f) => f.Puzzle.ID == puzzleId && f.Submitter == LoggedInUser).FirstOrDefaultAsync();
+            if (editableFeedback == null)
+            {
+                Feedback.SubmissionTime = DateTime.UtcNow;
+                Feedback.Submitter = LoggedInUser;
+                Feedback.Puzzle = await _context.Puzzles.Where(m => m.ID == puzzleId).FirstOrDefaultAsync();
+                if (Feedback.Puzzle == null)
+                {
+                    return NotFound();
```
jenetlan: Nit: Is it possible to add text to the NotFound constructor to indicate what was not found?

Spartan97: Added

ServerCore/Pages/Puzzles/SubmitFeedback.cshtml.cs
Spartan97: Added

ServerCore/Pages/Puzzles/SubmitFeedback.cshtml
Spartan97: Adding min/max as constants and clamping them before adding the feedback to the DB.

ServerCore/DataModel/Feedback.cs
tabascq: @morganbr ? @asyasky ?

Spartan97: It looks like User->PuzzleUser was committed by Megan a while ago. I must not have been on a current version. So this should just merge cleanly without a data migration.

PR #212, opened 2019-01-24
Edit admin / authors per event

jenetlan:


ServerCore/Pages/Events/Players.cshtml.cs
```diff
@@ -12,28 +13,108 @@ namespace ServerCore.Pages.Events
...
+
+            IList<PuzzleUser> allAuthors = await (from author in _context.EventAuthors
+                                                  where author.Event == Event
+                                                  select author.Author).ToListAsync();
+
+            // note - if there's a faster way to count the puzzles / join all this data please let me know - Jenna
+            IList<IGrouping<int, PuzzleAuthors>> allPuzzles = await (from puzzleAuthor in _context.PuzzleAuthors
+                                                     where puzzleAuthor.Puzzle.Event == Event
+                                                     group puzzleAuthor by puzzleAuthor.Author.ID into puzzleGroup
+                                                     select puzzleGroup)
```
morganbr: You can do something like:
select new {Name = puzzleGroup.Key, Count = puzzlegroup.Count()}

You can then enumerate all authors and compare that to these results to find any that came back as zero. For an example, take a look at Teams/List.cshtml.cs

ServerCore/Pages/Events/AddAdminOrAuthor.cshtml.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+        }
+
+        public async Task<IActionResult> OnGetAsync(bool addAdmin)
+        {
+            this.addAdmin = addAdmin;
+
+            if (addAdmin)
+            {
+                Users = await (from user in _context.PuzzleUsers
```
morganbr: This is all users of all events of all time. Do we have any reasonable way to narrow it down?

ServerCore/Pages/Events/Players.cshtml.cs
jenetlan: Done :D Thank you :)

ServerCore/Pages/Events/Players.cshtml
```diff
@@ -1,64 +1,154 @@
...
+            </tr>
+        }
+    </tbody>
+</table>
+
+<h4>All author e-mails</h4>
+
+<p>@Model.AuthorEmails</p>
+
+<h3>Players</h3>
```
morganbr: Consider having this be separate or optional to avoid putting a ton of content on this page when it's not wanted.

ServerCore/Pages/Events/AddAdminOrAuthor.cshtml.cs
```diff
@@ -0,0 +1,108 @@
...
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.RazorPages;
+using Microsoft.AspNetCore.Mvc.Rendering;
+using Microsoft.EntityFrameworkCore;
+using ServerCore.DataModel;
+using ServerCore.ModelBases;
+
+namespace ServerCore.Pages.Events
+{
+    [Authorize("IsEventAdmin")]
```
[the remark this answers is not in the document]
jenetlan: I think that's something that we'll want to handle through a separate flow. I would see it working out such that we have a separate page where a global admin can add themselves as an event admin, and then they'll be able to access this page. I added an issue for that: https://github.com/PuzzleServer/mainpuzzleserver/issues/221

ServerCore/Pages/Events/AddAdminOrAuthor.cshtml.cs
jenetlan: Added an issue for that type of functionality because yes: https://github.com/PuzzleServer/mainpuzzleserver/issues/220

ServerCore/Pages/Events/Players.cshtml
jenetlan: Will do separately - good call :)

PR #223, opened 2019-01-26
Home FAQ Rules!

tabascq:
Adding a default homepagq, FAQ, and rules, as well as a PH20-specific set. The specific set of pages to use is controlled by a new HomePartial property on Event, so we can easily point all betas at the same content and easily add new content.


ServerCore/Pages/Resources/PH20/FAQPartial.cshtml
```diff
@@ -0,0 +1,98 @@
+﻿<div class="container">
+    <h1>FAQ</h1>
```
jenetlan: I'm going to assume that you copied this from somewhere and not review the actual text. If you'd like me to actually review the content lmk

ServerCore/Pages/Resources/Default/FAQPartial.cshtml
```diff
@@ -0,0 +1,3 @@
+﻿<div class="container">
+    <h2>Default FAQ!</h2>
```
jenetlan: wow....you take the bunnies out of this and then have the audacity to request a review....


...thankfully you have so much goodwill stocked up from making the demo event rabbit-themed that I can let this slide :P

Data/DataModel/Event.cs
```diff
@@ -1,79 +1,87 @@
...
+        /// </summary>
+        public string UrlString { get; set; }
+
+        [NotMapped]
+        public string EventID => UrlString ?? ID.ToString();
+
+        /// <summary>
+        /// The prefix of the partial files that provide the Home, FAQ, and Rules content.
+        /// </summary>
+        public string HomePartial { get; set; }
```
jenetlan: Do you have an example of how this would look? I'm not really sure how I would reference this in an event - an example string might help (maybe as a comment?)

PR #226, opened 2019-01-31
Updated css to un-fix the nav-bar

### LINKED ISSUE #210: Nav bar covers top of page when resized
When the nav bar gets resized, sometimes it goes to 2 or three lines, causing it to cover the top portion of the current page. 

Note: If it gets resized passed 767px, it collapses into a hamburger menu, which solves the problem.

lambertwang:
Fixes #210 
Removes the fixed display property from the navbar and the top padding from the page body.

### DIFF
--- a/ServerCore/wwwroot/css/site.css
+++ b/ServerCore/wwwroot/css/site.css
@@ -1,5 +1,4 @@
 body {
-    padding-top: 50px;
     padding-bottom: 20px;
 }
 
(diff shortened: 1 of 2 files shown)


jaylorch:
According to the documentation at https://getbootstrap.com/docs/3.4/components/#navbar, if you want to use navbar-fixed-top you need to include padding at the top, e.g., with body { padding-top: 70px; }.  Commit f4164ed removed that padding, though.

tabascq:
Not sure what is going on overall, but I tried this change locally and got a really silly looking result. Look at the empty space above the navbar:

[cid:image003.png@01D4ED63.8ED53670]

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

jaylorch:
How do I look at [cid:image003.png@01D4ED63.8ED53670]?

jaylorch:
Anyway, if that's not the fix, I don't know what is.  But there is still the issue that when the navbar runs over to two lines (which only happens at certain window widths, including the one on iPads), the top gets covered.

Issue #225, opened 2019-01-29
Submissions is mixing events


tabascq:
Showing submissions is showing all puzzles by an author regardless of event.

tabascq:
Addressed with #229

Issue #224, opened 2019-01-29
Demo event needs better lockout defaults


tabascq:
Currently it locks out at 0...

tabascq:
The demo event was fine, this was really an issue with edits to the event properties zapping these values. Addressed with #229.

Issue #192, opened 2019-01-20
Send sign-up link to teammates


jenetlan:
From the team members or add members page, add a way to send an e-mail to teammates that has a direct link to the 'apply' page for that team

tabascq:
Addressed with #229.

Issue #178, opened 2019-01-17
On the registration, show which team a user has applied to, if any


jenetlan:
On the /Teams/List page (which is where unregistered users go to sign up for a team), edit the "Join" column once a user has applied to a team to say something like "Request pending" to let the user know which team they've requested to join.

tabascq:
Addressed with #229.

PR #229, opened 2019-01-31
Fixes for: #225, #224, #192, #178, #145

tabascq:


ServerCore/Pages/Teams/List.cshtml.cs
```diff
@@ -48,6 +50,11 @@ public async Task<IActionResult> OnGetAsync()
                 UserOnTeam = false;
             }
 
+            var applications = from TeamApplication application in _context.TeamApplications
+                                                           where application.Player == LoggedInUser && application.Team.Event == Event
+                                                           select application.Team;
+            AppliedTeam = await applications.FirstOrDefaultAsync();
```
jenetlan: Any reason you did this on two lines? Usually we've been doing them in one. It's fine this way - I was just curious

ServerCore/Pages/Events/Edit.cshtml
```diff
@@ -125,24 +125,24 @@
...
-                <span asp-validation-for="Event.LockoutDurationMultiplier" class="text-danger"></span>
+                <label asp-for="EditableEvent.LockoutDurationMultiplier" class="control-label"></label>
+                <input asp-for="EditableEvent.LockoutDurationMultiplier" class="form-control" />
+                <span asp-validation-for="EditableEvent.LockoutDurationMultiplier" class="text-danger"></span>
             </div>
             <div class="form-group">
-                <label asp-for="Event.MaxSubmissionCount" class="control-label"></label>
-                <input asp-for="Event.MaxSubmissionCount" class="form-control" />
-                <span asp-validation-for="Event.MaxSubmissionCount" class="text-danger"></span>
+                <label asp-for="EditableEvent.MaxSubmissionCount" class="control-label"></label>
```
jenetlan: Does this actually change the defaults? The name switched, but I'm not seeing the corresponding place where the value changed

ServerCore/Pages/Events/Import.cshtml.cs
```diff
@@ -37,7 +37,11 @@ public async Task<IActionResult> OnPostImportAsync()
                 return NotFound();
             }
 
-            // TODO: validate that you are an admin of both events!!!
+            // verify that we're an admin of the import event. current event administratorship is already validated.
+            if (await _context.EventAdmins.Where(ea => ea.Event.ID == ImportEventID && ea.Admin == LoggedInUser).FirstOrDefaultAsync() == null)
```
morganbr: You can use ```if (await !_context.EventAdmins.Where(...).AnyAsync())``` to avoid actually pulling in the record.

ServerCore/Pages/Teams/List.cshtml.cs
tabascq: I copy/pasted this from elsewhere in the code. I think @morganbr uses this style. If I were to guess, I would say it's because when the query is written in this keyword-heavy syntax instead of a cascade of functions, you have to put the whole thing in parens to call FirstOrDefaultAsync and it'll look weird.

ServerCore/Pages/Events/Edit.cshtml
tabascq: No, the defaults were actually reasonable, but the first edit to the event properties would whack them.

Issue #128, opened 2018-12-07
Show hint coins in scoring


morganbr:
The player score page should show hint coins earned and used

tabascq:
Addressed with #231.

PR #231, opened 2019-02-02
Show Hints Used (#128)

tabascq:
Tabulate the number of hints used, and show that count as a sortable column in standings.

### DIFF
--- a/Data/DataModel/Team.cs
+++ b/Data/DataModel/Team.cs
@@ -32,14 +32,10 @@ namespace ServerCore.DataModel
         /// <summary>
         /// The number of hint coins this team currently has
         /// </summary>
-
-        /// <summary>
-        /// The number of hint coins this team currently has
-        /// </summary>
         public int HintCoinCount { get; set; }
 
         /// <summary>
-        /// The number of hint coins this team currently has
+        /// The number of hint coins this team currently has used
         /// </summary>
         public int HintCoinsUsed { get; set; }
     }
--- a/Data/DataModel/Team.cs
+++ b/Data/DataModel/Team.cs
@@ -32,6 +32,15 @@ namespace ServerCore.DataModel
         /// <summary>
         /// The number of hint coins this team currently has
         /// </summary>
+
+        /// <summary>
+        /// The number of hint coins this team currently has
+        /// </summary>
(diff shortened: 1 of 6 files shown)


Data/DataModel/Team.cs
```diff
@@ -29,9 +29,18 @@ public class Team
...
+        /// The number of hint coins this team currently has
+        /// </summary>
+
         /// <summary>
         /// The number of hint coins this team currently has
         /// </summary>
         public int HintCoinCount { get; set; }
+
+        /// <summary>
+        /// The number of hint coins this team currently has
```
jenetlan: In the GitHub diff this comment is showing up as being identical to the one above - should this one say "The number of hint coins this team has used" ?

Data/DataModel/Team.cs
```diff
@@ -29,9 +29,18 @@ public class Team
 
         public virtual List<Submission> Submissions { get; set; }
 
+        /// <summary>
```
morganbr: This looks like it got duplicated

ServerCore/Pages/Events/Standings.cshtml
```diff
@@ -33,6 +33,11 @@
                             @Html.DisplayNameFor(model => model.Teams[0].Score)
                         </a>
                     </th>
+                    <th>
+                        <a asp-page="./Standings" asp-route-sort="@(Model.SortForColumnLink(StandingsModel.SortOrder.HintsUsedAscending, StandingsModel.SortOrder.HintsUsedDescending))">
+                            Hints Used
```
morganbr: This would be the number of tickets, not the number of hints. It's an interesting point though... which is more interesting to display?

Data/DataModel/Team.cs
tabascq: Yep, copy/paste bug, will fix.

ServerCore/Pages/Events/Standings.cshtml
tabascq: I'll display tickets, will clarify the title.

Data/DataModel/Team.cs
tabascq: yep oops

Issue #185, opened 2019-01-17
Role-switcher should only show roles the user has access to


jenetlan:
In the /Shared/_RoleSwitcherPartial.cshtml file, right now the dropdown always includes every possible role. However, some users may not have access to author or admin roles, in which case they should not be included in the dropdown. Especially for players, the dropdown doesn't even need to exist (although the page title still should).

Kenny suggested adding IsEventAuthor and IsEventAdmin methods to the EventSpecificPageModel (similar to IsRegisteredUser) to help with this.

If it helps to move the dropdown label (the "Event name - Admin / Author" part) into the _RoleSwitcherPartial.cshtml file, feel free to do so :)

tabascq:
Addressed with #232.

PR #232, opened 2019-02-04
Role Switcher Filter (#185)

tabascq:


ServerCore/Pages/Shared/_RoleSwitcherPartial.cshtml
```diff
@@ -1,24 +1,40 @@
...
+    {
     <li><a asp-page=@AdminRoute asp-route-id="@Event.ID" asp-route-eventRole="@ModelBases.EventRole.admin" asp-route-teamId=@RoutingTeamId asp-route-puzzleId=@RoutingPuzzleId>Admin mode</a></li>
     <li class="divider"></li>
+    }
+    @if (IsAuthor)
+    {
     <li><a asp-page=@AuthorRoute asp-route-id="@Event.ID" asp-route-eventRole="@ModelBases.EventRole.author" asp-route-teamId=@RoutingTeamId asp-route-puzzleId=@RoutingPuzzleId>Author mode</a></li>
     <li class="divider"></li>
+    }
     <li><a asp-page=@PlayRoute asp-route-id="@Event.ID" asp-route-eventRole="@ModelBases.EventRole.play" asp-route-teamId=@RoutingTeamId asp-route-puzzleId=@RoutingPuzzleId>Player mode</a></li>
```
jenetlan: We'll want to hide this item / possibly hide the dropdown overall if the user isn't an author or an admin - users who are just players don't need to know that role switching exists and a non-functional dropdown could be confusing. We can tackle that separately though if you'd like - this is still a huge step forward :)

tabascq: Note the IsAdmin || IsAuthor check at line 11 - we do hide the dropdown in those cases.

tabascq: See line 11 above - I already added this check.

Issue #228, opened 2019-01-31
Enforce team and player registration limits


tabascq:
Currently all registration limits are zero and so we clearly aren't enforcing a thing. We should start.

We should also set up defaults for new events.

        public int MaxNumberOfTeams { get; set; }
        public int MaxTeamSize { get; set; }
        public int MaxExternalsPerTeam { get; set; }

tabascq:
Addressed with #233

PR #233, opened 2019-02-05
Team Registration Limits (#228)

tabascq:
Applying limits on number of teams, number of players, and number of externals per team.

Admins have god powers and can add teams over the limit, players over the limit, and externals over the limit - but players cannot.

[We aren't validating the employee aliases but we could choose to manually audit them if desired.]

### DIFF
--- a/ServerCore/Pages/Teams/Index.cshtml
+++ b/ServerCore/Pages/Teams/Index.cshtml
@@ -11,77 +11,90 @@
 <h2>All Teams</h2>
 
 <p>
-    @if (Model.Event.IsTeamRegistrationActive && Model.EventRole == ModelBases.EventRole.admin)
+    @if (Model.EventRole == ModelBases.EventRole.admin)
     {
+        if (!Model.Event.IsTeamRegistrationActive)
+        {
+            <div class="alert alert-danger" role="alert">
+                This event is not currently open for registration.
+            </div>
+        }
+        else if (Model.Event.MaxNumberOfTeams <= Model.Teams.Count)
+        {
+            <div class="alert alert-danger" role="alert">
+                This event is full.
+            </div>
+        }
+
         <a asp-page="Create">Create New</a>
     }
-    else
-    {
-        <div class="alert alert-danger" role="alert">
-            This event is not currently open for registration.
-        </div>
-    }
 </p>
+<h4>Registered Teams: @Model.Teams.Count/@Model.Event.MaxNumberOfTeams, Registered Players: @Model.PlayerCount/@(Model.Event.MaxNumberOfTeams * Model.Event.MaxTeamSize)</h4>
 <table class="table">
     <thead>
         <tr>
             <th>
-                @Html.DisplayNameFor(model => model.Teams[0].Name)
+                @Html.DisplayNameFor(model => model.Teams.FirstOrDefault().Key.Name)
             </th>
             <th>
-                @Html.DisplayNameFor(model => model.Teams[0].RoomID)
+                Size
             </th>
             <th>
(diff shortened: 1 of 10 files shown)


ServerCore/Pages/Teams/Index.cshtml
```diff
@@ -11,77 +11,90 @@
...
-    @if (Model.Event.IsTeamRegistrationActive && Model.EventRole == ModelBases.EventRole.admin)
+    @if (Model.EventRole == ModelBases.EventRole.admin)
     {
+        if (!Model.Event.IsTeamRegistrationActive)
+        {
+            <div class="alert alert-danger" role="alert">
+                This event is not currently open for registration.
+            </div>
+        }
+        else if (Model.Event.MaxNumberOfTeams <= Model.Teams.Count)
```
jenetlan: Super small nit: I think this gets easier to read if you reverse it like you had for players above:
(Model.Teams.Count >= Model.Event.MaxNumberOfTeams)
It reads closer to the way you'd say it in a sentence:
"If the number of teams is greater than or equal to the max allowed number of teams"
The sentence the other way sounds awkward because the maximum is a constant so it's odd to check it relative to something else:
"If the max allowed number of teams is less than or equal to the number of teams"
Anecdotally at least, it feels like it took me longer to understand / double check this one than the ones above

jenetlan: On the plus side I absolutely love the super descriptive error messages - thanks for updating all of those :D

ServerCore/Pages/Teams/Index.cshtml
```diff
@@ -11,77 +11,90 @@
...
+            <th>
+                Size
             </th>
             <th>
-                @Html.DisplayNameFor(model => model.Teams[0].RoomID)
+                @Html.DisplayNameFor(model => model.Teams.FirstOrDefault().Key.RoomID)
             </th>
             <th>
-                @Html.DisplayNameFor(model => model.Teams[0].CustomRoom)
+                @Html.DisplayNameFor(model => model.Teams.FirstOrDefault().Key.CustomRoom)
```
jenetlan: Should we add an issue to go through and fix this elsewhere? I'm pretty sure this pattern is in a few other places as well

ServerCore/Pages/Teams/TeamListBase.cs
```diff
@@ -0,0 +1,49 @@
...
+        public TeamListBase(PuzzleServerContext serverContext, UserManager<IdentityUser> userManager) : base(serverContext, userManager)
+        {
+        }
+
+        /// <summary>
+        /// All teams
+        /// </summary>
+        public Dictionary<Team, int> Teams { get; set; }
+
+        public int PlayerCount { get; set; }
```
jenetlan: It might help to have a comment here that this is players across all teams - on first read it looks like it could be for just one team, and it was only after jumping down to the function that I realized it was for everyone (although admittedly it wouldn't make sense to have it here for just one team)

ServerCore/Pages/Teams/Index.cshtml
tabascq: The issue here is that Teams stopped being an IList, so the previous code actually broke. This wasn't me being pedantic.

tabascq: To be clearer: I don't think there was anything wrong with the previous code so I don't think we need to make a pass for this.

ServerCore/Pages/Teams/Index.cshtml
tabascq: Fixed in both places, thanks!

Issue #168, opened 2019-01-17
Add user to team in demo event


morganbr:
When the demo event gets created, put the user on a team so they can access play pages.

tabascq:
fixed with #234

Issue #123, opened 2018-12-07
Put demo event creation in a transactionscope


morganbr:
Creating a demo event is likely to fail as we test with bugs. It would be good to put the creation in a transactionscope so we don't get partially created broken events.

tabascq:
Fixed with #234

PR #234, opened 2019-02-07
Demo Fixes - #123, #168

tabascq:
Putting the whole demo in a transaction and also letting you immediately put yourself on a team to save time.

### DIFF
--- a/ServerCore/Pages/Events/CreateDemo.cshtml
+++ b/ServerCore/Pages/Events/CreateDemo.cshtml
@@ -31,22 +31,30 @@
             <div class="form-group">
                 <div class="checkbox">
                     <label>
+                        <input asp-for="AddCreatorToLoneWolfTeam" /> @Html.DisplayNameFor(model => model.AddCreatorToLoneWolfTeam)
+                    </label>
+                </div>
+            </div>
+            <div class="form-group">
+                <div class="checkbox">
+                    <label>
                         <input asp-for="StartTheEvent" /> @Html.DisplayNameFor(model => model.StartTheEvent)
                     </label>
                 </div>
             </div>
-            <br/>
+            <br />
             <div class="form-group">
                 <b>Usage:</b>
                 <ul>
                     <li>The default event start time is now (in UTC), and the default event end is one day later. Adjust them now if you want.</li>
                     <li>Related times will update to roughly matching values.</li>
                     <li>All puzzles have a partial named PARTIAL, and an answer named ANSWER.</li>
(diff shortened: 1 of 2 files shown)


ServerCore/Pages/Events/CreateDemo.cshtml.cs
```diff
@@ -57,194 +61,223 @@ public async Task<IActionResult> OnPostAsync()
...
+                    _context.PuzzleAuthors.Add(new PuzzleAuthors() { Puzzle = hard, Author = demoCreatorUser });
+                    _context.PuzzleAuthors.Add(new PuzzleAuthors() { Puzzle = meta, Author = demoCreatorUser });
+                }
+
+                // TODO: Files (need to know how to detect whether local blob storage is configured)
+                // Is there a point to adding Feedback or is that quick/easy enough to demo by hand?
+
+                await _context.SaveChangesAsync();
+
+                if (teamLoneWolf != null)
```
jenetlan: Is there a reason this isn't just part of the if statement on line 225? Seems like it could be done there - not sure what I'm missing

ServerCore/Pages/Events/CreateDemo.cshtml
```diff
@@ -28,25 +28,33 @@
                 <input asp-for="Event.AnswerSubmissionEnd" class="form-control" />
                 <span asp-validation-for="Event.AnswerSubmissionEnd" class="text-danger"></span>
             </div>
+            <div class="form-group">
+                <div class="checkbox">
+                    <label>
+                        <input asp-for="AddCreatorToLoneWolfTeam" /> @Html.DisplayNameFor(model => model.AddCreatorToLoneWolfTeam)
```
jenetlan: Do you know what this defaults to? It would be nice if we could default to yes since I'm guessing the common test case is to check on puzzles from a player's perspective rather than to test team registration from a player's perspective

ServerCore/Pages/Events/CreateDemo.cshtml.cs
tabascq: It needs to be somewhere after a SaveChangesAsync() in order to be valid - that's what gets the team ID set up.

ServerCore/Pages/Events/CreateDemo.cshtml
tabascq: So far it's defaulting to false just like the "start the event" checkbox. Can turn these to true later if the group prefers.

Issue #180, opened 2019-01-17
Make puzzle names in play UI link to files


morganbr:
The play UI currently has a separate column for puzzle links. Instead, the puzzle title should link to the file and we can remove the separate column.

tabascq:
Addressed with #237.

Issue #179, opened 2019-01-17
Remove OrderInGroup from play UI


morganbr:
The play page currently includes the OrderInGroup column, which should just be used for sorting, not displayed.

tabascq:
Addressed with #237.

PR #237, opened 2019-02-08
PolishPlayPage (#179, #180)

tabascq:
Reducing columns in this UX, as well as resorting so the group (e.g. module) comes first and is the default sort key. This will suit PH20 better.

### DIFF
--- a/ServerCore/Pages/Teams/Play.cshtml.cs
+++ b/ServerCore/Pages/Teams/Play.cshtml.cs
@@ -29,7 +29,7 @@ namespace ServerCore.Pages.Teams
 
         public SortOrder? Sort { get; set; }
 
-        private const SortOrder DefaultSort = SortOrder.PuzzleAscending;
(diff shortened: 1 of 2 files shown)


ServerCore/Pages/Teams/Play.cshtml.cs
```diff
@@ -29,7 +29,7 @@ public PlayModel(PuzzleServerContext serverContext, UserManager<IdentityUser> us
 
         public SortOrder? Sort { get; set; }
 
-        private const SortOrder DefaultSort = SortOrder.PuzzleAscending;
+        private const SortOrder DefaultSort = SortOrder.GroupAscending;
```
morganbr: Should this have a secondary sort using OrderInGroup?

tabascq: It does, see lines 68 and 71 of the same file.

morganbr: Ah, github reviews. Looks good.

Issue #184, opened 2019-01-17
Add PuzzleUser view / edit page (not create or delete)


jenetlan:
Players and Admins should be able to view and edit some PuzzleUser properties. Specifically:

- Name
- EmployeeAlias
- PhoneNumber
- TShirtSize
- VisibleToOthers

This page should be linked from each row in the /Events/Players page for admins and it should probably be its own item in the navigation header for now (you can talk to Megan about whether or not it should be added to the other user properties but we can always do that later)

tabascq:
Addressed with #240

PR #240, opened 2019-02-11
Edit Player Properties (#184)

tabascq:
Updated the profile page for the logged in user to update all their properties.

Also added a page that can be accessed from the team members list or the admin users list, which allows editing if the user is the current user or a global admin.

Along the way, fixed a bug where editing could clear the global admin flag.

### DIFF
--- a/ServerCore/Areas/Identity/Pages/Account/Edit.cshtml
+++ b/ServerCore/Areas/Identity/Pages/Account/Edit.cshtml
@@ -16,6 +16,7 @@
     <div class="col-md-4">
         <form method="post" asp-route-returnUrl="@Model.ReturnUrl">
             <div asp-validation-summary="ModelOnly" class="text-danger"></div>
+            @* NOTE: If you add more properties here, please add them to ExternalLogin.cshtml as well! *@
             <input type="hidden" asp-for="PuzzleUser.ID" />
             <input type="hidden" asp-for="PuzzleUser.IdentityUserId" />
             <input type="hidden" asp-for="PuzzleUser.IsGlobalAdmin" />
@@ -49,6 +50,7 @@
                 <input asp-for="PuzzleUser.TShirtSize" class="form-control" />
                 <span asp-validation-for="PuzzleUser.TShirtSize" class="text-danger"></span>
             </div>
+            @* NOTE: If you add more properties here, please add them to ExternalLogin.cshtml as well! *@
             <div class="form-group">
                 <input type="submit" value="Save" class="btn btn-default" />
             </div>
--- a/ServerCore/Areas/Identity/Pages/Account/Edit.cshtml
+++ b/ServerCore/Areas/Identity/Pages/Account/Edit.cshtml
@@ -1,0 +1,62 @@
+﻿@page
+@model EditModel
+@{
+    ViewData["Title"] = "Edit User Properties";
+}
+
+<h2>@ViewData["Title"]</h2>
+<h4>Edit Properties for @Model.PuzzleUser.Name</h4>
+<hr />
+
+<p class="text-info">
+    This profile is shared between all events.
+</p>
+
+<div class="row">
+    <div class="col-md-4">
+        <form method="post" asp-route-returnUrl="@Model.ReturnUrl">
+            <div asp-validation-summary="ModelOnly" class="text-danger"></div>
+            <input type="hidden" asp-for="PuzzleUser.ID" />
+            <input type="hidden" asp-for="PuzzleUser.IdentityUserId" />
+            <input type="hidden" asp-for="PuzzleUser.IsGlobalAdmin" />
+            <div class="form-group">
+                <label asp-for="PuzzleUser.Email"></label>
+                <input asp-for="PuzzleUser.Email" class="form-control" />
+                <span asp-validation-for="PuzzleUser.Email" class="text-danger"></span>
+            </div>
+            <div class="form-group">
+                <label asp-for="PuzzleUser.Name"></label>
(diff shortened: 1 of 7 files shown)


ServerCore/Areas/Identity/Pages/Account/Edit.cshtml.cs
```diff
@@ -0,0 +1,79 @@
...
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Identity;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.RazorPages;
+using Microsoft.EntityFrameworkCore;
+using ServerCore.DataModel;
+
+namespace ServerCore.Areas.Identity.Pages.Account
+{
+    [AllowAnonymous]
```
jenetlan: Can we only allow this for global admins? For players it would be good to always send them to the same place to edit their user settings

ServerCore/Areas/Identity/Pages/Account/Edit.cshtml.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        [BindProperty]
+        public PuzzleUser PuzzleUser { get; set; }
+
+        public string ReturnUrl { get; set; }
+
+        public async Task<IActionResult> OnGetAsync(int userId, string returnUrl = null)
+        {
+            var thisPuzzleUser = await PuzzleUser.GetPuzzleUserForCurrentUser(_context, User, _userManager);
```
jenetlan: If we always redirect the user to the built-in settings page that simplifies these logic checks quite a bit since it will only allow global admins :)

ServerCore/Areas/Identity/Pages/Account/Edit.cshtml.cs
```diff
@@ -0,0 +1,79 @@
+﻿using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Identity;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.RazorPages;
+using Microsoft.EntityFrameworkCore;
+using ServerCore.DataModel;
+
+namespace ServerCore.Areas.Identity.Pages.Account
```
jenetlan: Can you add a comment here and on the other account settings page warning us to keep them both in sync?

ServerCore/Areas/Identity/Pages/Account/Edit.cshtml.cs
```diff
@@ -0,0 +1,79 @@
...
+            ReturnUrl = returnUrl;
+
+            return Page();
+        }
+
+        public async Task<IActionResult> OnPostAsync(string returnUrl = null)
+        {
+            if (!ModelState.IsValid)
+            {
+                // TODO: For some reason the ID property is not reported as valid. No clue how validation works and all other properties are strings anyway. Ignore for now?
```
jenetlan: @asyasky @morganbr ? any ideas?

ServerCore/Areas/Identity/Pages/Account/Edit.cshtml
```diff
@@ -0,0 +1,62 @@
...
+<p class="text-info">
+    This profile is shared between all events.
+</p>
+
+<div class="row">
+    <div class="col-md-4">
+        <form method="post" asp-route-returnUrl="@Model.ReturnUrl">
+            <div asp-validation-summary="ModelOnly" class="text-danger"></div>
+            <input type="hidden" asp-for="PuzzleUser.ID" />
+            <input type="hidden" asp-for="PuzzleUser.IdentityUserId" />
```
morganbr: Please validate that this hasn't been tampered with

ServerCore/Areas/Identity/Pages/Account/Edit.cshtml.cs
tabascq: I will remove this comment, as it is stale and reflected a misunderstanding at the time that I wrote it. Remain calm! All is well!

ServerCore/Areas/Identity/Pages/Account/Edit.cshtml.cs
tabascq: OK, I'll restrict this.

ServerCore/Areas/Identity/Pages/Account/Edit.cshtml.cs
tabascq: Yep, I'll change the policy attribute and then remove all the checks.

ServerCore/Areas/Identity/Pages/Account/Edit.cshtml
tabascq: I don't understand what this comment means.

tabascq: Name-dropping @morganbr  so he sees this.

ServerCore/Areas/Identity/Pages/Account/Edit.cshtml.cs
tabascq: done

jenetlan:
Thanks for the updates! :D

ServerCore/Areas/Identity/Pages/Account/Edit.cshtml
tabascq: Sitting next to Morgan - comment withdrawn since this is an admin only page.

PR #244, opened 2019-02-14
Cheats and Whistles

tabascq:
Support for cheat codes (aka Fast Forwards) and global lockouts (aka Whistle Stops).

Also fixed a few trivial bugs found along the way while testing these features.


ServerCore/PuzzleStateHelper.cs
```diff
@@ -185,7 +185,7 @@ public static async Task SetUnlockStateAsync(PuzzleServerContext context, Event
             }
 
             // Award hint coins
-            if (value != null && puzzle.HintCoinsForSolve != 0)
+            if (value != null && puzzle != null && puzzle.HintCoinsForSolve != 0)
```
morganbr: Why would puzzle be null?

tabascq: null means "all puzzles". In theory I would amass all hints in this case but what would you spend them on?

PR #246, opened 2019-02-20
Fix #188

### LINKED ISSUE #188: Can't add author to a puzzle
SqlException: The INSERT statement conflicted with the FOREIGN KEY constraint "FK_PuzzleAuthors_Puzzles_PuzzleID". The conflict occurred in database "PuzzleServer", table "dbo.Puzzles", column 'ID'.
The statement has been terminated.

tabascq:
Unsure how this code ever worked at all. Maybe it only worked in the empty DB case. Regardless, it works now.

### DIFF
--- a/ServerCore/Pages/Puzzles/Edit.cshtml.cs
+++ b/ServerCore/Pages/Puzzles/Edit.cshtml.cs
@@ -118,12 +118,12 @@ namespace ServerCore.Pages.Puzzles
 
         public async Task<IActionResult> OnPostAddAuthorAsync()
         {
-            if (!(await _context.EventAuthors.Select(m => m.Author.ID == NewAuthorID && m.Event == Event).AnyAsync()))
+            if (!(await _context.EventAuthors.Where(m => m.Author.ID == NewAuthorID && m.Event == Event).AnyAsync()))
             {
                 return NotFound();
             }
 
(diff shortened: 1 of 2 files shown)


ServerCore/Pages/Puzzles/Edit.cshtml.cs
```diff
@@ -118,12 +118,12 @@ public async Task<IActionResult> OnPostAsync()
...
         public async Task<IActionResult> OnPostAddAuthorAsync()
         {
-            if (!(await _context.EventAuthors.Select(m => m.Author.ID == NewAuthorID && m.Event == Event).AnyAsync()))
+            if (!(await _context.EventAuthors.Where(m => m.Author.ID == NewAuthorID && m.Event == Event).AnyAsync()))
             {
                 return NotFound();
             }
 
-            if (!(await _context.PuzzleAuthors.Select(m => m.PuzzleID == Puzzle.ID && m.AuthorID == NewAuthorID).AnyAsync()))
+            if (!(await _context.PuzzleAuthors.Where(m => m.PuzzleID == Puzzle.ID && m.AuthorID == NewAuthorID).AnyAsync()))
```
morganbr: Rather than use and roundtrip Puzzle.ID, could you just accept puzzleID as a parameter to the method since it's already in the route? That takes care of auth for this too.

PR #248, opened 2019-02-21
Fix Team Creation

tabascq:


tabascq:
Got both Create and Edit working

morganbr:
Obsoleted by #250

PR #250, opened 2019-02-24
Missing fixes from DBIntegrity

morganbr:
I found a set of fixes that I somehow didn't push with DBIntegrity. This should fix team creation, editing and deletion, puzzle deletion, and feedback submission.


morganbr:
@tabascq, I know you had some of the team fixes in #248, but I realized there were more bugs when I found my change.

ServerCore/Pages/Teams/Delete.cshtml.cs
```diff
@@ -37,7 +37,7 @@ public async Task<IActionResult> OnPostAsync(int teamId)
                 await TeamHelper.DeleteTeamAsync(_context, Team);
             }
 
-            return RedirectToPage("./Index");
+            return RedirectToPage("./List");
```
tabascq: is this destination right for admins as well? Other places have needed to check which EventRole is active and I'm not sure if this is one of them.

morganbr: Changed to send admins to index (list works too, but index is more useful)

morganbr:
@tabascq, yep you can just close that one out.

morganbr:
Thanks for the review!

PR #262, opened 2019-03-04
Support for sync (of pieces, annotations, and solved puzzles)

jaylorch:
This check-in adds support for sync.

Sync allows a user (well, really Javascript on behalf of the user) to post a sync request to /{eventId}/api/Sync/{puzzleId}.  Such a sync request can include a list of puzzle IDs to query the solved status of, an encoded list of new annotations to upload, a minimum number of solves, and a last sync time.  The minimum number of solves n, if present, means "please give me any puzzle pieces that I've earned for having solve count n or higher".  The last sync time means "don't bother giving me any annotations earlier than this; I have all of those".  The response to a sync includes a list of puzzles solved (among the ones requested), a list of all annotations (since the last sync time), a list of pieces earned, and the maximum solve count reflected in those pieces.

To test it, you can use the three puzzles and metapuzzle in the Sync Test group that are automatically created in the demo.  Go to /{eventId}/{puzzleId}/SyncTest.html, where {puzzleId} is the ID of the puzzle named "Sync Test".  The model for that page will automatically include the initial sync requests so that the page doesn't need to do an immediate sync.  The page will sync every minute, and whenever the user clicks the sync button.  It also allows the user to enter answers and assignments that are uploaded as assignments; to see sync working, you can open it in two tabs.  To make more pieces get revealed to the page, you'll need to mark puzzles solved; the solve count is equal to the number of puzzles worth at least 10 points that aren't in the Sync Test group.


ServerCore/SyncController.cs
```diff
@@ -0,0 +1,565 @@
...
+
+            /// <summary>
+            ///   If an error was encountered while processing the raw parameters passed to the sync
+            ///   controller, this will be a non-null string that can be returned to the user to
+            ///   give a meaningful error message.  It's intended that if the request comes from a
+            ///   page we created, no error will ever be raised.
+            /// </summary>
+            public readonly string DecodingError;
+
+            public DecodedSyncRequest(List<int> query_puzzle_ids, int? min_solve_count, string annotations, string last_sync_time,
```
morganbr: Why is this doing custom deserialization rather than just using Newtonsoft Json to deserialize the whole class?

morganbr:
@jaylorch, can you please talk us through this in person during the next meeting?

ServerCore/SyncController.cs
```diff
@@ -0,0 +1,565 @@
...
+        ///   itself if it's different from the last time it found out the version.
+        ///
+        ///   So, the point of GetSiteVersion is to return a string that changes with every update
+        ///   to the page of the corresponding puzzle.
+        /// </summary>
+        private string GetSiteVersion(Puzzle puzzle)
+        {
+            // if (puzzle.ID == 8) { return "2" };   // use a line like this to indicate a puzzle-specific version
+            // if (puzzle.ID == 97) { return "3" };  // use a line like this to indicate a puzzle-specific version
+            return "1";
```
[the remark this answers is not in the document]
jaylorch: The point here is to only update this *if* we update the code during the event.  I want the update to the code for the page to be atomic with the update to this version number.  After all, this is the version of the *code*.

ServerCore/SyncController.cs
jaylorch: HTTP allows you to POST a set of variables, some of which are optional, so I'm taking advantage of that.  Essentially, my API has a set of inputs, and I use a separate POST variable for each of them.

jaylorch: Also, I don't want to worry about different browser implementations of Javascript encoding DateTimes differently, so I do all the serialization/deserialization of DateTimes on the server.

ServerCore/Pages/Events/CreateDemo.cshtml.cs
```diff
@@ -31,6 +31,12 @@ public CreateDemoModel(PuzzleServerContext context, UserManager<IdentityUser> us
             _userManager = userManager;
         }
 
+        private Piece MakePiece(Puzzle puzzle, int progressLevel, int clueID, string answerPattern, int puzzlePos, string clue)
+        {
+            string contents = $"{{\"clue_id\": {clueID}, \"answer_pattern\": \"{answerPattern}\", \"puzzle_pos\": {puzzlePos}, \"clue\": \"{clue}\"}}";
```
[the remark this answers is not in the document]
jaylorch: Good point. Done.

ServerCore/Pages/Pieces/SyncTest.cshtml.cs
```diff
@@ -0,0 +1,62 @@
...
+        public int HeatSyncId { get; set; }
+        public int LipSyncId { get; set; }
+        public int MetapuzzleId { get; set; }
+        public string InitialSyncString { get; set; }
+
+        public SyncTestModel(PuzzleServerContext serverContext, UserManager<IdentityUser> userManager): base(serverContext, userManager)
+        {
+        }
+
+        //TODO: Not working [Authorize(Policy = "PlayerCanSeePuzzle")]
```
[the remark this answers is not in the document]
jaylorch: The error seems to be gone, so I put the policy back.

ServerCore/SyncController.cs
```diff
@@ -0,0 +1,565 @@
...
+                // string contents.  Or at least, it's supposed to be; we have to check it carefully.
+
+                AnnotationRequests = null;
+                if (annotations != null) {
+                    List<dynamic> annotations_decoded = null;
+
+                    // First, decode it into a list using the JSON deserializer.
+                    
+                    try {
+                        annotations_decoded = JsonConvert.DeserializeObject<List<dynamic>>(annotations);
```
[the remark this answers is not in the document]
jaylorch: Good idea. Done.

ServerCore/SyncController.cs
```diff
@@ -0,0 +1,534 @@
...
+                                         select state.Puzzle).ToListAsync();
+
+            int maxSolveCount = 0;
+            List<int> solvedPuzzles = new List<int>();
+            foreach (var solvedPuzzle in solves) {
+                // I'm pretty sure the query above will never return any puzzles that aren't
+                // from the current event.  But, just in case, out of paranoia, we check
+                // for them and don't count them.
+
+                if (solvedPuzzle.Event.ID != eventId) {
```
tabascq: just add this to the where clause

ServerCore/SyncController.cs
```diff
@@ -0,0 +1,534 @@
...
+                    continue;
+                }
+
+                // If the request is asking whether certain puzzles are solved, check if it's
+                // in the set and, if so, put it in the list of solved puzzles to inform the
+                // requester of.
+                    
+                if (queryPuzzleIdSet != null && queryPuzzleIdSet.Contains(solvedPuzzle.ID)) {
```
tabascq: solveValue >= 10 check here?

ServerCore/SyncController.cs
```diff
@@ -0,0 +1,534 @@
...
+        /// <summary>
+        ///   This routine stores in the database any annotations the requester has uploaded.
+        /// </summary>
+        private async Task StoreAnnotations(DecodedSyncRequest request, SyncResponse response, int puzzleId, int teamId)
+        {
+            if (request.AnnotationRequests == null) {
+                return;
+            }
+
+            foreach (var annotationRequest in request.AnnotationRequests) {
```
tabascq: If you wrap the whole thing in a transaction does that help with the concurrency problems and the special SQL etc.?

jaylorch: Unfortunately, no, because SQL's default setting is read committed, not strict serialization.  If you have the database set up for strict serialization, though, let me know and I'd be happy to use a transaction instead!

tabascq: As if I had any clue :-)

ServerCore/SyncController.cs
jaylorch: All that we're doing here is responding to a question the requester has about which puzzles have been solved.  For instance, the requester might want to know if a certain Fast Forward was taken, and that Fast Forward might have a solve value under 10.  So we want the requester to be able to ask such questions even about puzzles whose solve value is under 10.

ServerCore/Pages/Pieces/SyncTest.cshtml
```diff
@@ -0,0 +1,2283 @@
+@page "/{eventId}/{puzzleId}/SyncTest.html"
```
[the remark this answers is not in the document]
jaylorch: I think it's fine, but just to be extra-safe I'll change the Policy to IsEventAdmin.

ServerCore/SyncController.cs
```diff
@@ -0,0 +1,534 @@
...
+                else {
+                    try {
+                        LastSyncTime = JsonConvert.DeserializeObject<DateTime>(last_sync_time);
+                    }
+                    catch (JsonException) {
+                        LastSyncTime = null;
+                    }
+                }
+
+                var my_annotations = new List<AnnotationRequest> { new AnnotationRequest { key = 7, contents = "foo" } };
```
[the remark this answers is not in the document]
jaylorch: Oops, yeah, those lines aren't important any more.

ServerCore/SyncController.cs
```diff
@@ -0,0 +1,534 @@
...
+                    try {
+                        // First, decode it into a list using the JSON deserializer.
+                        AnnotationRequests = JsonConvert.DeserializeObject<List<AnnotationRequest>>(annotations);
+                    }
+                    catch (JsonException) {
+                        response.AddError("Could not deserialize annotations list");
+                        AnnotationRequests = null;
+                    }
+
+                    // If that succeeded, check each of the lements for validity.
```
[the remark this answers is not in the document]
jaylorch: I'll fix this in the next version, which I'm about to push.

ServerCore/SyncController.cs
```diff
@@ -0,0 +1,534 @@
...
+        }
+
+        /// <summary>
+        ///   A SyncResponse gathers all the data that needs to be sent back to the requester of a
+        ///   sync, and supplies it as an IActionResult when the caller calls GetResult().
+        /// </summary>
+        public class SyncResponse
+        {
+            private string currentError;
+            private dynamic response;
```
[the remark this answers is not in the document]
jaylorch: I'll update it to use a Dictionary<string, object> instead, and insert a comment explaining why.  I'll also fix the comment saying I'm returning an IActionResult.

ServerCore/SyncController.cs
jaylorch: Good point, I will.

PR #263, opened 2019-03-04
Puzzle picker

morganbr:
Adds a puzzle picker page for PH20 whistle stops. This will allow teams to choose between different whistle stops (and could conceivably be used for other puzzle choices in the future). Also adds a Description field to puzzles that shows up both on the picker page and on the admin/author pages.


Data/Migrations/PuzzleServerContextModelSnapshot.cs
```diff
@@ -474,6 +475,8 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                         .ValueGeneratedOnAdd()
                         .HasAnnotation("SqlServer:ValueGenerationStrategy", SqlServerValueGenerationStrategy.IdentityColumn);
 
+                    b.Property<string>("Description");
```
morganbr: Put this in manually and revert everything else

ServerCore/Pages/EventSpecific/PH20/PuzzlePicker.cshtml.cs
```diff
@@ -0,0 +1,139 @@
...
+            }
+
+            // Restrict this page to whistle stops
+            if (puzzle.MinutesOfEventLockout == 0)
+            {
+                return NotFound();
+            }
+
+            Team team = await UserEventHelper.GetTeamForPlayer(_context, Event, LoggedInUser);
+            var puzzleStateQuery = PuzzleStateHelper.GetFullReadOnlyQuery(_context, Event, null, team);
```
tabascq: pass "puzzle" for param 3 and then you do not need the where clause below.

tabascq: oh - I see you reuse this query below and that's why you did it like this. Well, you can certainly remove the team filter from the where clause since you specify it here.

PR #276, opened 2019-03-07
Faster Puzzle Entry

tabascq:
Streamlined the Create page to only a few fields and a selection of puzzle types that fills out defaults for many others.

Also streamlined the Edit page to separate properties into basic and advanced buckets, and shorten the overall height so that the Prerequisites section is visible.


ServerCore/Pages/Puzzles/Create.cshtml.cs
```diff
@@ -20,31 +20,82 @@ public CreateModel(PuzzleServerContext serverContext, UserManager<IdentityUser>
...
 
             await _context.SaveChangesAsync();
 
-            return RedirectToPage("./Index");
+            return RedirectToPage("./Edit", new { puzzleId = p.ID });
+        }
+
+        public class MiniPuzzle
+        {
+            public string Name { get; set; }
```
morganbr: If you put ```[Required]``` on this, the page will validate it

PR #278, opened 2019-03-07
Auth checks & fixes

jenetlan:
I went through all of the pages I know of in the site and checked that the appropriate auth was in place. I believe all of the checks are currently correct. I also had to update the routes for some team pages to include the team id for the auth checks to work properly.

### DIFF
--- a/ServerCore/Pages/Teams/Create.cshtml.cs
+++ b/ServerCore/Pages/Teams/Create.cshtml.cs
@@ -33,7 +33,7 @@ namespace ServerCore.Pages.Teams
 
             if (EventRole == EventRole.play && GetTeamId().Result != -1)
             {
-                return Forbid();
+                return NotFound("You are already on a team and cannot create a new one.");
             }
 
             if (EventRole == EventRole.admin && !await LoggedInUser.IsAdminForEvent(_context, Event))
--- a/ServerCore/Pages/Teams/Create.cshtml.cs
+++ b/ServerCore/Pages/Teams/Create.cshtml.cs
@@ -28,7 +28,12 @@ namespace ServerCore.Pages.Teams
 
             if (EventRole != EventRole.play && EventRole != EventRole.admin)
             {
-                return NotFound();
+                return Forbid();
+            }
+
+            if (EventRole == EventRole.play && GetTeamId().Result != -1)
+            {
+                return Forbid();
             }
 
             if (EventRole == EventRole.admin && !await LoggedInUser.IsAdminForEvent(_context, Event))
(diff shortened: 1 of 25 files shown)


ServerCore/Pages/Teams/Create.cshtml.cs
```diff
@@ -28,7 +28,12 @@ public async Task<IActionResult> OnGetAsync()
 
             if (EventRole != EventRole.play && EventRole != EventRole.admin)
             {
-                return NotFound();
+                return Forbid();
+            }
+
+            if (EventRole == EventRole.play && GetTeamId().Result != -1)
+            {
+                return Forbid();
```
morganbr: Should we do anything friendlier to let people know they're already on a team?

ServerCore/Pages/Teams/Hints.cshtml.cs
```diff
@@ -10,6 +11,8 @@
 
 namespace ServerCore.Pages.Teams
 {
+    [Authorize(Policy = "PlayerCanSeePuzzle")]
```
morganbr: Were you able to verify that having these two attributes is an AND relationship? I thought it was an OR (meaning that now either being able to see the puzzle or being on the team would allow access)

ServerCore/Pages/Teams/Play.cshtml.cs
```diff
@@ -31,10 +31,15 @@ public PlayModel(PuzzleServerContext serverContext, UserManager<IdentityUser> us
...
 
-        public async Task OnGetAsync(SortOrder? sort)
+        public async Task OnGetAsync(SortOrder? sort, int teamId)
         {
+            TeamID = teamId;
             Team myTeam = await UserEventHelper.GetTeamForPlayer(_context, Event, LoggedInUser);
-            if (myTeam != null)
+            if (teamId != myTeam.ID)
+            {
+                throw new Exception("Unauthorized");
```
morganbr: Should this return Unauthorized()? (you'd have to make the method return Task<IActionResult>

ServerCore/Pages/Teams/Hints.cshtml.cs
jenetlan: Yes I was :)

ServerCore/Pages/Teams/Play.cshtml.cs
jenetlan: Removed - since the page itself now has an auth check that covers this case

ServerCore/Pages/Teams/Create.cshtml.cs
jenetlan: Changed to NotFound with a message

PR #280, opened 2019-03-07
Added special case text for free hints.

mbgoseco:
Updated EF Core Tools.


mbgoseco:
- [x] Added condition to display unique text for a hint with cost of 0
- [x] Updated Entity Framework Core Tools

Data/Migrations/20190307195457_init.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
```
morganbr: Since you didn't make any database changes, you shouldn't need this file or the Designer file

ServerCore/ServerCore.csproj
```diff
@@ -24,7 +24,12 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.All" Version="2.1.2" />
     <PackageReference Include="Microsoft.AspNetCore.Authentication.MicrosoftAccount" Version="2.1.2" />
+    <PackageReference Include="Microsoft.EntityFrameworkCore" Version="2.2.2" />
```
morganbr: I believe these packages get automatically included by other packages, so we shouldn't need them explicitly added or upgraded. Did you hit issues without them?

morganbr:
Thanks for the contribution, @mbgoseco! I've added a couple of small comments. 

Out of curiosity, how did you happen to find us?

jenetlan:
@mbgoseco are you planning on finishing this PR? If not we will either take it over or remove it next week

tabascq:
Actually, someone else just made a far more expansive change that covers this. I'm going to close this PR, feel free to submit a new one if you think there is something the other person missed.

Thanks!

PR #281, opened 2019-03-07
Require Unique Team Names 

c-costello:
- [X] Extracted Team information from Team object
- [X] Searched for a already existing Team with the same name and Event
- [X] Set conditional that if duplicate exists, return not found. Else, add the Team to the Teams table.
- [X] Commented our code

### DIFF
--- a/ServerCore/Pages/Teams/Create.cshtml.cs
+++ b/ServerCore/Pages/Teams/Create.cshtml.cs
@@ -89,6 +89,18 @@ namespace ServerCore.Pages.Teams
                     return NotFound("Registration is full. No further teams may be created at the present time.");
                 }
 
+                //Isolate team name and Event ID
+                string teamName = Team.Name;
+                Event teamEvent = Team.Event;
+                int teamEventID = Team.Event.ID;
+                //Search for a team that matches Team Name and Event ID
+                Team duplicate = await _context.Teams.FirstOrDefaultAsync(t => t.Name == Team.Name && t.Event.ID == Team.Event.ID);
+                //if duplicate exists(is not null) return not found. Otherwise, add the team to the Teams table. 
+                if (duplicate != null)
+                {
+                    return NotFound("This team name is already taken.");
+                }
+
                 _context.Teams.Add(Team);
 
                 if (EventRole == EventRole.play)


ServerCore/Pages/Teams/Create.cshtml.cs
```diff
@@ -89,6 +89,18 @@ public async Task<IActionResult> OnPostAsync()
                     return NotFound("Registration is full. No further teams may be created at the present time.");
                 }
 
+                //Isolate team name and Event ID
+                string teamName = Team.Name;
```
morganbr: It looks like these three variables aren't being used. Can you please remove them?

morganbr:
Thanks for the contribution, @c-costello! This generally looks good, but I've flagged a minor comment. 

Out of curiosity, how did you happen to find us?

ServerCore/Pages/Teams/Create.cshtml.cs
```diff
@@ -89,6 +89,18 @@ public async Task<IActionResult> OnPostAsync()
...
+                Event teamEvent = Team.Event;
+                int teamEventID = Team.Event.ID;
+                //Search for a team that matches Team Name and Event ID
+                Team duplicate = await _context.Teams.FirstOrDefaultAsync(t => t.Name == Team.Name && t.Event.ID == Team.Event.ID);
```
lambertwang: Do you know if it's possible to do a case-insensitive comparison for team names? For example, we want to consider "TeAm NaMe" and "tEaM nAmE" as duplicates.

lambertwang: Probable something like:
```
Team duplicate = await _context.Teams
    .FirstOrDefaultAsync(t => (t.Name.ToLower() == Team.Name.ToLower()) && 
                              (t.Event.ID == Team.Event.ID));
```

joncarmsft: If Teams is an actual .NET list then string.Compare is your best bet unless you have already restricted the team names to a safe set of characters. [See the docs](https://docs.microsoft.com/en-us/dotnet/api/system.string.compare?view=netframework-4.7.2#System_String_Compare_System_String_System_String_System_StringComparison_).

If that is some magic Linq-to-SQL query then I don't know.

jenetlan:
@c-costello are you planning on finishing this PR? If not we will either take it over or remove it next week

jenetlan:
Closing due to lack of activity

PR #287, opened 2019-03-12
Fixing team create experience

jenetlan:
Missed updating the redirect to the 'details' page after creating a team when I added teamId as a required parameter. Fixing that.

### DIFF
--- a/ServerCore/Pages/Teams/Create.cshtml.cs
+++ b/ServerCore/Pages/Teams/Create.cshtml.cs
@@ -122,9 +122,10 @@ namespace ServerCore.Pages.Teams
                 transaction.Commit();
             }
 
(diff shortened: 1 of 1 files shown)


ServerCore/Pages/Teams/Create.cshtml.cs
```diff
@@ -122,9 +122,10 @@ public async Task<IActionResult> OnPostAsync()
                 transaction.Commit();
             }
 
-            if(EventRole == ModelBases.EventRole.play)
+            int teamId = GetTeamId().Result;
```
morganbr: This should use ```await``` instead of ```.Result```

jenetlan: Thanks - fixed! :D

PR #290, opened 2019-03-12
Fix duplicate response logic to only check responses for this puzzle

morganbr:
Responses were incorrectly being checked against all responses in the database. The bulk uploader also needs to check for duplicates in the input set.

### DIFF
--- a/ServerCore/Pages/Responses/CreateBulk.cshtml.cs
+++ b/ServerCore/Pages/Responses/CreateBulk.cshtml.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


ServerCore/Pages/Responses/CreateBulk.cshtml.cs
```diff
@@ -53,6 +54,8 @@ public async Task<IActionResult> OnPostAsync(int puzzleId)
             StringReader responseTextReader = new StringReader(ResponseText ?? string.Empty);
             StringReader noteReader = new StringReader(Note ?? string.Empty);
 
+            HashSet<string> submissions = new HashSet<string>();
```
tabascq: unused? Do you need to build this and use it as well as checking existing responses?

morganbr: It's used on line 109 to build and check.

PR #291, opened 2019-03-12
Mark teams solved as responses are added during the event

tabascq:


ServerCore/PuzzleStateHelper.cs
```diff
@@ -514,5 +515,37 @@ private static async Task UnlockAnyPuzzlesThatThisSolveUnlockedAsync(PuzzleServe
...
+                        if (response.IsSolution && s.State.SolvedTime == null)
+                        {
+                            await SetSolveStateAsync(context, puzzle.Event, puzzle, s.State.Team, s.Submission.TimeSubmitted);
+                        }
+                    }
+
+                    await context.SaveChangesAsync();
+                }
+
+                transaction.Commit();
```
morganbr: No need to commit if you're not saving changes

PR #297, opened 2019-03-13
Better sort of responses

tabascq:


ServerCore/Pages/Responses/Index.cshtml.cs
```diff
@@ -26,7 +26,7 @@ public IndexModel(PuzzleServerContext serverContext, UserManager<IdentityUser> u
         public async Task OnGetAsync(int puzzleId)
         {
             Puzzle = await _context.Puzzles.Where(m => m.ID == puzzleId).FirstOrDefaultAsync();
-            Responses = await _context.Responses.Where((r) => r.Puzzle != null && r.Puzzle.ID == puzzleId).ToListAsync();
+            Responses = await _context.Responses.Where((r) => r.Puzzle != null && r.Puzzle.ID == puzzleId).OrderByDescending((r) => r.IsSolution).ThenBy((r) => r.ResponseText).ThenBy((r) => r.SubmittedText).ToListAsync();
```
morganbr: I might have gone by SubmittedText first, but this is an improvement either way.

tabascq: ResponseText is very deliberate, to put all DC's together in a block.

PR #312, opened 2019-03-15
Improve submission formatting of ligatures and accented characters.

nshomber:


jenetlan:
This is definitely an interesting change - may I ask for a sample of why you expect this to be necessary (not from your actual puzzle of course)? From my reading it sounds like you expect possible submissions to include accents or other special characters that you'd like to turn into un-accented letters to compare to the official submissions, but I want to make sure I'm understanding correctly. I have a few more questions I've separated out below

Data/DataModel/Response.cs
```diff
@@ -80,7 +81,15 @@ public static string FormatSubmission(string submission)
...
-            return Regex.Replace(submission, @"[^a-zA-Z\d]", string.Empty).ToUpper();
+            // Use Compatibility Decomposition (FormKD) so that
+            // 1. Compatible equivalents like exponents, ligatures, and Roman numerals are converted into preserved common characters.
+            // 2. Canonical composite characters like the accented e (\u00E9) characters in re'sume' are decomposed to preserve the starter characters.
+            if (!submission.IsNormalized(NormalizationForm.FormKD))
+            {
+                submission = submission.Normalize(NormalizationForm.FormKD);
+            }
+
+            return Regex.Replace(submission, @"[^a-zA-Z\d]", string.Empty).ToUpperInvariant();
```
jenetlan: What is the difference between the current ToUpper and ToUpperInvariant?

jenetlan:
Is this a change that should be made to the author version of this page where they enter the responses for the puzzle? I could see it not being necessary since the authors can see and edit the normalized version but it seems like a good idea to be consistent

Data/DataModel/Response.cs
nshomber: The main difference is that `ToUpper` uses the thread's _current_ locale while `ToUpperInvariant` uses the invariant locale. The most common place where they can get different results is if the `Thread.CurrentThread.CurrentCulture` is Turkish (Turkey) where i->İ, as opposed to most of the rest of the world where i->I. (note the dot over the Turkish version since dotted and dotless 'I' are distinct characters in the Turkish alphabet).

nshomber:
> Is this a change that should be made to the author version of this page where they enter the responses for the puzzle? I could see it not being necessary since the authors can see and edit the normalized version but it seems like a good idea to be consistent

Yes, I believe so. That way authors will get the same experience when setting answers as when players attempt to submit them. I believe that this single change will already update all of those paths because `Response.FormatSubmission` is called from the following places:
- `Response.Submission` setter
- `Submission.Submission` setter
- `CreateBulkModel.OnPostAsync` for each of the submissions read from `submittedTextReader`

nshomber:
Further justification for the **Compatibility** normalization form:
Suppose a puzzle involved working with Japanese text, or a player generally used a Japanese IME on their computer. It is conceivable that they may attempt to submit an answer composed of the full-width ASCII variants instead of the lower 7-bit ASCII values we are more familiar with in Latin scripts. For example: ＡＮＳＷＥＲ would be turned into the empty string by the current validation as these characters are in the range FF21-FF39 instead of the ASCII A-Z. 

By using a compatibility normalization, these would first be replaced with their ASCII equivalents as ANSWER, before being passed on to the rest of the submission system.

Further justification for the **Decomposed** normalization form:
> may I ask for a sample of why you expect this to be necessary (not from your actual puzzle of course)?

Suppose that the answer to a puzzle was CAFE. If you try typing this in Microsoft Word, you may notice that it auto-corrects to CAFÉ with an accent. If a player then copy/pastes this string as a submission (or an author tries to set a response for it), the existing code will turn this into the string "CAF", dropping the É since it does not fit the `[A-Za-z\d]` rule in the Regex.

Rather than waste a team's time and possibly increase their lockout times (due to an incorrect submission), I believe the server should gracefully remove the accent and treat it as though they had submitted CAFE. 

I achieved this behavior in my change by using the decomposed normalization form which will split the accented E character into two characters: `E` and` ́  `(combining accent) before stripping the accent in the call to `Regex.Replace`.

Data/DataModel/Response.cs
jenetlan: Awesome - thanks for the explanation :)

jenetlan:
This is great - thanks for the explanations and clarifications :)

PR #314, opened 2019-03-17
The Play That Goes Wrong (contains light spoilers)

jaylorch:
This adds the page necessary to run the PH20 metapuzzle The Play That Goes Wrong.  Be careful when reading the .cshtml file; it contains mild spoilers for how the metapuzzle will work.  It doesn't contain the key elements needed to solve it, though; those will be added by the author as "pieces".


ServerCore/Pages/EventSpecific/PH20/ThePlayThatGoesWrong.cshtml.cs
```diff
@@ -0,0 +1,55 @@
...
+            if (team == null) { return Unauthorized(); }
+
+            Puzzle thisPuzzle = await _context.Puzzles.FirstOrDefaultAsync(m => m.ID == puzzleId && m.Event.ID == EventId);
+            if (thisPuzzle == null || !thisPuzzle.Name.Equals("The Play That Goes Wrong")) {
+                return Unauthorized();
+            }
+            MetapuzzleId = puzzleId;
+
+            var helper = new SyncHelper(_context);
+            List<int> query_puzzle_ids = Enumerable.Range(MetapuzzleId - 10, 11).ToList();
```
tabascq: Not sure that you can assume the IDs are contiguous - no clue what the importer will do to this assumption. But it's certainly good enough for beta.

ServerCore/Pages/EventSpecific/PH20/ThePlayThatGoesWrong.cshtml
```diff
@@ -0,0 +1,3684 @@
+@page "/{eventId}/{puzzleId}/ThePlayThatGoesWrong.html"
+@model ServerCore.Pages.EventSpecific.PH20.ThePlayThatGoesWrongModel
+
+@{
+  Layout = null;
```
tabascq: So you intend for this page to have n header bar, etc.? Why?

ServerCore/Pages/EventSpecific/PH20/ThePlayThatGoesWrong.cshtml.cs
jaylorch: Yep.  It's just a hack for now.  I'll have to change a bunch of stuff when the puzzle IDs change anyway.

ServerCore/Pages/EventSpecific/PH20/ThePlayThatGoesWrong.cshtml
jaylorch: I don't want to use the default layout for two reasons:  (1) there's no way for me to add material to the <head> element, and (2) it confines the body to a pretty narrow window and I want the whole width of the browser.  It might be nice to have the header and footer, once they're settled on.

PR #318, opened 2019-03-21
Fix homepage

jenetlan:
Temp fixes to remove most of the links that don't work for most people

image


jenetlan:
Added an issue to make it look nice later: https://github.com/PuzzleServer/mainpuzzleserver/issues/319

PR #320, opened 2019-03-21
Eagerly create PuzzleStatePerTeam

morganbr:
Eagerly create PuzzleStatePerTeam rather than requiring it to be created lazily. Leaves in the lazy creation to handle dev instances and the sandbox

### DIFF
--- a/ServerCore/Pages/Events/CreateDemo.cshtml.cs
+++ b/ServerCore/Pages/Events/CreateDemo.cshtml.cs
@@ -389,7 +389,6 @@ namespace ServerCore.Pages.Events
                     _context.PuzzleAuthors.Add(new PuzzleAuthors() { Puzzle = meta, Author = demoCreatorUser });
                 }
 
-
                 // TODO: Files (need to know how to detect whether local blob storage is configured)
                 // Is there a point to adding Feedback or is that quick/easy enough to demo by hand?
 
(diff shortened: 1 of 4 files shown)


ServerCore/Pages/Events/CreateDemo.cshtml.cs
```diff
@@ -389,6 +389,7 @@ public async Task<IActionResult> OnPostAsync()
                     _context.PuzzleAuthors.Add(new PuzzleAuthors() { Puzzle = meta, Author = demoCreatorUser });
                 }
 
+
```
tabascq: no need for another blank line here

ServerCore/Pages/Events/Import.cshtml.cs
```diff
@@ -123,6 +123,26 @@ public async Task<IActionResult> OnPostImportAsync()
...
+                            _context.HintStatePerTeam.Add(new HintStatePerTeam() { Hint = destHint, TeamID = team.ID });
+                        }
+                    }
+
+                    // PuzzleStatePerTeam
+                    foreach (Team team in _context.Teams.Where(t => t.Event == Event))
+                    {
+                        int newPuzzleId = puzzleCloneMap[sourcePuzzle.ID].ID;
+                        bool hasPuzzleStatePerTeam = await (from pspt in _context.PuzzleStatePerTeam
+                                                            where pspt.PuzzleID == puzzleCloneMap[sourcePuzzle.ID].ID &&
```
tabascq: Do you mean newPuzzleId on this line?

PR #323, opened 2019-03-21
Team registration improvements

jenetlan:
Changes to improve team registration ahead of opening PH20 team registration.


ServerCore/Pages/Teams/Delete.cshtml
```diff
@@ -14,49 +14,89 @@
...
+                <div>(none - please enter this before the start of the event)</div>
+            }
+            else
+            {
+                @Html.DisplayFor(model => model.Team.PrimaryPhoneNumber)
+            }
         </dd>
         <dt>
-            @Html.DisplayNameFor(model => model.Team.PrimaryContactEmail)
+            Secondary phone number (optional)
```
tabascq: no need to repeat the optional-ness on the delete link :-)

tabascq: Ditto email below

ServerCore/Pages/Teams/List.cshtml
```diff
@@ -16,25 +16,25 @@
             Registration will be open soon.
         </div>
     }
-    else if (DateTime.UtcNow > Model.Event.TeamRegistrationBegin)
+    else if (DateTime.UtcNow > Model.Event.TeamRegistrationEnd)
     {
         <h2>Registration closed</h2>
         <div>
-            You are not part of a team for this event, and unfortuantely we're no longer accepting new teams into the event. :(
+            You are not part of a team for this event, and unfortuantely we're no longer accepting new teams into the event. You can still request to join an existing team.
```
tabascq: We should only show the "you can still join" conditionally, only when the event is still accepting team changes. Not going to block on it for opening registration though.

ServerCore/Pages/Teams/Index.cshtml.cs
```diff
@@ -15,7 +15,7 @@ public IndexModel(PuzzleServerContext serverContext, UserManager<IdentityUser> u
 
         public async Task<IActionResult> OnGetAsync()
         {
-            await LoadTeamDataAsync();
+            await LoadTeamDataAsync(false);
```
tabascq: can you use a named parameter here? basically, replace "false" with "onlyIncludeTeamsLookingForMembers: false". Same for the case where you pass true.

ServerCore/Pages/Teams/TeamListBase.cs
```diff
@@ -24,18 +24,32 @@ public TeamListBase(PuzzleServerContext serverContext, UserManager<IdentityUser>
...
-                                         select team).ToListAsync();
-
-            Teams = await (from team in _context.Teams
-                           where team.Event == Event
-                           join teamMember in _context.TeamMembers on team equals teamMember.Team
-                           group teamMember by teamMember.Team into teamCounts
-                           select new { Team = teamCounts.Key, Count = teamCounts.Count() }).ToDictionaryAsync(x => x.Team, x => x.Count);
-
+            List<Team> allTeams = null;
+            if (onlyIncludeTeamsLookingForMembers) {
```
tabascq: I don't like this part because it skews the registration counts for teams and players. I'd rather we just remove teams from the list at the end, after that's been properly calculated.

ServerCore/Pages/Teams/Delete.cshtml
jenetlan: updated :)

ServerCore/Pages/Teams/List.cshtml
jenetlan: Added :)

ServerCore/Pages/Teams/Index.cshtml.cs
jenetlan: Removed the parameter since the check got moved to just the list html file

ServerCore/Pages/Teams/TeamListBase.cs
jenetlan: Moved the check to the list html file - thanks for helping debug!

ServerCore/Pages/Teams/TeamListBase.cs
```diff
@@ -26,10 +26,10 @@ public TeamListBase(PuzzleServerContext serverContext, UserManager<IdentityUser>
 
         protected async Task LoadTeamDataAsync()
         {
-            List<Team> allTeams = await (from team in _context.Teams
-                                         where team.Event == Event
-                                         select team).ToListAsync();
-
+            List<Team> allTeams = null;
```
tabascq: super nit - could remove the part that sets to null.

jenetlan: fixed :)

PR #325, opened 2019-03-22
Fixes to LoginPartial formatting

jaylorch:
This fixes two issues with LoginPartial:  (1) pages that were including it were wrapping it in a ul even though it creates its own ul. (2) It was putting the class "navbar-right" on the form in addition to the ul.


jaylorch:
I've dealt with the merge conflict.

PR #342, opened 2019-03-27
Puzzle version numbers and dynamic sync clients (minor spoilers in deleted content)

jaylorch:
Each puzzle has a PuzzleVersion field that gets incremented whenever its author uploads or deletes file content.

The sync controller can now respond to Get requests at /{eventId}/{puzzleId}/api/Sync/client.  In this case, it fetches the alphabetically-latest file associated with the puzzle whose name contains "client", and serves it as HTML.  But, first, it computes a sync result and substitutes it for "@SYNC" wherever it appears in the HTML.

The upshot of all this is that now, for The Play That Goes Wrong, I can update the puzzle during the event without requiring any new deployment.  All I have to do is point the solver to /{eventId}/{puzzleId}/api/Sync/client, and by uploading a new client*.html I can have it automatically update.


ServerCore/SyncController.cs
```diff
@@ -244,11 +234,15 @@ private string GetSiteVersion(Puzzle puzzle)
             foreach (var solvedPuzzle in solves) {
                 // If the request is asking whether certain puzzles are solved, check if it's
                 // in the set and, if so, put it in the list of solved puzzles to inform the
-                // requester of.
+                // requester of.  Also, if we've been asked to query all puzzles in the excluded
+                // group and this is one of them, put it in the list of solved puzzles.
```
tabascq: Do I want to know why? Seems counterintuitive but I don't know what the excluded group is.

ServerCore/SyncController.cs
```diff
@@ -489,8 +484,74 @@ public SyncController(PuzzleServerContext context, UserManager<IdentityUser> use
...
+        
+        // GET api/Sync/client
+        [HttpGet]
+        [Route("{eventId}/{puzzleId}/api/Sync/client")]
+        public async Task<IActionResult> Index(string eventId, int puzzleId)
+        {
+            Event currentEvent = await EventHelper.GetEventFromEventId(context, eventId);
+            if (currentEvent == null)
+            {
+                return Content("ERROR:  That event doesn't exist");
```
tabascq: You can return NotFound(), Forbid(), and things like this so you get proper return codes. Those take strings as well so you can add the ones you have.

ServerCore/SyncController.cs
```diff
@@ -489,8 +484,74 @@ public SyncController(PuzzleServerContext context, UserManager<IdentityUser> use
...
+            }
+
+            // Start doing a sync asynchronously while we download the file contents.
+
+            var helper = new SyncHelper(context);
+            Task<Dictionary<string, object>> responseTask = helper.GetSyncResponse(currentEvent.ID, team.ID, puzzleId, null, 0, null, null, true);
+
+            // Find the material file with the latest-alphabetically ShortName
+
+            var materialFiles = await (from f in context.ContentFiles
```
tabascq: Why not just delete the old material file so there's always one anyway? If that's a bad idea, why not use OrderByDescending here and take the first one?

ServerCore/SyncController.cs
jaylorch: Ah, thanks for the info.  I was annoyed by the many pages on the website that just give an ACCESS DENIED or FORBIDDEN error without giving the slightest clue what the problem is, which is why I did it this way.  But it's nice to know there's a way to get the best of both worlds, a nicely-formatted error that still has some useful information.

ServerCore/SyncController.cs
jaylorch: Good point.  I'll rename that from "excluded group" to something that conveys "group this puzzle is part of".  (It was "excluded" before because it doesn't count toward solve count.)

ServerCore/SyncController.cs
jaylorch: The main reason for having code dealing with multiple concurrent versions is that it's currently not supported to overwrite an existing materials file. :-(  So if you want to make sure there isn't a point where there's no materials file, you have to first add one and then delete the old one.  So there's a point where there are more than one, and I wanted code to handle that.

And, yes, it's a good idea to switch to use OrderByDescending.  I'll do that.

ServerCore/SyncController.cs
jaylorch: Well, I tried using Forbid("..."), but it produced really ugly error messages.  An example of what the user sees is below.  So I'll stick with the informative error messages I have.

InvalidOperationException: No authentication handler is registered for the scheme 'ERROR: That event doesn't exist'. The registered schemes are: Identity.Application, Identity.External, Identity.TwoFactorRememberMe, Identity.TwoFactorUserId, Microsoft. Did you forget to call AddAuthentication().Add[SomeAuthHandler]("ERROR: That event doesn't exist",...)?

Issue #139, opened 2018-12-13
Need to import ContentFiles


tabascq:
I left this out of the importer because I don't know how blob storage works or is even configured, so I couldn't do this part.

tabascq:
Fixed by #343

PR #343, opened 2019-03-27
Import Content Files (#139)

tabascq:
In preparation for the big PH20 import...

### DIFF
--- a/Data/DataModel/ContentFile.cs
+++ b/Data/DataModel/ContentFile.cs
@@ -11,6 +11,19 @@ namespace ServerCore.DataModel
     /// </summary>
     public class ContentFile
     {
+        public ContentFile()
+        {
+        }
+
+        public ContentFile(ContentFile source)
+        {
+            FileType = source.FileType;
+            Event = source.Event;
+            ShortName = source.ShortName;
+            UrlString = source.UrlString;
+            Puzzle = source.Puzzle;
+        }
+
         [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
         public int ID { get; set; }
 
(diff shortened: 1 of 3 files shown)


ServerCore/FileManager.cs
```diff
@@ -22,6 +23,37 @@ public class FileManager
...
+        public static async Task<Uri> CloneBlobAsync(string fileName, int eventId, Uri sourceUri)
+        {
+            CloudBlockBlob blobSource = new CloudBlockBlob(sourceUri, StorageAccount.Credentials);
+
+            CloudBlockBlob blob = await CreateNewBlob(fileName, eventId);
+            await blob.StartCopyAsync(blobSource);
+
+            while (blob.CopyState.Status == CopyStatus.Pending)
+            {
+                await Task.Delay(1000);
```
morganbr: I'm worried the sleeping could cause page timeouts for the import page, which could leave us in a torn state. Do we actually need to wait for it to finish copying?

Data/DataModel/ContentFile.cs
```diff
@@ -11,6 +11,19 @@ namespace ServerCore.DataModel
...
+        public ContentFile()
+        {
+        }
+
+        public ContentFile(ContentFile source)
+        {
+            FileType = source.FileType;
+            Event = source.Event;
+            ShortName = source.ShortName;
+            UrlString = source.UrlString;
```
morganbr: Would it make sense to exclude the properties that should be rewritten? The only persistent properties are FileType and ShortName

ServerCore/FileManager.cs
tabascq: Removed, good justification thanks

Data/DataModel/ContentFile.cs
tabascq: I'm trying to have every class copy everything. I agree it's overkill in this case but it will make it easier to find missing properties overall.

PR #350, opened 2019-03-28
Changed lockout algorithm to be simpler

### LINKED ISSUE #344: Submission lockout too permissive
In the sandbox, I was able to spam 11 submissions and only got delayed once (by two minutes after the sixth incorrect answer).  This is far too permissive.  I suggest the following algorithm:

If the number of incorrect submissions so far is N, and N>=5, then compute MinimumAllowableSubmitTime = TimeOfLastIncorrectSubmission + Max(5, N-4) minutes.  Don't allow submissions until the current time is MinimumAllowableSubmitTime.

lambertwang:
Fixes #344


Data/Migrations/PuzzleServerContextModelSnapshot.cs
```diff
@@ -15,7 +15,7 @@ protected override void BuildModel(ModelBuilder modelBuilder)
         {
 #pragma warning disable 612, 618
             modelBuilder
-                .HasAnnotation("ProductVersion", "2.1.8-servicing-32085")
+                .HasAnnotation("ProductVersion", "2.1.1-rtm-30846")
```
jaylorch: It seems like we should leave the ProductVersion unchanged.  But maybe this doesn't matter?

ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -174,88 +177,47 @@ private async Task SetupContext(int puzzleId)
...
-             * minutes.
-             */
-
-            return DateTime.UtcNow.AddMinutes(
-                (consecutiveWrongSubmissions -
-                 ev.LockoutIncorrectGuessLimit) *
-                ev.LockoutDurationMultiplier);
-
+            return DateTime.UtcNow.AddMinutes(Math.Max(
+                ev.LockoutBaseDuration,
```
jaylorch: The identifier name LockoutBaseDuration seems misleading, suggesting a base that's added to rather than a maximum.  So I'd suggest renaming this to something like LockoutMaxMinutes.

jaylorch:
I suggest making the changes I suggested in my comments before merging, but this looks great.  Thanks!

tabascq:
Hey @jaylorch , Lambert and I had a discussion about this yesterday after you left. I do not like that this has no cooldown period at all. If someone gets frustrated with a puzzle and sends in some spam answers, and then someone else picks it up the next day, they should not be carrying the penalty from the day before, at least not for all events the system could host.

jaylorch:
So you want an event-specific cooldown period, and for PH20 we could set it to some large number of hours?  By "cooldown period", I assume you mean how far back the lockout check looks when computing the wrong-submission count.

I learned just now (when I was reviewing this pull request) that there's an absolute limit of 50 incorrect submissions per puzzle per team, so I'm less worried about a cooldown period than I was before learning about that.

tabascq:
Yes - that's what the original was designed to do, and one of the things we wondered is whether the default event properties just need tweaking. That's where the 50 comes from, for instance.

tabascq:
(and the PH20 event properties of course)

tabascq:
Existing properties:

        /// <summary>
        /// The window of time where if a team enters a certain number of
        /// incorrect answers would cause the team to be locked out of
        /// submitting additional answers for a brief amount of time.
        /// </summary>
        public double LockoutIncorrectGuessPeriod { get; set; } // current default is 1 minute

        /// <summary>
        /// The amount of incorrect submissions required to initiate a lockout.
        /// </summary>
        public int LockoutIncorrectGuessLimit { get; set; } // current default is 5

        /// <summary>
        /// The multiplier for the lockout duration for consecutive lockouts.
        /// </summary>
        public double LockoutDurationMultiplier { get; set; } // current default is 2

        /// <summary>
        /// The maximum number of incorrect submissions a team can have for
        /// a single puzzle before being locked out entirely and being placed
        /// in email-only mode.
        /// </summary>
        public uint MaxSubmissionCount { get; set; } // current default is 50

tabascq:
As discussed, the original algorithm works fine, though its defaults need adjusting.

PR #354, opened 2019-03-29
Zip file upload

### LINKED ISSUE #65: Unpack zipped ContentFiles
Some puzzle contents could be small static websites (e.g. an html page with some images). They need to end up in the same blob directory for relative links to work. A way to do that could be to accept a zip file and extract it on the server before uploading all of the files to a blob.

We would need a way to differentiate zip files that aren't intended to be extracted from those that are.

### LINKED ISSUE #490: Alphabetize file list
Files are listed in random order in the file upload view.

morganbr:
Allow zip files to be uploaded and expanded into a single directory. This is mostly useful for static pages that want to use relative links (e.g. to embed images).

Fixes #65. Fixes #490

### DIFF
--- a/ServerCore/FileManager.cs
+++ b/ServerCore/FileManager.cs
@@ -50,7 +50,7 @@ namespace ServerCore
 
         private static async Task<CloudBlockBlob> CreateNewBlob(string fileName, int eventId)
         {
-            CloudBlobContainer eventContainer = await GetOrCreateEventContainerAsync(eventId);
+            CloudBlobDirectory puzzleDirectory = await GetRandomDirectoryAsync(eventId);
 
             CloudBlockBlob blob = puzzleDirectory.GetBlockBlobReference(fileName);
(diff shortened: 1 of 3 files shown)


tabascq:
I'm not sure how to import this easily and am anxious about people using it before that's ready. Is there anything that currently really needs this, or have we all worked around it already?

morganbr:
@tabascq, I might still want it for something, though I'm not certain. For import, maybe we could carry the randomized directory names between events. If we didn't want to do that, we could add a column to the table tracking an origin, but that seems more complicated.

tabascq:
I think the import would need to:

- enumerate all directories pointed to by the original event
- duplicate all those directories in full, maintaining a map from original directory to the duplicate
- Clone all ContentFiles records, and update the URIs to reflect the map contents

It's doable, just takes time to get right and I only want the hassle if it's necessary.

PR #358, opened 2019-03-31
Fix #356 by moving a where clause

### LINKED ISSUE #356: Final metapuzzle doesn't appear in puzzle list
When I look at the list of puzzles I've authored (https://puzzlehunt.azurewebsites.net/1/author/Puzzles), my metameta doesn't appear.  It used to appear on this page, so this is a regression.

tabascq:
By moving the filetype where-clause before the join instead of after,  we get the right results and what I hope is equivalent-or-better perf.

### DIFF
--- a/ServerCore/Pages/Puzzles/Index.cshtml.cs
+++ b/ServerCore/Pages/Puzzles/Index.cshtml.cs
@@ -40,9 +40,8 @@ namespace ServerCore.Pages.Puzzles
             }
 
             Puzzles = await (from Puzzle p in query
-                             join ContentFile joinFile in _context.ContentFiles on p equals joinFile.Puzzle into fileJoin
+                             join ContentFile joinFile in _context.ContentFiles.Where((j) => j.FileType == ContentFileType.Puzzle) on p equals joinFile.Puzzle into fileJoin
                              from ContentFile file in fileJoin.DefaultIfEmpty()
-                             where file == null || file.FileType  == ContentFileType.Puzzle
                              orderby p.Group, p.OrderInGroup, p.Name
                              select new PuzzleView { Puzzle = p, Content = file }).ToListAsync();
 
--- a/ServerCore/Pages/Puzzles/Index.cshtml.cs
+++ b/ServerCore/Pages/Puzzles/Index.cshtml.cs
@@ -40,7 +40,7 @@ namespace ServerCore.Pages.Puzzles
             }
 
             Puzzles = await (from Puzzle p in query
-                             join ContentFile joinFile in _context.ContentFiles.Where((j) => j.FileType == ContentFileType.Puzzle) on p equals joinFile.Puzzle into fileJoin
+                             join ContentFile joinFile in _context.ContentFiles.Where((j) => j.Event == Event && j.FileType == ContentFileType.Puzzle) on p equals joinFile.Puzzle into fileJoin
                              from ContentFile file in fileJoin.DefaultIfEmpty()
                              orderby p.Group, p.OrderInGroup, p.Name
                              select new PuzzleView { Puzzle = p, Content = file }).ToListAsync();
--- a/ServerCore/Pages/Teams/Play.cshtml.cs
+++ b/ServerCore/Pages/Teams/Play.cshtml.cs
@@ -64,9 +64,8 @@ namespace ServerCore.Pages.Teams
             // Note: EF gotcha is that you have to join into anonymous types in order to not lose valuable stuff
             var visiblePuzzlesQ = from Puzzle puzzle in puzzlesInEventQ
                                   join PuzzleStatePerTeam pspt in stateForTeamQ on puzzle.ID equals pspt.PuzzleID
-                                  join ContentFile joinContent in _context.ContentFiles on puzzle equals joinContent.Puzzle into fileJoin
+                                  join ContentFile joinContent in _context.ContentFiles.Where((j) => j.FileType == ContentFileType.Puzzle) on puzzle equals joinContent.Puzzle into fileJoin
                                   from ContentFile content in fileJoin.DefaultIfEmpty()
-                                  where content == null || content.FileType == ContentFileType.Puzzle
                                   select new { Puzzle = puzzle, State = pspt, Content = content };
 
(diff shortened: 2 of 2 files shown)


morganbr:
@tabascq , can you please post the queries from your debugger output that this produced?

tabascq:
I have absolutely no idea how to do that. Give me detailed instructions and I will do my best to follow them.

tabascq:
Oh wait, you mean the queries that are *already there* in my output. Sure, one sec :-)

morganbr:
Debug one of the pages and paste your debugger's output pane here

tabascq:
I think it's this one:

SELECT [p].[ID], [p].[CustomURL], [p].[Description], [p].[EventID], [p].[Group], [p].[HintCoinsForSolve], [p].[HintsAreCumulative], [p].[IsCheatCode], [p].[IsFinalPuzzle], [p].[IsGloballyVisiblePrerequisite], [p].[IsMetaPuzzle], [p].[IsPuzzle], [p].[MaxAnnotationKey], [p].[MinPrerequisiteCount], [p].[MinutesOfEventLockout], [p].[MinutesToAutomaticallySolve], [p].[Name], [p].[OrderInGroup], [p].[PuzzleVersion], [p].[SolveValue], [p].[SupportEmailAlias], [p].[Token], [t].[ID], [t].[EventID], [t].[FileType], [t].[PuzzleID], [t].[ShortName], [t].[UrlString]
FROM [Puzzles] AS [p]
LEFT JOIN (
    SELECT [j].[ID], [j].[EventID], [j].[FileType], [j].[PuzzleID], [j].[ShortName], [j].[UrlString]
    FROM [ContentFiles] AS [j]
    WHERE [j].[FileType] = 0
) AS [t] ON [p].[ID] = (
    SELECT TOP(1) [subQuery0].[ID]
    FROM [Puzzles] AS [subQuery0]
    WHERE [subQuery0].[ID] = [t].[PuzzleID]
)

ServerCore/Pages/Puzzles/Index.cshtml.cs
```diff
@@ -40,9 +40,8 @@ public async Task<IActionResult> OnGetAsync()
             }
 
             Puzzles = await (from Puzzle p in query
-                             join ContentFile joinFile in _context.ContentFiles on p equals joinFile.Puzzle into fileJoin
+                             join ContentFile joinFile in _context.ContentFiles.Where((j) => j.FileType == ContentFileType.Puzzle) on p equals joinFile.Puzzle into fileJoin
```
morganbr: Mixing the two kinds of LINQ syntax is a bit weird. I'm not certain of the right syntax, so you can leave this as-is if you need to. My guess on the right syntax would be:

```
from Puzzle p in query
let puzzleFiles = from content in _context.ContentFiles
where content.FileType == ContentFileType.Puzzle
join ContentFile joinFile in puzzleFiles on p equals joinFile.Puzzle into fileJoin
from ContentFile file in fileJoin.DefaultIfEmpty()
orderby p.Group, p.OrderInGroup, p.Name
select new PuzzleView { Puzzle = p, Content = file }).ToListAsync();
```

tabascq:
Jay, I also find joins to be completely inscrutable as well and cannot write one to save my life. But I think I know just enough to make this work. A join takes two inputs, and I prefiltered one of those inputs (the one from ContentFiles) so it doesn't ever see materials, etc.

morganbr:
Just realized the new query is probably not filtering the content files by event. That produces the right output, but may be slow when we have multiple events. You might need to filter the ContentFile's event Id

tabascq:
Morgan, I can barely read the LINQ syntax you use and do not feel remotely comfortable writing it. That may well be it but I'm afraid to try it because I certainly can't apply it to the other one.

jaylorch:
One thing that puzzles me about the join is why DefaultIfEmpty doesn't need a parameter saying what the default *is* if empty.  It seems like it magically knows that the default if empty should be an element with all the puzzle fields as you want them, instead of (as I would think) a default join element that has all the puzzle fields null.

morganbr:
@jaylorch it's using the default .NET value for a class, which is null.

tabascq:
I added the event filter (using the only syntax I can write)

morganbr:
Can we see the new debug output?

tabascq:
SELECT [p].[ID], [p].[CustomURL], [p].[Description], [p].[EventID], [p].[Group], [p].[HintCoinsForSolve], [p].[HintsAreCumulative], [p].[IsCheatCode], [p].[IsFinalPuzzle], [p].[IsGloballyVisiblePrerequisite], [p].[IsMetaPuzzle], [p].[IsPuzzle], [p].[MaxAnnotationKey], [p].[MinPrerequisiteCount], [p].[MinutesOfEventLockout], [p].[MinutesToAutomaticallySolve], [p].[Name], [p].[OrderInGroup], [p].[PuzzleVersion], [p].[SolveValue], [p].[SupportEmailAlias], [p].[Token], [t].[ID], [t].[EventID], [t].[FileType], [t].[PuzzleID], [t].[ShortName], [t].[UrlString]
FROM [Puzzles] AS [p]
LEFT JOIN (
    SELECT [j].[ID], [j].[EventID], [j].[FileType], [j].[PuzzleID], [j].[ShortName], [j].[UrlString]
    FROM [ContentFiles] AS [j]
    WHERE ([j].[EventID] = @__Event_1_ID) AND ([j].[FileType] = 0)
) AS [t] ON [p].[ID] = (
    SELECT TOP(1) [subQuery0].[ID]
    FROM [Puzzles] AS [subQuery0]
    WHERE [subQuery0].[ID] = [t].[PuzzleID]
)
WHERE [p].[EventID] = @__Event_0_ID
ORDER BY [p].[Group], [p].[OrderInGroup], [p].[Name], [p].[ID]

morganbr:
K. It looks like it might be doing some extra subqueries, but it's probably the best we can do. Ship it.

jaylorch:
For anyone else who was curious, I found the pattern of using DefaultIfEmpty for a left outer join documented here:  https://docs.microsoft.com/en-us/dotnet/csharp/linq/perform-left-outer-joins

PR #364, opened 2019-04-06
Display team lists in alphabetical order

jaylorch:
Without this change, team lists get displayed in random order on various pages that inherit from TeamListBase.  The reason for this is that the page display iterates through a Dictionary<Team, int> but a dictionary doesn't order its keys.

So, this change replaces the model's Dictionary<Team, int> that maps teams to player count with two data structures.  The first is a List<Team> that has the teams in alphabetical order by name.  The second is a Dictionary<int, int> that maps team IDs to player count.


Vroo:
Thanks. FYI, you can pass for a developer in my book. :-)

Yeah, I can change that later. As Kenny mentioned to me the priority is
features over perfect code.

--- Bruce

Issue #362, opened 2019-04-04
Email infrastructure


Vroo:
I'm going to add code to send mail. This is just the infrastructure work, not wiring it up in specific cases where we want to send mail. There are other issues that need this work, like #202, #22, #166.

PR #366, opened 2019-04-06
F

Vroo:
Implement #362.
I don't know what the pattern is for singletons in this project. I'm using `MailHelper.initialize` and `MailHelper.Singleton`. Happy to change that to whatever is the standard. I couldn't find an existing use case to copy.

### DIFF
--- a/ServerCore/Helpers/MailHelper.cs
+++ b/ServerCore/Helpers/MailHelper.cs
@@ -1,0 +1,100 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Net.Mail;
+using Microsoft.Extensions.Configuration;
+
+/// <summary>
+/// Helper class for sending mail with Mailjet.
+/// </summary>
+public class MailHelper
+{
+    private string PublicSecret;
+    private string PrivateSecret;
+
+    public bool Enabled { get; private set; }
+    private bool IsDev;
+
+    private const string FromAddress = "msph@puzzlehunt.org";
+    private const string ToAddressForBCC = "msph-bcc@puzzlehunt.org";
+    
+    public static MailHelper Singleton { get; private set; }
+
+    public static void initialize(IConfiguration configuration, bool isDev)
+    {
+        Singleton = new MailHelper(configuration, isDev);
+    }
+
+    private MailHelper(IConfiguration configuration, bool isDev)
+	{
+        try
+        {
+            PublicSecret = configuration["Mailjet-Public"];
+            PrivateSecret = configuration["Mailjet-Private"];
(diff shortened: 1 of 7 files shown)


ServerCore/Helpers/MailHelper.cs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            foreach (string recipient in recipients) {
+                msg.To.Add(new MailAddress(recipient));
+            }
+        }
+
+        msg.Subject = subject;
+        msg.Body = body;
+
+        SmtpClient client = new SmtpClient("in.mailjet.com", 587);
```
tabascq: Is this specific choice of port something we need to document? No worries if not. I know there are multiple SMTP ports but I do not know enough to know the differences between them.

ServerCore/Pages/Hints/Edit.cshtml
```diff
@@ -13,6 +13,10 @@
     <a asp-page="./Index">Cancel</a>
 </div>
 
+<div>
```
tabascq: Not sure why these hint things are part of this PR, but if you feel they're truly ready to go I won't quibble, in the name of throughput.

ServerCore/Pages/Teams/Hints.cshtml
```diff
@@ -1,6 +1,48 @@
...
+        color: #aaaaaa;
+    }
+
+    /*
+     * Strikeout style for hints
+     */
+    .strike1 {
+        position: relative;
+    }
+    .strike1:before {
```
tabascq: I could have sworn I saw and reviewed these before...

tabascq: Indeed, I have them in my copy of the file. What happened? Did this get reverted somehow and you are replacing it?

tabascq:
FWIW, I think I see many of the hint changes in the master branch when viewed from the web site, which makes me extra baffled why they show up as diffs in the PR.

ServerCore/Pages/Hints/Edit.cshtml
Vroo: There are no changes related to hints. Github confusion. I will fix.

ServerCore/Helpers/MailHelper.cs
Vroo: That's the port that Mailjet tells us to use. I will add a link to their documentation.

Vroo: Actually, I have no idea how to fix. I will cherry-pick the commit over to a new branch and send out a new review after addressing the issues.

PR #368, opened 2019-04-06
MailHelper class and /testconfig page

Vroo:
Take 2 on mail infra work, without random repeats of hint code.

Assigning to Kenny since he already started looking at this.

### DIFF
--- a/ServerCore/Helpers/MailHelper.cs
+++ b/ServerCore/Helpers/MailHelper.cs
@@ -18,6 +18,26 @@ public class MailHelper
 
     private const string FromAddress = "msph@puzzlehunt.org";
     private const string ToAddressForBCC = "msph-bcc@puzzlehunt.org";
+    private static readonly char[] EmailSeparators = { ',', ';' };
+
+    /// <summary>
+    /// This unsubscribe footer is needed because the lack of one will give us a higher
+    /// spamminess score with some email services and might damage deliverability for
+    /// everyone. Our mail is transactional, not commercial, so we do not need to comply
+    /// with CAN-SPAM but email services won't know that. Please do not remove or change.
+    /// (Note that this doesn't actually comply with CAN-SPAM since it doesn't auto-remove
+    /// them. We can't do that since we sent most mail bcc and don't know who they are.)
+    /// </summary>
+    private const string MailUnsubscribeFooter = @"
+
+--
+This mail was sent by msph@puzzlehunt.org, Redmond WA 98005 because you registered
+for a hunt and entered this email address on your profile or team page.
+To unsubscribe: go to https://puzzlehunt.azurewebsites.net/Identity/Account/Manage and login.
+Then remove your email address from your profile or delete your account. If you have entered
+your email as the contact address for a team, then you also need to remove it on the team page.
+";
+
 
     public static MailHelper Singleton { get; private set; }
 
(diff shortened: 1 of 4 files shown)


ServerCore/Helpers/MailHelper.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            msg.To.Add(new MailAddress(ToAddressForBCC));
+            foreach (string recipient in recipients) {
+                try
+                {
+                    msg.Bcc.Add(new MailAddress(recipient));
+                }
+                catch (FormatException)
+                {
+                    ;  // ignore
```
morganbr: No semicolon required for this.

ServerCore/Helpers/MailHelper.cs
```diff
@@ -0,0 +1,117 @@
...
+        client.EnableSsl = true;
+        client.UseDefaultCredentials = false;
+        client.Credentials = new NetworkCredential(PublicSecret, PrivateSecret);
+        client.Send(msg);
+    }
+
+    private void WriteMailToConsole(IEnumerable<string> recipients, string subject, string body)
+    {
+        // In dev environment, log the email instead of sending it if not configured.
+        Console.WriteLine("----------------------------------------");
```
morganbr: I'm not sure if Console.WriteLine shows up anywhere. You might try Debug.WriteLine to have it show up in the debugger if it's not.

ServerCore/Helpers/MailHelper.cs
```diff
@@ -0,0 +1,117 @@
...
+        msg.Body = body;
+
+        SmtpClient client = new SmtpClient("in.mailjet.com", 587);
+        // Unlike port 25 which is for general SMTP, port 587 is for
+        // MSA (Message Submission Agent) use and is authenticated.
+        client.DeliveryMethod = SmtpDeliveryMethod.Network;
+        client.EnableSsl = true;
+        client.UseDefaultCredentials = false;
+        client.Credentials = new NetworkCredential(PublicSecret, PrivateSecret);
+        client.Send(msg);
```
morganbr: It might be good to use SendMailAsync here and make MailHelper.Send an async method by changing the signature to ```async Task SendAsync(IEnumerable<string> recipients, bool bcc, string subject, string body)```

ServerCore/Helpers/MailHelper.cs
```diff
@@ -0,0 +1,117 @@
...
+                    ;  // ignore
+                }
+            }
+        }
+        else
+        {
+            foreach (string recipient in recipients) {
+                try
+                {
+                    msg.To.Add(new MailAddress(recipient));
```
morganbr: Should this either bcc players or send them separate messages in order to avoid reply-all problems?

ServerCore/Helpers/MailHelper.cs
```diff
@@ -0,0 +1,117 @@
...
+
+    public bool Enabled { get; private set; }
+    private bool IsDev;
+
+    private const string FromAddress = "msph@puzzlehunt.org";
+    private const string ToAddressForBCC = "msph-bcc@puzzlehunt.org";
+    
+    public static MailHelper Singleton { get; private set; }
+
+    public static void initialize(IConfiguration configuration, bool isDev)
```
morganbr: Nit -- Capital first letter for methods in C# please.

ServerCore/Helpers/MailHelper.cs
```diff
@@ -0,0 +1,117 @@
...
+    {
+        Singleton = new MailHelper(configuration, isDev);
+    }
+
+    private MailHelper(IConfiguration configuration, bool isDev)
```
morganbr: Looks like indentation might be off here or on the line below

morganbr:
Thanks, @Vroo ! I've put in a few minor comments and we'll be happy to merge after they're addressed.

ServerCore/Helpers/MailHelper.cs
Vroo: The `;` allows me to put a breakpoint there.

ServerCore/Helpers/MailHelper.cs
```diff
@@ -0,0 +1,117 @@
...
+            return;
+        }
+        MailMessage msg = new MailMessage();
+
+        msg.From = new MailAddress(FromAddress);
+
+        if (bcc)
+        {
+            msg.To.Add(new MailAddress(ToAddressForBCC));
+            foreach (string recipient in recipients) {
```
morganbr: Please put braces on the following line. If you're using visual studio, try running ctrl+k ctrl+f on this file, which will automatically format it to match our standards.

ServerCore/Helpers/MailHelper.cs
Vroo: Sending many emails will make us hit our email quota that much faster. Bcc is the way to avoid reply-all. Since the only case where we'd want them to see each other's emails is if one team member was sending mail to the rest of the team (which isn't contemplated). So I'll make bcc the only choice unless there's a single email.

ServerCore/Helpers/MailHelper.cs
Vroo: oops.

ServerCore/Helpers/MailHelper.cs
Vroo: done

ServerCore/Helpers/MailHelper.cs
Vroo: I haven't used the new async stuff in C# before. I understand the benefit of calling `SendMailAsync` instead. Since this function returns void, why would we need to change it to return a Task instead? Wouldn't the async call just work? What would we do with the Task?

ServerCore/Helpers/MailHelper.cs
Vroo: done

ServerCore/Helpers/MailHelper.cs
Vroo: Oh. I get it. If I do that then I don't need to wrap the call in `await (Task.Run())`?

Vroo: It works fine if I switch to `SendMailAsync` and remove the `Task.Run`. I'll push it with that change. Let me know if it's wrong.

Vroo:
I just pushed an update. Please let me know if you have further comments. Thanks. @morganbr

ServerCore/Helpers/MailHelper.cs
```diff
@@ -0,0 +1,128 @@
...
+        Singleton = new MailHelper(configuration, isDev);
+    }
+
+    private MailHelper(IConfiguration configuration, bool isDev)
+    {
+        try
+        {
+            PublicSecret = configuration["Mailjet-Public"];
+            PrivateSecret = configuration["Mailjet-Private"];
+            Enabled = true;
```
tabascq: I'm playing with adding mails for some scenarios and I found this line to be helpful:

            Enabled = PublicSecret != null && PrivateSecret != null;

ServerCore/Helpers/MailHelper.cs
```diff
@@ -0,0 +1,128 @@
...
+            IsDev = isDev;
+        }
+        catch
+        {
+            Enabled = false;
+        }
+        Enabled = false;
```
tabascq: That can't be right.

ServerCore/Helpers/MailHelper.cs
tabascq: I want a team application to *not* Bcc the team and the applicant. Can I do that?

ServerCore/Helpers/MailHelper.cs
```diff
@@ -0,0 +1,128 @@
...
+        client.SendMailAsync(msg);
+    }
+
+    private void WriteMailToConsole(IEnumerable<string> recipients, string subject, string body)
+    {
+        // In dev environment, log the email instead of sending it if not configured.
+        Debug.WriteLine("----------------------------------------");
+        foreach (string recipient in recipients)
+        {
+            Console.WriteLine("To: " + recipient);
```
tabascq: Debug.WriteLine

ServerCore/Helpers/MailHelper.cs
Vroo: Yeah, I could do it that way. I need the try/catch anyway, since `configuration[]` throws an exception if you haven't set the secrets and I don't know a way to say "get or default".

ServerCore/Helpers/MailHelper.cs
Vroo: done

ServerCore/Helpers/MailHelper.cs
Vroo: Fixed. I stuck that in to test something and forgot to take it out.

Vroo:
For some reason, github won't let me reply to "I want a team application to not Bcc the team and the applicant. Can I do that?" so here it is:

A team application is when an unsolicited person requests to join a team? Since we don't have captains, you do want everyone to se it and for them all to see everyone else that got it. OK, that's a legit reason for non-bcc. I will add that capability back in with a new function.

tabascq:
I want to be able to do this:

            MailHelper.Singleton.Send(new string[] { Team.PrimaryContactEmail, LoggedInUser.Email },
                false,
                $"{LoggedInUser.Name} is applying to join {Team.Name}",
                $"To accept or reject this request, visit your team page.");

Vroo:
Again, can't reply directly. That's fine but it won't work the way the code is written now if they put multiple emails in the primary contact field. I'll go ahead and add that in here.

tabascq:
I have checked the entire PH20 registration list and there is never more than a single address. [Many teams have no primary address at all, but we should make this illegal.]

Vroo:
FYI, I had already implemented the comma/semicolon splitting before Kenny's comment and it was supported by the old server. I also had both html support and the unsubscribe footer in another branch and added them here, the latter since it's possible Kenny will get his code sending mail out deployed before mine.

PR #370, opened 2019-04-06
Fixing crashes (empty hints and submissions)

jenetlan:
Fixed a bug where if there are no hints for a puzzle the player view of that puzzle's hints crashes

Fixed a bug where if there are no submissions for a puzzle the player view of that puzzle's submissions crashes


Vroo:
Thanks! I was panicked when I found this tonight.

PR #371, opened 2019-04-06
Fix exception when there are no hints for a puzzle.

Vroo:
Fix an exception when trying to show the hints for a puzzle and there are no hints. Sorry about this.


tabascq:
@jenetlan already got to this - see #370 .

tabascq:
Closing since I took #370.

PR #386, opened 2019-04-09
During the beta, once a puzzle is solved, all other hints become free.

Vroo:
I don't have time to finish testing this now but I have a question so I figure I'd go ahead and upload it and ask. It looks like there is no concept of beta in the app right now. I thought I could check the event name and if it has 'beta' in it, it's a beta. But Hints has no access to the event object. (I think it doesn't actually need it because all the puzzle ids and team ids are unique across all events. Is that right?)

So what I did instead was hardcode `IsBeta = true` and then we would need to redeploy later with that flag set false. Any better ideas?

### DIFF
--- a/ServerCore/Pages/Teams/Hints.cshtml.cs
+++ b/ServerCore/Pages/Teams/Hints.cshtml.cs
@@ -15,6 +15,8 @@ namespace ServerCore.Pages.Teams
     [Authorize(Policy = "PlayerIsOnTeam")]
     public class HintsModel : EventSpecificPageModel
     {
+        private static bool IsBeta = true;  // TODO: I don't know how to tell if this is a beta; this code seems not to have the event object available
+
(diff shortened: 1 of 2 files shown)


ServerCore/Pages/Teams/Hints.cshtml.cs
```diff
@@ -15,6 +15,8 @@ namespace ServerCore.Pages.Teams
     [Authorize(Policy = "PlayerIsOnTeam")]
     public class HintsModel : EventSpecificPageModel
     {
+        private static bool IsBeta = true;  // TODO: I don't know how to tell if this is a beta; this code seems not to have the event object available
```
tabascq: EventSpecificPageModel has an Event property on it. I think you can use Event.Name.Contains("Beta") like you wanted.

tabascq:
Thanks! Merging now.

PR #387, opened 2019-04-11
Email Scenarios

tabascq:
Using the new emailer class (Thanks, Bruce! Thanks, Roy!), I've implemented mailer support for a few scenarios:

- Applying to join a team (mails the primary contact and the applicant on To, so they can have a discussion)
- Accepting an application (mails the applicant on To)
- Revising/creating a response (mails all team members of teams who submitted that response, on Bcc)
- Revising a hint (mails all team members of teams who asked for that hint, on Bcc)
- Entering email mode (mails all authors of that puzzle, on Bcc)

### DIFF
--- a/ServerCore/Areas/Identity/Pages/Account/Edit.cshtml.cs
+++ b/ServerCore/Areas/Identity/Pages/Account/Edit.cshtml.cs
@@ -38,6 +38,15 @@ namespace ServerCore.Areas.Identity.Pages.Account
 
         public async Task<IActionResult> OnPostAsync(string returnUrl = null)
         {
+            if (string.IsNullOrEmpty(PuzzleUser.Email))
+            {
+                ModelState.AddModelError("PuzzleUser.Email", "An email is required.");
+            }
+            else if (!MailHelper.IsValidEmail(PuzzleUser.Email))
+            {
+                ModelState.AddModelError("PuzzleUser.Email", "This email address is not valid.");
+            }
+
             if (!ModelState.IsValid)
             {
                 return Page();
--- a/ServerCore/Helpers/TeamHelper.cs
+++ b/ServerCore/Helpers/TeamHelper.cs
@@ -117,7 +117,7 @@ namespace ServerCore.Helpers
             await context.SaveChangesAsync();
 
             MailHelper.Singleton.SendPlaintextWithoutBcc(new string[] { team.PrimaryContactEmail, user.Email },
-                $"{user.Name} has now joined {team.Name}!",
+                $"{Event.Name}: {user.Name} has now joined {team.Name}!",
                 $"Have a great time!");
 
             return new Tuple<bool, string>(true, "");
--- a/ServerCore/Helpers/TeamHelper.cs
+++ b/ServerCore/Helpers/TeamHelper.cs
@@ -115,6 +115,11 @@ namespace ServerCore.Helpers
 
             context.TeamMembers.Add(Member);
             await context.SaveChangesAsync();
+
+            MailHelper.Singleton.SendPlaintextWithoutBcc(new string[] { team.PrimaryContactEmail, user.Email },
+                $"{user.Name} has now joined {team.Name}!",
+                $"Have a great time!");
+
             return new Tuple<bool, string>(true, "");
         }
     }
--- a/ServerCore/Pages/Events/CreateDemo.cshtml.cs
+++ b/ServerCore/Pages/Events/CreateDemo.cshtml.cs
@@ -355,19 +355,19 @@ namespace ServerCore.Pages.Events
                 //
                 // Create teams. Can we add players to these?
                 //
-                Team team1 = new Team { Name = "Team Bugs", Event = Event };
+                Team team1 = new Team { Name = "Team Bugs", Event = Event, IsLookingForTeammates = true };
                 _context.Teams.Add(team1);
 
-                Team team2 = new Team { Name = "Team Babs", Event = Event };
+                Team team2 = new Team { Name = "Team Babs", Event = Event, IsLookingForTeammates = true };
                 _context.Teams.Add(team2);
 
-                Team team3 = new Team { Name = "Team Buster", Event = Event };
+                Team team3 = new Team { Name = "Team Buster", Event = Event, IsLookingForTeammates = true };
                 _context.Teams.Add(team3);
 
                 Team teamLoneWolf = null;
                 if (AddCreatorToLoneWolfTeam)
                 {
-                    teamLoneWolf = new Team { Name = "Lone Wolf", Event = Event };
+                    teamLoneWolf = new Team { Name = "Lone Wolf", Event = Event, IsLookingForTeammates = true };
                     _context.Teams.Add(teamLoneWolf);
                 }
 
--- a/ServerCore/Pages/Events/CreateDemo.cshtml.cs
+++ b/ServerCore/Pages/Events/CreateDemo.cshtml.cs
@@ -361,7 +361,7 @@ namespace ServerCore.Pages.Events
                 Team team2 = new Team { Name = "Team Babs", Event = Event, IsLookingForTeammates = true };
                 _context.Teams.Add(team2);
 
-                Team team3 = new Team { Name = "Team Buster", Event = Event, IsLookingForTeammates = true };
+                Team team3 = new Team { Name = "Team Buster", Event = Event, IsLookingForTeammates = false };
                 _context.Teams.Add(team3);
 
                 Team teamLoneWolf = null;
--- a/ServerCore/Pages/Hints/Edit.cshtml.cs
+++ b/ServerCore/Pages/Hints/Edit.cshtml.cs
@@ -58,11 +58,11 @@ namespace ServerCore.Pages.Hints
 
                 var teamMembers = await (from TeamMembers tm in _context.TeamMembers
                                          join HintStatePerTeam hspt in _context.HintStatePerTeam on tm.Team equals hspt.Team
-                                         where hspt.Hint.Id == Hint.Id
+                                         where hspt.Hint.Id == Hint.Id && hspt.UnlockTime != null
                                          select tm.Member.Email).ToListAsync();
                 MailHelper.Singleton.SendPlaintextBcc(teamMembers,
-                    $"Hint updated for '{Hint.Description}' on {puzzleName}",
-                    $"The new content for this hint is: '{Hint.Content}'.");
+                    $"{Event.Name}: Hint updated for {puzzleName}",
+                    $"The new content for '{Hint.Description}' is: '{Hint.Content}'.");
             }
             catch (DbUpdateConcurrencyException)
             {
--- a/ServerCore/Pages/Hints/Edit.cshtml.cs
+++ b/ServerCore/Pages/Hints/Edit.cshtml.cs
@@ -53,6 +53,16 @@ namespace ServerCore.Pages.Hints
             try
             {
                 await _context.SaveChangesAsync();
+
+                var puzzleName = await _context.Hints.Where(m => m.Id == Hint.Id).Select(m => m.Puzzle.Name).FirstOrDefaultAsync();
+
+                var teamMembers = await (from TeamMembers tm in _context.TeamMembers
+                                         join HintStatePerTeam hspt in _context.HintStatePerTeam on tm.Team equals hspt.Team
+                                         where hspt.Hint.Id == Hint.Id
+                                         select tm.Member.Email).ToListAsync();
+                MailHelper.Singleton.SendPlaintextBcc(teamMembers,
+                    $"Hint updated for '{Hint.Description}' on {puzzleName}",
+                    $"The new content for this hint is: '{Hint.Content}'.");
             }
             catch (DbUpdateConcurrencyException)
             {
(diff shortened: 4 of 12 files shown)


ServerCore/Pages/Events/CreateDemo.cshtml.cs
```diff
@@ -355,19 +355,19 @@ public async Task<IActionResult> OnPostAsync()
...
-                Team team1 = new Team { Name = "Team Bugs", Event = Event };
+                Team team1 = new Team { Name = "Team Bugs", Event = Event, IsLookingForTeammates = true };
                 _context.Teams.Add(team1);
 
-                Team team2 = new Team { Name = "Team Babs", Event = Event };
+                Team team2 = new Team { Name = "Team Babs", Event = Event, IsLookingForTeammates = true };
                 _context.Teams.Add(team2);
 
-                Team team3 = new Team { Name = "Team Buster", Event = Event };
+                Team team3 = new Team { Name = "Team Buster", Event = Event, IsLookingForTeammates = true };
```
jenetlan: Can we switch one to false so we have a wider range of states for testing?

ServerCore/Helpers/TeamHelper.cs
```diff
@@ -115,6 +115,11 @@ public static async Task DeleteTeamAsync(PuzzleServerContext context, Team team)
 
             context.TeamMembers.Add(Member);
             await context.SaveChangesAsync();
+
+            MailHelper.Singleton.SendPlaintextWithoutBcc(new string[] { team.PrimaryContactEmail, user.Email },
+                $"{user.Name} has now joined {team.Name}!",
+                $"Have a great time!");
```
jenetlan: Something here should probably mention the event name to give the recipient some context?

ServerCore/PuzzleStateHelper.cs
```diff
@@ -569,6 +569,14 @@ public static async Task UpdateTeamsWhoSentResponse(PuzzleServerContext context,
 
                     await context.SaveChangesAsync();
                     transaction.Commit();
+
+                    var teamMembers = await (from TeamMembers tm in context.TeamMembers
```
Vroo: Thanks for doing this. I wrote this but hadn't had time to verify it. This is similar to what I wrote except that we already know the relevant teams. Here's what I wroteto collect the primary emails:
```
List<String> primaryContacts = (from matchingResponse in submissionsThatMatchResponse
                               select matchingResponse.State.Team.PrimaryContactEmail).ToList();
```
and here's the replacement untested code when you said to send to all team members:
```
HashSet<int> TeamIDs = new HashSet<int>(from matchingResponse in submissionsThatMatchResponse
                                        select matchingResponse.State.Team.ID);

List<String> teamMembers = await (from TeamMembers tm in context.TeamMembers
                                  where TeamIDs.Contains(tm.Team.ID)
                                  select tm.Member.Email).ToListAsync();
```

ServerCore/PuzzleStateHelper.cs
```diff
@@ -569,6 +569,14 @@ public static async Task UpdateTeamsWhoSentResponse(PuzzleServerContext context,
...
+                                             join Submission sub in context.Submissions on tm.Team equals sub.Team
+                                             where sub.Puzzle == response.Puzzle && sub.SubmissionText == response.SubmittedText
+                                             select tm.Member.Email).ToListAsync();
+                    MailHelper.Singleton.SendPlaintextBcc(teamMembers,
+                        $"Response updated for '{response.SubmittedText}' on {response.Puzzle.Name}",
```
morganbr: Consider puzzle name first

ServerCore/PuzzleStateHelper.cs
tabascq: How on earth did I miss that?!?! Thanks, I feel kind of silly right now. :-)

ServerCore/Pages/Hints/Edit.cshtml.cs
```diff
@@ -53,6 +53,16 @@ public async Task<IActionResult> OnPostAsync()
             try
             {
                 await _context.SaveChangesAsync();
+
+                var puzzleName = await _context.Hints.Where(m => m.Id == Hint.Id).Select(m => m.Puzzle.Name).FirstOrDefaultAsync();
+
+                var teamMembers = await (from TeamMembers tm in _context.TeamMembers
+                                         join HintStatePerTeam hspt in _context.HintStatePerTeam on tm.Team equals hspt.Team
+                                         where hspt.Hint.Id == Hint.Id
+                                         select tm.Member.Email).ToListAsync();
```
jenetlan: I think this needs to check what the state is / whether or not the team used the hint.

ServerCore/Pages/Hints/Edit.cshtml.cs
```diff
@@ -53,6 +53,16 @@ public async Task<IActionResult> OnPostAsync()
...
+                MailHelper.Singleton.SendPlaintextBcc(teamMembers,
+                    $"Hint updated for '{Hint.Description}' on {puzzleName}",
```
jenetlan: Should we start all of the subjects with the event name? ex.
"{Event}: Hint updated for..."

ServerCore/Pages/Teams/Apply.cshtml.cs
```diff
@@ -74,6 +74,12 @@ public async Task<IActionResult> OnGet(int teamID)
 
             await _context.SaveChangesAsync();
 
+            // TODO I am well aware that it would be far better to include a direct link to the details page,
+            // but I cannot for the life of me figure out how to generate a URL for a Razor page.
+            MailHelper.Singleton.SendPlaintextWithoutBcc(new string[] { Team.PrimaryContactEmail, LoggedInUser.Email },
+                $"{LoggedInUser.Name} is applying to join {Team.Name}",
```
jenetlan: Same comment about event name

ServerCore/Pages/Teams/Apply.cshtml.cs
```diff
@@ -74,6 +74,12 @@ public async Task<IActionResult> OnGet(int teamID)
 
             await _context.SaveChangesAsync();
 
+            // TODO I am well aware that it would be far better to include a direct link to the details page,
+            // but I cannot for the life of me figure out how to generate a URL for a Razor page.
+            MailHelper.Singleton.SendPlaintextWithoutBcc(new string[] { Team.PrimaryContactEmail, LoggedInUser.Email },
```
jenetlan: Do you need to check if the PrimaryContactEmail is null first? Will it just bounce if that's the case?

ServerCore/PuzzleStateHelper.cs
jenetlan: Event name

ServerCore/PuzzleStateHelper.cs
```diff
@@ -569,6 +569,14 @@ public static async Task UpdateTeamsWhoSentResponse(PuzzleServerContext context,
...
+                        $"The new response for this submission is: '{response.ResponseText}'.");
```
jenetlan: Should we include a bonus comment if the submissions was changed to an answer?

ServerCore/PuzzleStateHelper.cs
morganbr: I'm not certain, but it's possible the HashSet version could end up pulling all TeamMembers from the database and then checking them in the frontend's HashSet. If performance is an issue on this page, we might need to measure it.

tabascq: After looking at the code more carefully: I do think it is important to send to all players rather than primary contacts because most teams do not have a group email and you never know who's solving what. The old site did it this way. But I will be careful thinking about what data I might already have. And Morgan's statement concerns me enough that I'm not sure which version I'll use. But I'll definitely pick one of them :-)

ServerCore/Pages/Hints/Edit.cshtml.cs
```diff
@@ -53,6 +53,16 @@ public async Task<IActionResult> OnPostAsync()
             try
             {
                 await _context.SaveChangesAsync();
+
+                var puzzleName = await _context.Hints.Where(m => m.Id == Hint.Id).Select(m => m.Puzzle.Name).FirstOrDefaultAsync();
```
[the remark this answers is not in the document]
tabascq: Actually I can't, because Hint.Puzzle is null. See the code a few lines above, at line 45.

ServerCore/PuzzleStateHelper.cs
```diff
@@ -569,6 +569,14 @@ public static async Task UpdateTeamsWhoSentResponse(PuzzleServerContext context,
 
                     await context.SaveChangesAsync();
                     transaction.Commit();
+
```
morganbr: The query above and possibly below need to ensure that the team's submission was for this puzzle. Otherwise we get a bug where if a team submitted "foo" for puzzle A, and puzzle B adds "foo" as a response, this logic triggers for puzzle B

ServerCore/Pages/Hints/Edit.cshtml.cs
tabascq: In the name of brevity, I am only going to do this for registration since we may have parallel registration, but you ought to only be *playing* in one event at a time.

ServerCore/Pages/Teams/Apply.cshtml.cs
tabascq: We should have a separate work item to disallow null primary contacts.

ServerCore/PuzzleStateHelper.cs
tabascq: Nah, the response will make that clear. "Correct!"

ServerCore/PuzzleStateHelper.cs
tabascq: Indeed!

ServerCore/Pages/Teams/Apply.cshtml.cs
Vroo: The mail code will drop any invalid email address including null.

Vroo: I just realized mail code will probably get an exception if all the email addresses are invalid and you're not doing bcc. But since we use SendMailAsync without an async method or callback, I think that exception will just get dropped.

ServerCore/Pages/Hints/Edit.cshtml.cs
tabascq: I take it back, I just did it everywhere.

ServerCore/Pages/Teams/Apply.cshtml.cs
tabascq: Agreed but for the applicant's benefit we should make sure the team is configured to see the request. This code is safe to check in before we do that part, thanks to your error handling code.

ServerCore/PuzzleStateHelper.cs
tabascq: I'm opting to leave this as is for now but I am keeping Bruce's alternative in mind. I'll try to measure this when I can, by changing a response in the sandbox after it's deployed and configured.

ServerCore/Pages/Teams/Apply.cshtml.cs
tabascq: Ditto the player's email, we should make that required as well if it isn't already.

Vroo: Yes. Here's the regex I think we should validate against
`/[^@\s]+@[.A-Za-z0-9-]+(\s+[,;]\s+[^@\s]+@[.A-Za-z0-9-]+)/`.
There are lots of invalid email addresses this will match and this errs on the side of avoiding false negatives vs. avoiding false positives.

ServerCore/Pages/Teams/Create.cshtml
```diff
@@ -35,17 +35,22 @@ else
                 <span asp-validation-for="Team.Name" class="text-danger"></span>
             </div>
             <!-- TODO: Conditionally show/hide based on event property (not needed for PuzzleHunt)
-            <div class="form-group">
-                <label asp-for="Team.RoomID" class="control-label"></label>
-                <input asp-for="Team.RoomID" class="form-control" />
-                <span asp-validation-for="Team.RoomID" class="text-danger"></span>
-            </div>
-            -->
+    <div class="form-group">
```
morganbr: Looks like something happened to indentation here

ServerCore/Pages/Teams/Edit.cshtml
```diff
@@ -26,17 +26,22 @@
                 <span asp-validation-for="Team.Name" class="text-danger"></span>
             </div>
             <!-- TODO: Conditionally show/hide based on event property (not needed for PuzzleHunt)
-            <div class="form-group">
-                <label asp-for="Team.RoomID" class="control-label"></label>
-                <input asp-for="Team.RoomID" class="form-control" />
-                <span asp-validation-for="Team.RoomID" class="text-danger"></span>
-            </div>
-            -->
+    <div class="form-group">
```
morganbr: Same indentation issue

ServerCore/Areas/Identity/Pages/Account/Edit.cshtml.cs
```diff
@@ -38,6 +38,15 @@ public async Task<IActionResult> OnGetAsync(int userId, string returnUrl = null)
 
         public async Task<IActionResult> OnPostAsync(string returnUrl = null)
         {
+            if (string.IsNullOrEmpty(PuzzleUser.Email))
```
morganbr: Would it work to at [Required] to PuzzleUser.Email and Team.PrimaryContactEmail? That would do this automatically

morganbr: Consider String.IsNullOrWhitespace

PR #397, opened 2019-04-18
Event Cache

tabascq:
Caches lookups from the Events table, since we make a lot of calls to this. This eliminates 5 server roundtrips per page load. The cache expires after 1 minute and is shared among all users and threads.

### DIFF
--- a/ServerCore/Helpers/EventHelper.cs
+++ b/ServerCore/Helpers/EventHelper.cs
@@ -1,4 +1,6 @@
 using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.EntityFrameworkCore;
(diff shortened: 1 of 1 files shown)


tabascq:
@morganbr : Note that I could extend this by moving the bottom 3 queries on PuzzleUser to Event, and having Event implement its own lookup cache. Since the Event is discarded every minute, there goes the cache. I'd have the lookup caches on Event key off of userId, so that people would not need to call PuzzleUser.GetPuzzleUserForCurrentUser either.

tabascq:
#399 is the beefier version. All told I am shaving off 12-15 SQL queries from the Puzzles page.

morganbr:
As we discussed on Wednesday, it would be preferable to find a way to test this. That said, if we're going to try caching the event, it might make sense to try the ASP.NET Core MemoryCache (https://docs.microsoft.com/en-us/aspnet/core/performance/caching/memory?view=aspnetcore-2.2), which deals with eviction.

tabascq:
Yep, and I said I would write it as a backstop and a bar to try to clear. I said we'd only take it if other approaches did not go far enough. #400 is obviously a great step, though this does more in some areas.

I tried to read that MemoryCache link and it makes little sense to me.

tabascq:
I've read the link above a couple more times, and I guess I just don't see the benefits it provides over a simple dictionary. Why do more work for something simple? In particular, why force similar keys into the same cache namespace?

tabascq:
Abandoning; if we cache we should do it right.

PR #398, opened 2019-04-20
Display team lists in alphabetical order, fixes #377

### LINKED ISSUE #377: Sorted team list change broke team list
Change #364 to sort the team list triggered DuplicateKeyExceptions on the live site. It might be a bug in the join when teams have more than one member. The change was backed out. @jaylorch , you might want to take another shot at this one.

jaylorch:


jaylorch:
Not only does this alphabetize the teams, it should improve performance dramatically since it pushes the COUNT() operation to the database.  The code this replaces does the Count() locally on the server.

jaylorch:
Oh, hmm, my join operates on *all* teams, not just ones for this event.  Let me fix that....

jaylorch:
OK, the join now operates only on teams from the requested event.

jaylorch:
I don't know what was wrong with the original version.  And when I looked at the SQL it was generating, it was really complicated and I couldn't make sense of it.  But this new version creates a nice, compact, understandable SQL query.

jaylorch:
The failing of the old version was that it was generating records with duplicate keys, so that ToDictionaryAsync was crashing when trying to create a dictionary with multiple instances of the same key.

jaylorch:
A possible suspect in bug #377 is the fact that I was trying to be fancy and do the two queries overlapping.  Maybe the Entity Framework doesn't like that?  So in this check-in I'm not being fancy; I'm await'ing the result of the first query before launching the second.

tabascq:
Since the query looks exactly the same to me, I can't say that I have 100% confidence that this is the reason. I'm also not enough of an EF and/or SQL expert to be able to determine the issue, and I'm not enough of a git expert to efficiently roll back again if this goes wrong again. I'm OK with us trying it Wednesday, but I'd like to not approve it until then so it does not get accidentally merged in the interim and cause potential headaches.

If you ever reproed the original failure on your own machine, and are confident that this change has made such failures on your own machine go away, I'd be a little more willing to experiment. Let me know if that's the case.

jaylorch:
Alas, I could never reproduce the original failure on my own machine.  So, yes, Wednesday sounds like a good time for you to try it, when you're in the same room as people who can deploy and un-deploy.

PR #401, opened 2019-04-20
Jessica/all submissions

digit01Wave:
Make it so that the "View all correct answers" link on the Play page for the player shows all submissions. Currently, it only shows the answers to puzzles they had submitted to. 
With this change, we will now also see the puzzles automatically solved or marked as solved by the author


ServerCore/Pages/Teams/Answers.cshtml.cs
```diff
@@ -35,43 +35,51 @@ public async Task OnGetAsync(SortOrder? sort, int teamId)
...
             TeamID = teamId;
 
-            IQueryable<Submission> submissions = _context.Submissions
+            IQueryable<PuzzleStatePerTeam> puzzleStates = _context.PuzzleStatePerTeam
+                .Where((state) => state.TeamID == teamId && state.SolvedTime != null);
+
+            IQueryable<Submission> correctSubmissions = _context.Submissions
                 .Where((s) => s.TeamID == teamId && s.Response.IsSolution);
 
+            IQueryable<SubmissionView> finalSubmissions = puzzleStates.GroupJoin(
```
tabascq: The potential performance of this scares me, just because the query looks complex and I'm not confident that it's only making one server call. If you've measured the SQL traffic and are confident that this is a single call, OK.

But if not, then I feel like all you need is to evaluate puzzleStates.ToListAsync(), correctSubmissions.ToDictionaryAsync(s => s.PuzzleId), and then finalSubmissions is easy to build by enumerating puzzleStates and doing dictionary lookup in that dictionary.

ServerCore/ServerCore.csproj
```diff
@@ -9,12 +9,20 @@
   <ItemGroup>
     <Compile Remove="DataModel\**" />
     <Compile Remove="Data\**" />
+    <Compile Remove="Pages\Manage\**" />
```
tabascq: Is this part intentional? If so why?

digit01Wave: it did it automatically. Removing it now. Tested and everything still works

Issue #269, opened 2019-03-07
Add messages to author responses / player submissions explaining that the answer will be normalized


jenetlan:
We should put text for authors / players reminding them that responses / submissions are normalized (changed to all capitols, whitespace removed)

Issue #267, opened 2019-03-07
Update hint confirmation message if hint is free


jenetlan:
Right now when you unlock a hint with a cost of 0 you get the message "Are you sure you want to use **0** hint tickets to unlock?"

We should update that to special case it with something like "Are you sure you want to unlock this free hint?"

Issue #266, opened 2019-03-07
Simplify fastest solves display by reducing the precision


jenetlan:
Fastest solves currently shows 7 digits of precision. It does not need to show 7 digits of precision.

PR #402, opened 2019-04-20
Jessica/puzzle day bugs

digit01Wave:
Fixed the following minor bugs:
1. Reduce precision of fastest solves page (#266 )
2. Update hint confirmation message when hint is free (#267 )
3. Add message to player submissions that reminds how submissions are processed (#269)
4. Removed empty and unused Play and Manage folders from the csproj


ServerCore/Pages/Teams/Hints.cshtml
```diff
@@ -114,11 +114,15 @@ You have <b>
                         <form method="post" style="display:inline-block">
                             @if (Model.Team.HintCoinCount >= hint.AdjustedCost)
                             {
+                                var clickString = hint.AdjustedCost == 0
+                                    ? "return confirm('Are you sure you want to unlock this free hint?')"
```
tabascq: I like the direction, but: Do we even need to have confirmation for free hints? Seems unnecessary. Maybe "return true" or "".

digit01Wave: I think it'll be good to still have a confirmation for two reasons.
1. Solving pride. Sometimes I want to solve a puzzle with no hints even if they were free so I don't want to even accidentally see a hint.
2. It would be good to keep the same pattern of confirmation for hints so that people will always know that they will have less chance of "accidentally" unlocking  a hint

tabascq: Works for me! Thanks for the rationale.

PR #403, opened 2019-04-20
Local Time Zone (fixes #102)

### LINKED ISSUE #102: DateTimes displayed to users should be localized
Times are stored in the DB as UTC+0 but should maybe be shown to users in their local time zone. This crops up in a few places

tabascq:
All times displayed in the UX for players, authors, and admins now appear in the browser's local time zone, with the exception of the event creation/properties because I don't know how to do the time zone transitions for editable fields.

Internet wisdom is that the only way to do this well is to do it in Javascript because anything done on the server will use the server timezone instead of the user timezone. I searched around for a while to locate the easiest-looking version of this, and it seems to be working properly.


ServerCore/Pages/Teams/Status.cshtml
```diff
@@ -65,7 +65,7 @@
                     }
                     else
                     {
-                        @Html.DisplayFor(modelItem => item.SolvedTime)
+                        <time>@item.UnlockedTime.Value.ToString("yyyy-MM-ddTHH:mm:ssZ")</time>
```
jenetlan: Should be item.SolvedTime?

ServerCore/Pages/Teams/Answers.cshtml
```diff
@@ -43,7 +43,7 @@
         {
         <tr>
             <td>
-                @item.TimeSubmitted
+                <time>@item.TimeSubmitted.ToString("yyyy-MM-ddTHH:mm:ssZ")</time>
```
jenetlan: Any chance the format string could be part of the function so we don't have a copy-pasted string in quite so many places?

ServerCore/Pages/Hints/AuthorIndex.cshtml
```diff
@@ -77,7 +77,7 @@
                     @Html.DisplayFor(modelItem => item.Hint.Cost)
                 </td>
                 <td>
-                    @Html.DisplayFor(modelItem => item.UnlockTime)
+                    <time>@item.UnlockTime?.ToString("yyyy-MM-ddTHH:mm:ssZ")</time>
```
jenetlan: Is there a reason this is the only one with a null check?

ServerCore/wwwroot/js/site.js
```diff
@@ -1 +1,6 @@
 ﻿// Write your Javascript code.
+$("time").each(function (elem) {
+    var utctimeval = $(this).html();
+    var date = new Date(utctimeval);
+    $(this).html(date.toLocaleString());
```
jenetlan: Does it get translated to local time when it gets put into a Date object? I think toLoacleString affects the formatting of the date, not the timezone

PR #404, opened 2019-04-21
Count up hints and costs

tabascq:
Sum up the total hint cost for each puzzle on the puzzle checklist page.

### DIFF
--- a/ServerCore/Pages/Puzzles/Checklist.cshtml
+++ b/ServerCore/Pages/Puzzles/Checklist.cshtml
@@ -80,7 +80,7 @@
                     }
                 </td>
                 <td>
-                    @Html.DisplayFor(modelItem => item.Puzzle.MinPrerequisiteCount)
+                    @Html.DisplayFor(modelItem => item.Puzzle.MinPrerequisiteCount) of:
                 </td>
                 <td>
                     @Html.DisplayFor(modelItem => item.Prerequisites)
@@ -102,7 +102,7 @@
                     @Html.DisplayFor(modelItem => item.Hints)
                 </td>
                 <td>
-                    @Html.DisplayFor(modelItem => item.TotalHintCost)
+                    $@Html.DisplayFor(modelItem => item.TotalHintCost)
                 </td>
             </tr>
         }
@@ -122,7 +122,7 @@
                 @Model.TotalHints
             </td>
             <td>
-                @Model.TotalHintCost
+                $@Model.TotalHintCost
             </td>
         </tr>
     </tbody>
--- a/ServerCore/Pages/Puzzles/Checklist.cshtml
+++ b/ServerCore/Pages/Puzzles/Checklist.cshtml
@@ -45,60 +45,85 @@
             <th>
                 Hint count
             </th>
+            <th>
+                Hint cost
+            </th>
         </tr>
     </thead>
     <tbody>
         @foreach (var item in Model.PuzzleData)
         {
+            <tr>
+                <td>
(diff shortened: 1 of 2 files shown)


ServerCore/Pages/Puzzles/Checklist.cshtml
```diff
@@ -45,60 +45,85 @@
...
-            </td>
-            <td>
-                @if (item.Responses.HasAnswer)
-                {
-                    <p>&#10004;</p>
-                }
-            </td>
-            <td>
-                @Html.DisplayFor(modelItem => item.Hints)
+                @Model.TotalHintCost
```
Vroo: Since there are a bunch of columns on this page and the titles scroll off, I might suggest adding a literal `¢` in here.

Similarly on the min pre-regs column you could add the literal text ` of` so it reads `1 of`.

ServerCore/Pages/Puzzles/Checklist.cshtml.cs
```diff
@@ -73,7 +79,28 @@ public async Task<IActionResult> OnGetAsync()
...
+                int totalHintCostThisPuzzle = 0;
+                int hintsCountThisPuzzle = 0;
+                if (hints != null)
+                {
+                    int minHintCostThisPuzzle = 0;
+
+                    hintsCountThisPuzzle = hints.Count();
+                    foreach (int cost in hints)
+                    {
+                        minHintCostThisPuzzle = Math.Min(minHintCostThisPuzzle, cost);
```
Vroo: For clarity, I might call this `totalDiscount` or `fullDiscount`.

ServerCore/Pages/Puzzles/Checklist.cshtml.cs
```diff
@@ -73,7 +79,28 @@ public async Task<IActionResult> OnGetAsync()
...
+                    }
+
+                    totalHintCostThisPuzzle = Math.Abs(minHintCostThisPuzzle);
+
+                    foreach (int cost in hints)
+                    {
+                        totalHintCostThisPuzzle += Math.Max(0, cost + minHintCostThisPuzzle);
```
Vroo: This code confused me at first but I think will compute the correct value. I think I would include this comment:
```
// Negative cost hints will be ignored because Max(0,negative) is 0.
// Positive cost hints will only be counted for the cost above the discount.
// (The discount is counted against each other hint.)
```

ServerCore/Pages/Puzzles/Checklist.cshtml.cs
```diff
@@ -73,7 +79,28 @@ public async Task<IActionResult> OnGetAsync()
...
+                {
+                    int minHintCostThisPuzzle = 0;
+
+                    hintsCountThisPuzzle = hints.Count();
+                    foreach (int cost in hints)
+                    {
+                        minHintCostThisPuzzle = Math.Min(minHintCostThisPuzzle, cost);
+                    }
+
+                    totalHintCostThisPuzzle = Math.Abs(minHintCostThisPuzzle);
```
Vroo: the minHintCost will always be non-positive so this can be just `-` instead of `Math.Abs`. I think that's more clear.

ServerCore/Pages/Puzzles/Checklist.cshtml.cs
```diff
@@ -83,8 +110,12 @@ public async Task<IActionResult> OnGetAsync()
                     Prerequisites = prereqs != null ? string.Join(", ", prereqs) : "",
```
Vroo: It would be good if the prereqs were sorted here. Very noticeable when two puzzles have the same prereqs but they appear different.

PR #406, opened 2019-04-22
Bulk Mailer

tabascq:
Add a bulk mailer for all players and for all team primary contacts. Currently the mailer is under strict admin control in order to make sure that we have a coordinated sender who knows about our monthly limits.


ServerCore/Pages/Events/Mailer.cshtml.cs
```diff
@@ -0,0 +1,78 @@
...
+                    foreach (string p in primaries)
+                    {
+                        primariesSplit.AddRange(p.Split(',', ';'));
+                    }
+
+                    addresses = primariesSplit;
+                    break;
+            }
+
+            MailHelper.Singleton.SendPlaintextBcc(addresses, MailSubject, MailBody);
```
morganbr: If addresses is empty, we probably shouldn't try to send

morganbr: Did we end up with an async version of sending mail? If so, it's probably better to use it for large batches

ServerCore/Pages/Events/Mailer.cshtml.cs
```diff
@@ -0,0 +1,78 @@
...
+            if (!ModelState.IsValid)
+            {
+                return Page();
+            }
+
+            IEnumerable<string> addresses = Enumerable.Empty<string>();
+
+            switch(Group)
+            {
+                case MailGroup.AllPlayers:
```
morganbr: I suspect we also need a way to email small teams since we have a lot of them.

morganbr: For PuzzleDay we may need to email individual teams (e.g. "Go to the live event that you forgot"). I can't immediately think of a case for Hunt though.

ServerCore/Pages/Events/Mailer.cshtml.cs
tabascq: Not going to worry about this edge case right now, it's not going to happen in practice for PH20

tabascq: MailHelper calls SendMailAsync() and does not await the result.

ServerCore/Pages/Events/Mailer.cshtml.cs
tabascq: Added the ability to mail a team.

tabascq: Added the ability to mail a team.

PR #408, opened 2019-04-24
Only count puzzle toward solve count if it's worth 0 hint coins

jaylorch:


Vroo:
Why are we doing this? The strange thing is that I looked through the code to see how this was used and the values are written and never read. The only thing I can think of is that this is used by Jay's metameta and there are a few missing comments.

In that case, does this mean that Roy's puzzles can be worth 10 points again as they will not contribute to the metameta since they award hint tokens?

tabascq:
That's exactly why we're doing it. Roy can also make his meta worth 50. We discussed this all with him prior to the change.

Vroo:
Thanks. No rush (can wait until after the hunt) but it would be helpful somewhere to have a comment that explains what this is for, since future hunts might want something different. Renaming it MetaMetaSyncController might help clarify.

jaylorch:
It's called the SyncController out of a perhaps-misguided hope that it can be used in general, perhaps even unforeseen, ways in the future, not necessarily always for meta-metas.  For instance, if you want to write a one-off whistle-stop-style puzzle that just needs to sync annotations, you can do that and just ignore the solve-count feature.

For now, it's got various things hard-coded (like the way it decides the solve count), but in future we can have this vary based on to-be-added fields in puzzles.

PR #415, opened 2019-04-25
MinJsHackAttempt

tabascq:
Seems like this file doesn't want to build in our current environment. I've made my own version, which is semi-minified; let's see if it works.


jenetlan:
fyi this worked

PR #421, opened 2019-04-27
Errata Visibility

tabascq:
One thing that came out of the last PH20 meeting was that we needed a way for players to know that errata for a puzzle existed. This is that way: a simple caution icon plus an explanation at the bottom for what it means.

### DIFF
--- a/ServerCore/Pages/Teams/Play.cshtml
+++ b/ServerCore/Pages/Teams/Play.cshtml
@@ -54,7 +54,7 @@
                 @Html.DisplayFor(modelItem => item.Group)
             </td>
             <td>
-                @if (item.Errata != null)
+                @if (!string.IsNullOrWhiteSpace(item.Errata))
                 {
                 <text>&#x26A0; </text>
                 }
--- a/ServerCore/Pages/Teams/Play.cshtml
+++ b/ServerCore/Pages/Teams/Play.cshtml
@@ -54,15 +54,19 @@
             <td>
                 @if (item.Puzzle.CustomURL != null)
                 {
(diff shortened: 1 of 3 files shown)


ServerCore/Pages/Teams/Play.cshtml
```diff
@@ -54,15 +54,19 @@
...
                 }
                 else if (item.Content != null)
                 {
-                    @Html.ActionLink(item.Puzzle.Name, "Index", "Files", new { eventId = Model.Event.ID, filename = item.Content.ShortName }, new { target = "_blank" })
+                @Html.ActionLink(item.Puzzle.Name, "Index", "Files", new { eventId = Model.Event.ID, filename = item.Content.ShortName }, new { target = "_blank" })
                 }
                 else
                 {
-                    @Html.DisplayFor(modelItem => item.Puzzle.Name)
+                @if (item.Puzzle.Errata != null)
```
morganbr: I'm not sure if it would be null if somebody adds and removes errata. Maybe String.IsNullOrWhitespace?

ServerCore/Pages/Teams/Play.cshtml
```diff
@@ -100,3 +104,5 @@
     }
     </tbody>
 </table>
+
+<p>Note: Puzzles with a &#x26A0; before the name have errata, which will be provided on that puzzle's answer submission page.</p>
```
morganbr: Should this be at the top? This page can get long

ServerCore/Pages/Teams/Play.cshtml
tabascq: Good idea, done

ServerCore/Pages/Teams/Play.cshtml
tabascq: I don't think people need to look at it all the time. It'll be shorter initially and people will see what it means then.

PR #423, opened 2019-04-28
Play Page Perf Please

tabascq:
Updated queries to simplify the needs of the Play Page, since Morgan said it was overquerying.

### DIFF
--- a/ServerCore/Pages/Teams/Play.cshtml.cs
+++ b/ServerCore/Pages/Teams/Play.cshtml.cs
@@ -23,13 +23,15 @@ namespace ServerCore.Pages.Teams
         {
         }
 
-        public IList<PuzzleWithState> PuzzlesWithState { get; set; }
+        public IList<PuzzleView> PuzzleViews { get; set; }
 
         public int TeamID { get; set; }
 
         public SortOrder? Sort { get; set; }
 
         private const SortOrder DefaultSort = SortOrder.GroupAscending;
+
+        public bool ShowAnswers { get; set; }
 
         public async Task OnGetAsync(SortOrder? sort, int teamId)
         {
@@ -45,6 +47,8 @@ namespace ServerCore.Pages.Teams
                 throw new Exception("Not currently registered for a team");
             }
             this.Sort = sort;
+
+            ShowAnswers = Event.AnswersAvailableBegin <= DateTime.UtcNow;
 
             // all puzzles for this event that are real puzzles
             var puzzlesInEventQ = _context.Puzzles.Where(puzzle => puzzle.Event.ID == this.Event.ID && puzzle.IsPuzzle);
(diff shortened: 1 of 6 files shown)


ServerCore/Pages/Teams/Play.cshtml.cs
```diff
@@ -64,41 +68,55 @@ public async Task OnGetAsync(SortOrder? sort, int teamId)
...
                     break;
                 default:
                     throw new ArgumentException($"unknown sort: {sort}");
             }
 
+            PuzzleViews = await visiblePuzzlesQ.ToListAsync();
+
             Dictionary<int, ContentFile> files = await (from file in _context.ContentFiles
                                                         where file.Event == Event && file.FileType == ContentFileType.Puzzle
                                                         select file).ToDictionaryAsync(file => file.Puzzle.ID);
```
morganbr: My guess is that this line is the real culprit for over-querying -- I think file.Puzzle is triggering a query for every file

ServerCore/Pages/Teams/Play.cshtml.cs
```diff
@@ -64,41 +68,55 @@ public async Task OnGetAsync(SortOrder? sort, int teamId)
...
-                PuzzlesWithState.Add(new PuzzleWithState(visiblePuzzle.Puzzle, visiblePuzzle.State, content));
+                files.TryGetValue(puzzleView.ID, out ContentFile content);
+                puzzleView.Content = content;
+            }
+
+            if (ShowAnswers)
+            {
+                Dictionary<int, ContentFile> answers = await (from file in _context.ContentFiles
+                                                            where file.Event == Event && file.FileType == ContentFileType.Answer
+                                                            select file).ToDictionaryAsync(file => file.Puzzle.ID);
```
morganbr: This is likely a copy of the puzzle file over-query (but having this chunk of code does save us other overquerying once answers unlock)

ServerCore/Pages/Teams/Play.cshtml.cs
tabascq: OHHH makes sense.

tabascq: @morganbr , can I just expose ContentFile.PuzzleID? It's commented out with guidance I don't understand. ContentFile already exposes EventID...

morganbr: You can give it a shot. I seem to recall it might have had something to do with getting funny migrations, so please make sure adding a migration is a no-op.

tabascq: namespace Data.Migrations
{
    public partial class PuzzleID : Migration
    {
        protected override void Up(MigrationBuilder migrationBuilder)
        {

        }

        protected override void Down(MigrationBuilder migrationBuilder)
        {

        }
    }
}

tabascq:
I included my empty migration report, so I think we're good to go. Since the migration had nothing in it, I removed it. I also tested things prior to creating the migration, and everything worked fine.

PR #425, opened 2019-04-28
Fix #424: Don't send emails for Response/Hint Updated when there are …

### LINKED ISSUE #424: Response/Hint updated emails are sent even when there are no recipients (yet)
When a Response/Hint is changed, the updated message is sent out (to just the visible address for bcc's) even when there is nobody to receive the message. Not a huge problem, but unnecessarily uses mailer quota.

royleban:
Fix #424: Don't send emails for Response/Hint Updated when there are no actual recipients


royleban:
I considered that and decided against it because MailHelper should allow an email being sent to just the bcc address. In our usage, that address is a mailbox that is never read, but in some other usage, it could be GC who wants to see the emails.

royleban:
... and the send/nosend decision should be made outside of MailHelper, say by a configuration setting that is looked at when certain emails are sent.

tabascq:
I'm inclined to agree with Bruce on this. MailHelper takes one list of recipients - if that list turns out to be empty then there's no point in sending the mail. Might as well fix it once for all cases rather than have everyone need to remember to check (and some of the checks, like verifying valid address structure, are private).

vidapour:
Taking #426 instead.

PR #426, opened 2019-04-28
do not send mail to no one #424

Vroo:


royleban:
See also my different commit for this bug

tabascq:
I think this is a reasonable way to fix the issue, but I do not believe this is a pressing need for PH20 as we're past the point where we'll be updating responses for nobody. I do think we'll want this as Puzzleday and future events ramp up and fill out the database, so I think it's worth taking afterwards. I just won't want the added risk right now.

tabascq:
This change has merge conflicts due to other changes since the PR was created and is now missing a case as a result; I'm going to make a new PR that does roughly the same thing, and put you on it for review.

tabascq:
Closing in favor of #759

PR #428, opened 2019-04-29
Sort the module containing the final puzzle at the end.

Vroo:


tabascq:
FYI, I am working on this page to address a serious perf issue (PR #423) that may affect our ability to scale to a full event. That is much higher priority and there may not be time for this to get into PH20. Just setting expectations.

Vroo:
Not needed anymore. We can control ordering with Html.Raw

PR #429, opened 2019-04-29
Send email in batches of <50 recipients.

Vroo:


ServerCore/Helpers/MailHelper.cs
```diff
@@ -21,6 +21,12 @@ public class MailHelper
     private const string ToAddressForBCC = "msph-bcc@puzzlehunt.org";
     private static readonly char[] EmailSeparators = { ',', ';' };
 
+    /// <summary>
+    /// Mailjet has an undocumented limit of 50 recipients at a time.
+    /// This constant leaves 1 for the bcc address and 1 just in case.
+    /// </summary>
+    private const int MAX_RECIPIENTS = 48;
```
tabascq: Can we go to 40 for extreme safety?

ServerCore/Helpers/MailHelper.cs
```diff
@@ -192,16 +210,39 @@ private static void AddRecipients(MailAddressCollection collection, IEnumerable<
             string[] addresses = recipient.Split(EmailSeparators);
             foreach (string address in addresses)
             {
-                try
-                {
-                    collection.Add(new MailAddress(address.Trim()));
-                }
-                catch (FormatException)
+                string a = address.Trim();
+                if (a != "")
```
tabascq: !string.IsNullOrWhitespace(a)

ServerCore/Helpers/MailHelper.cs
Vroo: done

ServerCore/Helpers/MailHelper.cs
Vroo: it can't be null or we'd have an NPE on the previous line. I changed it although I think it's a little less clear this way.

tabascq: oh right sorry - can that previous line become address?.Trim() so we're covered?

tabascq: I'm not sure that helps either, just in extreme paranoia mode right now.

tabascq: Yeah, I've talked myself out of needing the address?.Trim().

Vroo: The result of recipient.split never includes nulls. In fact, it's never even an empty array. If the string is empty, it's an array with one empty string in it.

tabascq: yup yup yup

PR #432, opened 2019-04-30
add async logic and delay to mail sending

Vroo:


ServerCore/Helpers/MailHelper.cs
```diff
@@ -123,22 +124,23 @@ public void SendPlaintextWithoutBcc(IEnumerable<string> recipients, string subje
...
+    private async void SendInternalAsync(IEnumerable<string> recipients, bool bcc, string subject, string body, bool isHtml)
     {
         List<string> addresses = FlattenAddressLists(recipients);
         while (addresses.Count > 0)
         {
             int n = Math.Min(addresses.Count, MAX_RECIPIENTS);
             List<string> someAddresses = addresses.GetRange(0, n);
             addresses.RemoveRange(0, n);
-            SendInternal2(someAddresses, bcc, subject, body, isHtml);
+            SendInternal2Async(someAddresses, bcc, subject, body, isHtml);
```
tabascq: We should await this too

Vroo: I didn't do that because
(1) when I add the await there I get an error "Cannot await 'void'" and I couldn't figure out how to fix that.
(2) There's an await inside SendInternal2Async at the bottom where the real work is (the api call) and the code up to that point is fairly simple so it seemed like that stacked await was just extra overhead. If I think of await as a promise, there's not much reason to create a promise of a promise.

tabascq: Oh right - the pattern here is to make SendInternal2Async be a Task instead of a void. I missed that part. Right now, SendInternal2Async is returning when the SendMailAsync call begins, not when it ends. If you make it return a Task then it will return the object that can be awaited to know when the send is complete.

ServerCore/Helpers/MailHelper.cs
```diff
@@ -123,22 +124,23 @@ public void SendPlaintextWithoutBcc(IEnumerable<string> recipients, string subje
...
+            await Task.Delay(1000);
```
morganbr: What's the reason for the delay? Does the mail server force us to do it? This is unfortunate since a request timeout could get the sending killed.

tabascq: This is from some scrambling to get all mail messages sent from Sunday's mail. We were trying to figure out what was wrong. The mails did manage to send, and the throttling was identified to be a provider issue, so this may not be strictly necessary and we should do what is deemed to be safest. I think it's a bit of a dice roll, and we should deliberately roll those dice and send at least one scale test tonight.

Vroo: From discussion between Kenny, me, and Roy I think we are going to try to avoid taking this commit. That's why I did not update after Kenny's other comment. We got a note from Mailjet that said:

> Our Compliance Team has re-reviewed your case and it appears you were affected by an issue on our end still applying a limit to your account that has now been removed. We apologize for the inconvenience this has caused you. 

They fixed this by re-blocking our account and then unblocking it 3 minutes later which really confused us when the first mail came in. I sent a test with 200 mails, half going to 2 addresses without bcc and half going to 2 addresses with bcc (3 addresses including msph-bcc) and they all got delivered fairly within 5 minutes. Note that a hint update that gets sent to 1000 people will only be 25 mails since they are sent in batches of 40, not 1000 separate mails.

PR #433, opened 2019-05-02
Display raw links for files

morganbr:
Display raw links for files to allow external embedding


tabascq:
Is this for, like, now? Please let's discuss before merging anything.

morganbr:
Discussed offline and we're good to merge now.

PR #440, opened 2019-05-07
Hint Visibility and Perf

tabascq:
Make all hints visible after the event is over. This helps with archiving.
Improve the perf of the existing hints-taken page.

### DIFF
--- a/ServerCore/Pages/Hints/AuthorIndex.cshtml.cs
+++ b/ServerCore/Pages/Hints/AuthorIndex.cshtml.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authorization;
@@ -18,7 +19,7 @@ namespace ServerCore.Pages.Hints
         {
         }
 
-        public List<HintStatePerTeam> HintStatePerTeam { get; set; }
+        public List<HintView> HintViews { get; set; }
 
         public Puzzle Puzzle { get; set; }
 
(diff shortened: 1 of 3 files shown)


ServerCore/Pages/Hints/AuthorIndex.cshtml.cs
```diff
@@ -38,28 +39,36 @@ public async Task<IActionResult> OnGetAsync(int? puzzleId, int? teamId, SortOrde
...
+                        HintViews = await _context.HintStatePerTeam.Where((h) => h.UnlockTime != null && h.Team.Event == Event)
+                            .Select(hspt => new HintView { TeamId = hspt.TeamID, TeamName = hspt.Team.Name, PuzzleId = hspt.Hint.Puzzle.ID, PuzzleName = hspt.Hint.Puzzle.Name, Description = hspt.Hint.Description, Cost = hspt.Hint.Cost, UnlockTime = hspt.UnlockTime  })
+                            .ToListAsync();
                     }
                     else
                     {
-                        HintStatePerTeam = await _context.HintStatePerTeam.Where((h) => h.UnlockTime != null && h.TeamID == teamId).ToListAsync();
+                        HintViews = await _context.HintStatePerTeam.Where((h) => h.UnlockTime != null && h.TeamID == teamId)
+                            .Select(hspt => new HintView { TeamId = hspt.TeamID, TeamName = hspt.Team.Name, PuzzleId = hspt.Hint.Puzzle.ID, PuzzleName = hspt.Hint.Puzzle.Name, Description = hspt.Hint.Description, Cost = hspt.Hint.Cost, UnlockTime = hspt.UnlockTime })
```
morganbr: Since this has gotten more complex, it might be good to share with the previous query.

morganbr: Or possibly all of the queries

ServerCore/Pages/Teams/Hints.cshtml.cs
```diff
@@ -75,6 +75,15 @@ private async Task<IList<HintWithState>> GetAllHints(int puzzleID, int teamID)
                 {
                     hint.Discount = discount;
                 }
+
+                // if the event is over, show all hints
```
morganbr: To allow players to control hints they see afterward, it would be good to have something that special cases these for archiving

PR #442, opened 2019-05-07
Display all hint responses

tabascq:


ServerCore/Pages/Hints/Responses.cshtml.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            if (EventRole == EventRole.admin)
+            {
+                HintViews = await _context.Hints.Where(h => h.Puzzle.Event == Event)
+                    .Select(h => new HintView { PuzzleId = h.Puzzle.ID, PuzzleName = h.Puzzle.Name, Description = h.Description, Content=h.Content, Cost = h.Cost })
+                    .ToListAsync();
+            }
+            else
+            {
+                // Surely there is a way to get a join to do a bunch of this work, but joins are simply not for me. Someone else can fix later.
```
morganbr: To do this sort of join, you could grab the query from GetPuzzlesForAuthorAndEvent

morganbr: And it might be nice to do so since this change has a little less time pressure.

ServerCore/Pages/Hints/Responses.cshtml.cs
```diff
@@ -0,0 +1,110 @@
...
+                case SortOrder.PuzzleDescending:
+                    HintViews.Sort((a, b) => -a.PuzzleName.CompareTo(b.PuzzleName));
+                    break;
+                case SortOrder.DescriptionAscending:
+                    HintViews.Sort((a, b) => a.Description.CompareTo(b.Description));
+                    break;
+                case SortOrder.DescriptionDescending:
+                    HintViews.Sort((a, b) => -a.Description.CompareTo(b.Description));
+                    break;
+                case SortOrder.CostAscending:
```
morganbr: Consider taking the absolute value of costs for sorting

PR #446, opened 2019-05-09
Require non null/whitespace team names

morganbr:
Require team names to be filled in. This will show a validation message if a team is not filled in.


morganbr:
Need to ensure we reconcile the merge conflict between these migrations and the ones in #407

PR #447, opened 2019-05-09
Redirect players to relevant pages from team details pages for teams they're not on

jenetlan:
We had reports of issues where players would send the URL of the team to their teammates instead of the apply link. Instead of going to an unauthorized page, this change redirects players to either:
* The details page for the team they're already on
* The apply page for the team they're trying to reach

Fix for https://github.com/PuzzleServer/mainpuzzleserver/issues/444


ServerCore/Pages/Teams/Details.cshtml.cs
```diff
@@ -31,11 +32,25 @@ public async Task<IActionResult> OnGetAsync(int teamId = -1)
             {
                 return NotFound("Missing team id");
             }
-            else
+
+            // Auth checks (only allow admins OR players who are on the team)
+            int playerTeamId = await GetTeamId();
+            if (!(EventRole == EventRole.admin || (EventRole == EventRole.play && playerTeamId == teamId)))
```
morganbr: This needs to check that the user actually is an admin (this was previously handled by the base class, but that's no longer true due to efficiency issues)

ServerCore/Pages/Teams/Details.cshtml.cs
```diff
@@ -31,11 +32,25 @@ public async Task<IActionResult> OnGetAsync(int teamId = -1)
...
+                if (EventRole == EventRole.play)
+                {
+                    if (playerTeamId != -1)
+                    {
+                        return RedirectToPage("./Details", new { teamId = playerTeamId });
+                    }
+                    return RedirectToPage("./Apply", new { teamId = teamId });
+                }
+                // Everyone else gets an error
+                return NotFound("You do not have permission to veiw the details of this team.");
```
morganbr: Nit -- typo in "view'

ServerCore/Pages/Teams/Details.cshtml.cs
```diff
@@ -64,6 +79,11 @@ public async Task<IActionResult> OnGetAsync(int teamId = -1)
 
         public async Task<IActionResult> OnGetRemoveMemberAsync(int teamId, int teamMemberId)
         {
+            if (LoggedInUser == null)
```
morganbr: All of the entrypoints in this page need this check and role checks now

PR #456, opened 2019-05-16
Fix #410: Send e-mails to relevant team members when the puzzle errata is updated

### LINKED ISSUE #410: Email teams when new errata is added
Use the implemented email system to send an email to teams when new errata is added.

amgleitman:


ServerCore/Pages/Puzzles/Edit.cshtml.cs
```diff
@@ -103,11 +103,28 @@ public async Task<IActionResult> OnPostAsync(int puzzleId)
...
+                                      where puzzle.ID == puzzleId
+                                      select puzzle.Errata).FirstOrDefaultAsync();
+
             _context.Attach(Puzzle).State = EntityState.Modified;
 
             try
             {
                 await _context.SaveChangesAsync();
+
+                if (Puzzle.Errata != oldErrata)
```
morganbr: It looks like this would send mail if the errata was removed (with an empty string). I'd suggest checking String.IsNullOrWhitespace(Puzzle.Errata).

amgleitman: Fair point.  We should probably still send mail in this case, just with a different message.

PR #457, opened 2019-05-16
Add the ability to reject players, plus mail for reject/delete scenarios

tabascq:


ServerCore/Pages/Teams/Details.cshtml.cs
```diff
@@ -163,13 +163,58 @@ public async Task<IActionResult> OnGetAddMemberAsync(int teamId, int userId, int
...
+            }
+
+            var allApplications = from app in _context.TeamApplications
+                                  where app.Player == application.Player &&
+                                  app.Team.Event == Event
+                                  select app;
+            _context.TeamApplications.RemoveRange(allApplications);
+            await _context.SaveChangesAsync();
+
+            MailHelper.Singleton.SendPlaintextWithoutBcc(new string[] { team.PrimaryContactEmail, application.Player.Email },
```
jenetlan: Do we need the team primary e-mail? What do we expect them to do in that case?

ServerCore/Pages/Teams/Details.cshtml.cs
```diff
@@ -163,13 +163,58 @@ public async Task<IActionResult> OnGetAddMemberAsync(int teamId, int userId, int
...
+            {
+                return NotFound("Mismatched player and application");
+            }
+            Team team = await _context.Teams.FirstOrDefaultAsync(m => m.ID == teamId);
+            if (application.Team != team)
+            {
+                return Forbid();
+            }
+
+            var allApplications = from app in _context.TeamApplications
```
jenetlan: ...in what case can this be a list instead of one thing?

ServerCore/Pages/Teams/Details.cshtml.cs
```diff
@@ -163,13 +163,58 @@ public async Task<IActionResult> OnGetAddMemberAsync(int teamId, int userId, int
...
+                                                 where app.ID == applicationId
+                                                 select app).FirstOrDefaultAsync();
+            if (application == null)
+            {
+                return NotFound("Could not find application");
+            }
+
+            if (application.Player.ID != userId)
+            {
+                return NotFound("Mismatched player and application");
```
jenetlan: Is there a reason we're passing the player around in addition to the application? It seems like we might be able to just go off of that and simplify this a bit

ServerCore/Pages/Teams/Details.cshtml.cs
```diff
@@ -163,13 +163,58 @@ public async Task<IActionResult> OnGetAddMemberAsync(int teamId, int userId, int
...
+
+            var allApplications = from app in _context.TeamApplications
+                                  where app.Player == application.Player &&
+                                  app.Team.Event == Event
+                                  select app;
+            _context.TeamApplications.RemoveRange(allApplications);
+            await _context.SaveChangesAsync();
+
+            MailHelper.Singleton.SendPlaintextWithoutBcc(new string[] { team.PrimaryContactEmail, application.Player.Email },
+                $"{Event.Name}: Your application was rejected from {team.Name}",
```
morganbr: "rejected by"? "Application to {team.Name} was rejected"?

ServerCore/Pages/Teams/Details.cshtml.cs
tabascq: it's a required field, but there's no reason to include it here. I pulled it.

ServerCore/Pages/Teams/Details.cshtml.cs
tabascq: All the other cases treated it as a list so I did too. I don't think it can be more than one thing either.

ServerCore/Pages/Teams/Details.cshtml
```diff
@@ -274,7 +274,10 @@
                                 @Html.DisplayNameFor(model => model.Users[0].Item1.EmployeeAlias)
                             </th>
                             <th>
-                                Add player to team
+                                Approve player
+                            </th>
+                            <th>
+                                Reject player
```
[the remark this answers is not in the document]
tabascq: Changing to "reject application" so it is less personal, and I am now going to use "was not accepted" to the actual player instead of this word.

tabascq:
All comments addressed, should be good to go now!

Issue #270, opened 2019-03-07
Add getting started info to the event page by default


jenetlan:
Adding messaging to the event page (landing page) helping users figure out first actions (find a team, read the FAQ, etc.)

jenetlan:
Note: Someone expected FAQ to be linked from here, although they did later find it under resources

jbodner09:
Closing in favor of #870

PR #472, opened 2019-05-23
Adds sample puzzles to site and make frontpage more friendly to unregistered

philipzloh:
Fixes #472 and #270 

`SamplePartial.cshtml` has some ugly hard-coded stuff but that technically could be fixed later.

samplePuzzles


ServerCore/Pages/Shared/_Layout.cshtml
```diff
@@ -22,19 +17,17 @@
...
 
     <!-- TODO: Make this actually scalable and correct -->
-    @if (ev != null)
-    {
-        <link rel="stylesheet" href="~/css/@(ev.UrlString)/styles.css" />
+    @{
+        // Using ev here so that it isn't the same as Event that we bind in the <body>
+        // Works fine if we use Event, but VS will yell at you with an error in the IDE
+        var ev = (Model as ServerCore.ModelBases.EventSpecificPageModel)?.Event;
+        if (ev != null && ev.UrlString == "pd2019")
```
morganbr: Could you probe for the file or something instead of special-casing this event so this works equally well next year?

philipzloh: Tried to use a File.Exists but that required a fully-qualified path, or at least that's what I thought. Will keep digging.

PR #480, opened 2019-05-30
[DB Change] Add DC property to puzzles & remove puzzle details page

jenetlan:
Adding a data confirmation (DC) property to puzzles. In the future this can help us mark puzzles with DC on the list page and possibly include the DC icon on the submissions page, as well as allowing us to potentially adjust the lockout algorithm for DC puzzles.

Also removed the puzzle details page and redirected incoming links to the edit page instead. The details page was very far behind the properties that puzzles support and was just an extra step in getting to the useful info about each puzzle.


Data/DataModel/Puzzle.cs
```diff
@@ -96,6 +97,11 @@ public Puzzle (Puzzle source)
         /// </summary>
         public bool HintsAreCumulative { get; set; }
 
+        /// <summary>
+        /// True if this puzzle has data confirmation
+        /// </summary>
+        public bool HasDataConfirmation { get; set; } = false;
```
morganbr: Initializing to false is redundant

Data/Migrations/20190530024701_AddDCPropertyToPuzzles.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
```
morganbr: We'll need to merge between this migration and #478

PR #481, opened 2019-06-01
General-purpose Intern Meta

tabascq:
This is just a straightforward piece viewer with no frills of any kind. It verifies that the event is an intern event, in case this facility could somehow be used to get an inappropriate look at an official hunt.

Event-specific instructions can be added by creating a piece whose progress level is <= 0.


ServerCore/Pages/Pieces/InternMeta.cshtml.cs
```diff
@@ -0,0 +1,38 @@
...
+        
+        public async Task<IActionResult> OnGetAsync(int puzzleId)
+        {
+            if (!Event.IsInternEvent)
+            {
+                return Forbid("This page is only valid for intern events.");
+            }
+
+            Team team = await UserEventHelper.GetTeamForPlayer(_context, Event, LoggedInUser);
+            int solvedPuzzleCount = await PuzzleStateHelper.GetSparseQuery(_context, Event, null, team, null).Where(ps => ps.SolvedTime != null && ps.Puzzle.SolveValue >= 10).CountAsync();
```
jaylorch: It seems cleaner to just use context.PuzzleStatePerTeam.Where(state => state.Team == team) instead of GetSparseQuery.  If you do that, the code will be more self-documenting.  (As it was, I had to go look at GetSparseQuery, which is a pretty unintuitive name considering what it does, to figure out what this code did.)

ServerCore/Pages/Pieces/InternMeta.cshtml
```diff
@@ -0,0 +1,27 @@
...
+            <th>
+                @Html.DisplayNameFor(model => model.EarnedPieces[0].Contents)
+            </th>
+        </tr>
+    </thead>
+    <tbody>
+@foreach (var item in Model.EarnedPieces) {
+        <tr>
+            <td>
+                @Html.DisplayFor(modelItem => item.Contents)
```
jaylorch: A useful future frill would be the ability to have the pieces contain Markdown that gets rendered by this page.  (Alternately, you could just let pieces contain HTML that gets inserted raw into the page, but this is dangerous if you don't trust authors to sanitize their piece content.)

ServerCore/Pages/Pieces/InternMeta.cshtml.cs
```diff
@@ -0,0 +1,38 @@
...
+using Microsoft.AspNetCore.Identity;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.EntityFrameworkCore;
+using ServerCore.DataModel;
+using ServerCore.Helpers;
+using ServerCore.ModelBases;
+
+namespace ServerCore.Pages.Pieces
+{
+    [Authorize(Policy = "PlayerIsOnTeam")]
```
jaylorch: During PH20, I would have appreciated, as an author, a way to see the piece list (or at least piece count) for all teams or for a specific team.  So, as a bridge to that future, you might want to let the puzzle author view this page, not just members of the team.  (But presumably not in this initial no-frills version!)

ServerCore/Pages/Pieces/InternMeta.cshtml.cs
morganbr: Why SolveValue instead of IsPuzzle?

ServerCore/Pages/Pieces/InternMeta.cshtml
```diff
@@ -0,0 +1,27 @@
+﻿@page "/{eventId}/{eventRole}/Teams/{teamId}/InternMeta/{puzzleId}"
```
jenetlan: What does the puzzleId refer to here? Is is supposed to be the puzzleId of the meta?

ServerCore/Pages/Pieces/InternMeta.cshtml
```diff
@@ -0,0 +1,27 @@
...
+@model ServerCore.Pages.Pieces.InternMetaModel
+
+@{
+    ViewData["Title"] = "Intern Puzzleday Meta";
```
jenetlan: Is there any way to actually get to this page? Will we need to add that later?

ServerCore/Pages/Pieces/InternMeta.cshtml
```diff
@@ -0,0 +1,27 @@
...
+}
+
+<h2>Metapuzzle Pieces Earned</h2>
+
+<table class="table">
+    <thead>
+        <tr>
+            <th>
+                @Html.DisplayNameFor(model => model.EarnedPieces[0].Contents)
```
jenetlan: Does this cause the page to crash if there are no pieces?

ServerCore/Pages/Pieces/InternMeta.cshtml.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        public InternMetaModel(PuzzleServerContext serverContext, UserManager<IdentityUser> userManager) : base(serverContext, userManager)
+        {
+        }
+
+        public IList<Piece> EarnedPieces { get;set; }
+        
+        public async Task<IActionResult> OnGetAsync(int puzzleId)
+        {
+            if (!Event.IsInternEvent)
```
morganbr: Rather than limiting to intern events, I'd suggest having auth be that puzzleId be unlocked. That would also make module metas or puzzle hunt metas work.

ServerCore/Pages/Pieces/InternMeta.cshtml
morganbr: As mentioned below, I don't think this has to be limited to intern puzzleday.

ServerCore/Pages/Pieces/InternMeta.cshtml
morganbr: Should this be a more meaningful label to players?

ServerCore/Pages/Pieces/InternMeta.cshtml.cs
tabascq: So it doesn't count pregame puzzles etc. I am using the same "10 points or greater" trick to give us flexibility there.

ServerCore/Pages/Pieces/InternMeta.cshtml.cs
tabascq: I am explicitly limiting to intern events because this approach may be dangerous for nonintern events and I do not want to expose it for general module meta use.

ServerCore/Pages/Pieces/InternMeta.cshtml
tabascq: Not knowing what people want to do in nonintern events I do not want to make a page that can be used to possibly see the innards of those events.

ServerCore/Pages/Pieces/InternMeta.cshtml
tabascq: Open to suggestions

tabascq: No - all our index pages use the same approach and do not explode...

ServerCore/Pages/Pieces/InternMeta.cshtml
tabascq: Since part of the URL is the puzzle ID, I'm not sure how to do this. I was assuming we would put a link to it in a puzzle or make a puzzle with a CustomUrl.

ServerCore/Pages/Pieces/InternMeta.cshtml
tabascq: Yes, this is how the piece data works

ServerCore/Pages/Pieces/InternMeta.cshtml.cs
morganbr: That strikes me as more specific than necessary. Could we reasonably let a puzzle opt into this and perhaps have a way to tell which puzzles feed it? (My last module meta could have used this page, for example)

tabascq: Your last module meta could NOT have used this page because this page is counting solves over the entire event.

ServerCore/Pages/Pieces/InternMeta.cshtml
morganbr: CustomUrl can do it. That said, for puzzleday, we'd probably end up with a URL in the puzzle PDF since we don't use the puzzle links for it otherwise.

ServerCore/Pages/Pieces/InternMeta.cshtml
tabascq: I'll just remove it, there's no need to keep it for a single column table.

ServerCore/Pages/Pieces/InternMeta.cshtml
tabascq: Since we don't post PDFs for interns, we're more likely to make aka.ms/PD2019Meta or something.

tabascq: Now that I've got changes to the DB to denote when a page supports this meta, I've got a link to the pieces directly on the Play page.

ServerCore/Pages/Pieces/InternMeta.cshtml.cs
tabascq: Now supports a event-wide mode that looks at all puzzles and a module mode that looks for matching Group tags. With this change, it's approved for module meta use.

ServerCore/Pages/Pieces/InternMeta.cshtml
tabascq: Thanks for the discussion and the suggestions. This page is now opt-in and has been extended to handle module metas as well.

ServerCore/Pages/Pieces/InternMeta.cshtml.cs
tabascq: Not in the no-frills version but I understand the desire. We might implement general impersonation to address this.

ServerCore/Pages/Pieces/SimpleMeta.cshtml.cs
```diff
@@ -0,0 +1,56 @@
...
+using Microsoft.AspNetCore.Identity;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.EntityFrameworkCore;
+using ServerCore.DataModel;
+using ServerCore.Helpers;
+using ServerCore.ModelBases;
+
+namespace ServerCore.Pages.Pieces
+{
+    [Authorize(Policy = "PlayerIsOnTeam")]
```
morganbr: Should be PlayerCanSeePuzzle

ServerCore/Pages/Pieces/InternMeta.cshtml
tabascq: Agree but I'm not ready to look up how to render markdown from a Razor page yet.

Issue #453, opened 2019-05-16
[P0] Adjust lockout times for DC puzzles


jenetlan:
DC puzzles felt like they had overly-restrictive lockout times. We could adjust this with changes to the overall algorithm or we could look at adding a DC property to puzzles and having a DC-specific algorithm. (note: having the DC property for a puzzle would also let us display a reminder that the puzzle has DC on the submissions page and/or the play page)

tabascq:
Fixed by #486

PR #486, opened 2019-06-13
DC Lockout Fix

tabascq:
Changes the lockout algorithm so that partial responses give you a 1-puzzle credit towards lockout. You can't stack up credits, only cancel incorrect responses in the window.


tabascq:
Fixes #453

Issue #485, opened 2019-06-11
Puzzles without module order should sort alphabetically


morganbr:
The puzzle list defaults to sorting by module and then order in module. However, for puzzleday, they're both the same for all puzzles. They should sort alphabetically after order in module.

tabascq:
Fixed by #487

PR #487, opened 2019-06-13
Sort puzzles by name as a third sort key of other sorts fail

tabascq:
This keeps puzzles from appearing in essentially random order when everything has the same OrderInGroup (e.g. PuzzleDay)


tabascq:
Fixes #485

PR #497, opened 2019-06-20
Optimize submission

morganbr:
Optimizations for answer submission:
* Remove lazy creation of PuzzleStatePerTeam. We were already creating them eagerly, so this is just removing a backup. Includes a button to generate any that may be missing.
* Don't refresh the submissions page after processing a post request -- this avoids redoing most of the database queries
* Fix the submission page making a separate DB call for every submission that has a response

### DIFF
--- a/ServerCore/PuzzleStateHelper.cs
+++ b/ServerCore/PuzzleStateHelper.cs
@@ -38,8 +38,6 @@ namespace ServerCore
                     .Where(state => state.Puzzle == puzzle && state.Team == team);
             }
 
-#pragma warning disable IDE0031 // despite the compiler message, "teamstate?.UnlockedTime", etc does not compile here
-
             if (puzzle != null)
             {
                 return context.PuzzleStatePerTeam.Where(state => state.Puzzle == puzzle);
@@ -51,7 +49,8 @@ namespace ServerCore
                 {
                     return from state in context.PuzzleStatePerTeam
                            join auth in context.PuzzleAuthors on state.PuzzleID equals auth.PuzzleID
-                           where state.Team == team
+                           where state.Team == team &&
+                           auth.Author == author
                            select state;
                 }
                 else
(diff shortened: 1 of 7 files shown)


morganbr:
@tabascq, might you have time to take a look at this so we can test it during beta?

ServerCore/PuzzleStateHelper.cs
```diff
@@ -34,60 +35,29 @@ public static IQueryable<PuzzleStatePerTeam> GetFullReadOnlyQuery(PuzzleServerCo
...
-                    {
-                        Puzzle = puzzle,
-                        PuzzleID = puzzle.ID,
-                        Team = team,
-                        TeamID = team.ID
-                    });
+                    .Where(state => state.Puzzle == puzzle && state.Team == team);
             }
 
 #pragma warning disable IDE0031 // despite the compiler message, "teamstate?.UnlockedTime", etc does not compile here
```
tabascq: can remove this #pragma now

ServerCore/PuzzleStateHelper.cs
```diff
@@ -34,60 +35,29 @@ public static IQueryable<PuzzleStatePerTeam> GetFullReadOnlyQuery(PuzzleServerCo
...
-                           SolvedTime = teamstate == null ? null : teamstate.SolvedTime,
-                           Printed = teamstate == null ? false : teamstate.Printed,
-                           Notes = teamstate == null ? null : teamstate.Notes,
-                           IsEmailOnlyMode = teamstate == null ? false : teamstate.IsEmailOnlyMode
-                       };
+                if (author != null)
+                {
+                    return from state in context.PuzzleStatePerTeam
+                           join auth in context.PuzzleAuthors on state.PuzzleID equals auth.PuzzleID
+                           where state.Team == team
```
tabascq: where does this filter to the author? Since "author" is not present in the query I suspect it doesn't.

morganbr: Good catch, fixed

PR #508, opened 2019-06-22
Sort Map By Group, so metas will move to the left. Also filters out i…

tabascq:
…nvisible puzzles to save space.

### DIFF
--- a/ServerCore/Pages/Events/Map.cshtml.cs
+++ b/ServerCore/Pages/Events/Map.cshtml.cs
@@ -37,7 +37,7 @@ namespace ServerCore.Pages.Events
 
             if (EventRole == EventRole.admin)
             {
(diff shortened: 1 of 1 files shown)


ServerCore/Pages/Events/Map.cshtml.cs
```diff
@@ -37,7 +37,7 @@ public async Task<IActionResult> OnGetAsync(int? refresh)
 
             if (EventRole == EventRole.admin)
             {
-                puzzles = await _context.Puzzles.Where(p => p.Event == Event)
+                puzzles = await _context.Puzzles.Where(p => p.Event == Event && p.IsPuzzle)
```
morganbr: I found this useful during Hunt when I needed to see how timed unlocks were running. Could they just sort to the end?

tabascq: Not a bad idea, will fix

Issue #516, opened 2019-06-24
Truncate numbers on "All Feedback" page


lambertwang:
Minor nit: the averages on this page should be truncated 
maybe show 3.57 instead of 3.57142857142857

jenetlan:
Or round - not sure which is better

Issue #512, opened 2019-06-23
Feedback page clean-up


jenetlan:
On the feedback page:
* Make the average numbers at the top a bit smaller
* Add some indication of that the max value is and which value is good (higher values are more fun and more difficult)

PR #523, opened 2019-06-27
Feedback page bugs and fix role switcher

digit01Wave:
Feedback page bugs: 
#516 
#512 
#499 

Also fixed banner role picker to work for players


ServerCore/Pages/Shared/_RoleSwitcherPartial.cshtml
```diff
@@ -8,19 +8,17 @@
         var IsAdmin = EventRole == ModelBases.EventRole.admin || await EventModel.IsEventAdmin();
         var IsAuthor = EventRole == ModelBases.EventRole.author || await EventModel.IsEventAuthor();
 
-        if (IsAdmin || IsAuthor)
+        var IsRegistered = await EventModel.IsRegisteredUser();
```
tabascq: I'm not sure what this is trying to achieve. If the player is not an author or admin we should not show the switcher because there is nothing to switch to. Is there a specific bug you are trying to fix with this?

Issue #526, opened 2019-06-27
When you have answer submission soft-locked and serve out the time, you shouldn't see a message that makes you think you submitted twice


philipzloh:
When the timer disappears, it is replaced by a string that says "You already submitted X!"
But no, I only submitted X once -- it merely caused the time lock

![image](https://user-images.githubusercontent.com/836760/60298724-f78f3f80-98df-11e9-8491-ec54685c1b8b.png)

Issue #517, opened 2019-06-25
Add email for hints link


morganbr:
For events that do email support for puzzles, there should be links to email puzzle control with a fixed subject that can be filtered into the appropriate folder. We will need this for Puzzleday pregame.

PR #532, opened 2019-06-28
Answer submission page magic

### LINKED ISSUE #528: Answer submission page needs some UI love
Pls gib luv kthxbai

philipzloh:
##Bunch'a fixes

#528 
- submission history not shown if submitted nothing
- a ton of error messages friendlified
- answer submission box autofocused
- lockout timer can no longer go negative

#531 
Fixed (Edit: NOT) by using the POST -> REDIRECT -> GET trick so that when the POST happens, we redirect to a GET call, thereby avoiding the "did you want to resend the submission" scenario.

Edit: THIS HAS BEEN REVERTED because it would cause a 2x increase in page load.

#526 
Fixed by changing 
```
                if (timeDeltaSeconds <= 0) {
                    window.location.reload(false)
                }
```
to
```
                if (timeDeltaSeconds <= 0) {
                    window.location.href = window.location.href;
                }
```

#517 
- mailto link will now exist
- and have the right data in it 
- and BONUS: CC the team in case some random email account was used 
- and BONUS: be different for if you're in Email Mode

#530
image

### DIFF
--- a/ServerCore/Pages/Submissions/Index.cshtml
+++ b/ServerCore/Pages/Submissions/Index.cshtml
@@ -14,9 +14,11 @@
     
     // to: Add the proper support agent as recipient
     mailtoUrl += Model.PuzzleState?.Puzzle.SupportEmailAlias ?? (Model.Event?.ContactEmail ?? "puzzhunt@microsoft.com");
+    // cc: Cc the team's email address because what if this email account is a random personal one?
+    mailtoUrl += "?cc=" + Uri.EscapeDataString(Model.Team.PrimaryContactEmail);
 
     // subject: Make this be about this puzzle
-    mailtoUrl += "?subject=" + Uri.EscapeDataString("[" + Model.Puzzle.Name + "]");
+    mailtoUrl += "&subject=" + Uri.EscapeDataString("[" + Model.Puzzle.Name + "]");
     // subject: Make this be from this team
     mailtoUrl += Uri.EscapeDataString(" [" + Model.Team.Name + "]");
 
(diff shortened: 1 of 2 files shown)


ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -55,32 +55,32 @@ public async Task<IActionResult> OnPostAsync(int puzzleId, string submissionText
             SubmissionText = submissionText;
             if (!this.Event.IsAnswerSubmissionActive)
             {
-                return Page();
+                return RedirectToPage();
```
morganbr: This causes really bad perf on this page because reloading has to redo a bunch of queries.

ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -9,6 +9,37 @@
     ViewData["RoutingPuzzleId"] = ViewContext.RouteData.Values["puzzleId"];
 }
 
+@{
+    var mailtoUrl = "mailto:";
+    
+    // to: Add the proper support agent as recipient
+    mailtoUrl += Model.PuzzleState?.Puzzle.SupportEmailAlias ?? (Model.Event?.ContactEmail ?? "puzzhunt@microsoft.com");
```
morganbr: This is probably a good thing, but we'll need to go through the SupportEmailAlias fields we currently have set so they go to the right place

philipzloh: Yup :) I highlighted this in the tech meeting and the room said this logic is the right one (and we just need to be good about checking that stuff)

ServerCore/Pages/Submissions/Index.cshtml.cs
philipzloh: Ouch :)

Issue #534, opened 2019-07-01
When a new Response is created and we run around marking old submissions with the new response, we should only mark submissions for that puzzle


philipzloh:
![image](https://user-images.githubusercontent.com/836760/60466144-b82a6100-9c07-11e9-9731-22d356149887.png)

PR #536, opened 2019-07-02
Remove T-Shirt from the profile since we use the copy on swag.

tabascq:


ServerCore/PuzzleStateHelper.cs
```diff
@@ -475,6 +475,7 @@ public static async Task UpdateTeamsWhoSentResponse(PuzzleServerContext context,
                 var submissionsThatMatchResponse = await (from PuzzleStatePerTeam pspt in context.PuzzleStatePerTeam
                                                           join Submission sub in context.Submissions on pspt.Team equals sub.Team
                                                           where pspt.PuzzleID == response.PuzzleID &&
+                                                          sub.PuzzleID == response.PuzzleID &&
```
tabascq: This is the fix for Mariia's bug - forgot to unstage it. I'll remove if people are concerned.

morganbr: This is fine

philipzloh:
Fixes #534 by accident :D but welcomed

PR #540, opened 2019-07-04
Mail non-solvers of a puzzle, to support live events.

tabascq:


ServerCore/Pages/Events/Mailer.cshtml.cs
```diff
@@ -42,11 +48,16 @@ public MailerModel(PuzzleServerContext serverContext, UserManager<IdentityUser>
         {
         }
 
-        public async Task<IActionResult> OnGetAsync(MailGroup group, int? teamId)
+        public async Task<IActionResult> OnGetAsync(MailGroup group, int? puzzleId, int? teamId)
```
morganbr: Could this be a different method? As far as I can tell, providing both puzzleId and teamId or combining puzzleId with any MailGroup other than NonSolvers doesn't work.

tabascq: I don't know how

tabascq: Added validation

PR #543, opened 2019-07-09
Allow postgame

morganbr:
Allow submissions after the event is over without affecting the results

### DIFF
--- a/ServerCore/Pages/Submissions/Index.cshtml
+++ b/ServerCore/Pages/Submissions/Index.cshtml
@@ -70,8 +70,15 @@
                 <h4>Correct</h4>
                 <span>Answer: @Model.AnswerToken</span>
             </div>
+
+            @if (Model.PuzzleState.SolvedTime > Model.Event.AnswerSubmissionEnd)
+            { 
+                <div class="alert alert-warning" role="alert">
+                    <span>You solved this after the event ended and it did not count toward your ranking.</span>
+                </div>
+            }
         }
-        else if (@Model.Event.EventBegin > DateTime.UtcNow)
+        else if (DateTime.UtcNow < Model.Event.EventBegin)
         {
             <div class="col-md-12" style="padding-bottom: 15px;">
                 <h3>This event is not yet in session. No submissions will be accepted at this time.</h3>
@@ -100,10 +107,11 @@
         }
         else
         {
-            if (@Model.Event.AnswerSubmissionEnd < DateTime.UtcNow)
+            if (DateTime.UtcNow > @Model.Event.AnswerSubmissionEnd)
             {
-                <div class="alert alert-danger" role="alert">
-                    <h3>The event is over. You may check answers, but they will not count toward your ranking.</h3>
+                <div class="alert alert-warning" role="alert">
+                    <h4>The event is over.</h4>
+                    <span>You may check answers, but they will not count toward your ranking.</span>
                 </div>
             }
 
--- a/ServerCore/Pages/Submissions/Index.cshtml
+++ b/ServerCore/Pages/Submissions/Index.cshtml
@@ -57,8 +57,6 @@
     </div>
     <br />
 }
-<div class="row">
-    <div class="col-md-4">
         @if (Model.Puzzle.HasDataConfirmation)
         {
             <div style="font-style:italic;">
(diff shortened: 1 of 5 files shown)


morganbr:
Nate suggested adding the same banner to the puzzle list

morganbr:
Live comments:
* Use warning color
* Make size match "oops" warning
* Add warning to play page
* Add box to puzzles solved after the event saying it didn't count

ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -73,11 +71,10 @@
                 <span>Answer: @Model.AnswerToken</span>
             </div>
         }
-        else if (!@Model.Event.IsAnswerSubmissionActive)
+        else if (@Model.Event.EventBegin > DateTime.UtcNow)
```
jenetlan: Nit: I think this is more readable the other way - DateTime < EventBegin
That reads closer to how you'd describe it in English:
* Yes: If it's before the start of the event / if now is before the start of the event
* Awkward: If the event starts after now / if the event start is after now

ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -103,30 +100,38 @@
         }
         else
         {
-            if (Model.DuplicateSubmission)
+            if (@Model.Event.AnswerSubmissionEnd < DateTime.UtcNow)
```
jenetlan: Nit: Same here - I think it's more readable as "If it's after the submission deadline / if now is after the submission deadline" which is flipped from how it's written now

ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -103,30 +100,38 @@
...
-
-            <form method="post">
-                <div asp-validation-summary="ModelOnly" class="text-danger"></div>
-                <div class="form-group">
-                    <label class="control-label">Answer</label>
-                    <input autofocus name="submissionText" class="form-control" data-val="true" data-val-required="Your answer cannot be empty" />
-                    <span style="font-size:12px">Submitted answers will automatically be capitalized and stripped of non-alphanumeric characters</span><br />
-                    @Html.ValidationSummary(true)
-                    <span class="text-danger"><span class="field-validation-valid" data-valmsg-for="submissionText" data-valmsg-replace="true"></span></span>
+            <div class="row">
```
jenetlan: Is there a reason this row was added?

ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -53,7 +53,7 @@ public async Task<IActionResult> OnPostAsync(int puzzleId, string submissionText
             }
 
             SubmissionText = submissionText;
-            if (!this.Event.IsAnswerSubmissionActive)
+            if (Event.EventBegin > DateTime.UtcNow)
```
jenetlan: Nit: Same here

ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -53,7 +53,7 @@ public async Task<IActionResult> OnPostAsync(int puzzleId, string submissionText
...
             {
                 return Page();
```
jenetlan: Should we be returning an error here instead? This seems like an error case

ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -126,7 +126,7 @@ public async Task<IActionResult> OnPostAsync(int puzzleId, string submissionText
 
                 AnswerToken = submission.SubmissionText;
             }
-            else if (submission.Response == null)
+            else if (submission.Response == null && Event.IsAnswerSubmissionActive)
             {
                 // We also determine if the puzzle should be set to email-only mode.
                 if (IsPuzzleSubmissionLimitReached(
```
jenetlan: Does this case also cover lockouts? It looks like it might, which is good

ServerCore/Pages/Teams/Play.cshtml
```diff
@@ -7,7 +7,8 @@
     ViewData["AuthorRoute"] = "/Puzzles/Index";
     // Needs route data - ViewData["PlayRoute"] = "/Teams/Play";
 
-    Boolean unsolvedFilter = Model.StateFilter == PlayModel.PuzzleStateFilter.Unsolved;
+    bool unsolvedFilter = Model.StateFilter == PlayModel.PuzzleStateFilter.Unsolved;
+    bool canSubmit = Model.Event.EventBegin < DateTime.UtcNow;
```
jenetlan: Nit: Same here I think

ServerCore/Pages/Submissions/Index.cshtml
morganbr: It's just moving the row that was previously defined on line 60

ServerCore/Pages/Submissions/Index.cshtml.cs
morganbr: Looks like you're right.

PR #548, opened 2019-07-12
Emergency Intern Fixes

tabascq:


ServerCore/Pages/Teams/Edit.cshtml.cs
```diff
@@ -63,6 +63,12 @@ public async Task<IActionResult> OnPostAsync()
                 return Page();
             }
 
+            // replace the room if an intern event
```
philipzloh: Because "replace" means many things, I'd change the comment to say 

```
// keep the room field unchanged if an intern event
```

jenetlan:
Added an issue for actually disabling the field later: https://github.com/PuzzleServer/mainpuzzleserver/issues/549

tabascq:
Jenna: The field *is* disabled. It's just that it's not a hidden field so it gets cleared on edit. And I opted for this approach rather than a hidden field so that it can't be hacked in some way.

tabascq:
Thanks for the review!

Issue #529, opened 2019-06-27
Answer submission page should say "Incorrect" in the danger bar if the submission was incorrect


metapzl:
Today it only says it in the submission history

PR #551, opened 2019-07-13
Fixed bugs

digit01Wave:
Fixed the following bugs:
1. #529  - "Incorrect" in danger bar
2. #521  - disable text box when not allowed the change name
3. Fix error that shows up on "submit" page when team does not have a primary contact
4. Fix Author submission sorting (and took out unnecessary sorting)
5. Fix alphanumeric sorting of team composition page


ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -11,11 +11,14 @@
...
-    
+
     // to: Add the proper support agent as recipient
     mailtoUrl += Model.PuzzleState?.Puzzle.SupportEmailAlias ?? (Model.Event?.ContactEmail ?? "puzzhunt@microsoft.com");
     // cc: Cc the team's email address because what if this email account is a random personal one?
-    mailtoUrl += "?cc=" + Uri.EscapeDataString(Model.Team.PrimaryContactEmail);
+    if (!string.IsNullOrEmpty(Model.Team.PrimaryContactEmail))
+    {
+        mailtoUrl += "?cc=" + Uri.EscapeDataString(Model.Team.PrimaryContactEmail);
+    }
```
philipzloh: Oh this was the bug. This needs an else case that adds the `?` there though because everything after this expects a `?` to exist.

ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -101,18 +104,25 @@
...
-                <h4>Oops</h4>
-                <span>You've already submitted @Model.SubmissionText!</span>
-            </div>
+                <div class="alert alert-danger" role="alert">
+                    <h4>Oops</h4>
+                    <span>You've already submitted "@Model.SubmissionText!"</span>
+                </div>
+            }
+
+            if (!string.IsNullOrEmpty(Model.AnswerAlertMessage))
```
philipzloh: I know I opened that bug about Incorrect, but ideally we'd also put the partial responses here too if they exist.

ServerCore/Pages/Teams/Edit.cshtml
```diff
@@ -7,7 +7,14 @@
     //Needs route data - ViewData["AuthorRoute"] = "/Teams/Status";
     //Needs route data - ViewData["PlayRoute"] = "/Teams/Edit";
     Layout = "_teamLayout.cshtml";
+
+    bool canChangeTeamName = Model.CanChangeTeamName();
 }
+<style>
```
philipzloh: nope

ServerCore/Pages/Teams/Edit.cshtml
```diff
@@ -23,7 +30,14 @@
...
             <div class="form-group">
                 <label asp-for="Team.Name" class="control-label">Team name</label>
-                <input asp-for="Team.Name" class="form-control" />
+                @if (canChangeTeamName)
+                {
+                    <input asp-for="Team.Name" class="form-control" />
+                }
+                else
+                {
+                    <input class="form-control" asp-for="Team.Name" value="@Model.Team.Name" readonly />
```
philipzloh: instead of `readonly` do `disabled`

ServerCore/Pages/Teams/TeamCompositionSummary.cshtml.cs
```diff
@@ -110,22 +110,22 @@ public IActionResult OnPost(SortEnum sortBy, SortDirectionEnum sortDirection)
                 });
         }
 
-        private Func<TeamComposition, string> GetSortSelector()
+        private Func<TeamComposition, object> GetSortSelector()
```
philipzloh: uh are we able to typecheck here? like, Sortable? or something?

or should we do a `ToInt` or `ToNum` or something in the `case`s?

ServerCore/Pages/Teams/Edit.cshtml
digit01Wave: When I make it disabled, the form doesn't get the property when it gets submitted. This is especially a problem because the field will then clear itself when there are validation errors

tabascq:
Can we remove player submissions from this PR and re-review?

digit01Wave:
Not re

digit01Wave:
@tabascq , the fixes for 2 and 5 can be found in this new PR https://github.com/PuzzleServer/mainpuzzleserver/pull/761

1 has been fixed in another merged PR
3 is no longer necessary because someone else disabled the ability for teams to not have a primary contact.
4. I can't seem to reproduce an issue with this one so I ignored it

tabascq:
Closing since #758 and #761 contain all the fixes now.

Vroo:
@digit01Wave I'm not sure specifically what "Fix error that shows up on "submit" page when team does not have a primary contact" is specifically referring to, but I believe it is still possible for teams to not have a primary contact. This happens when a team is created by an admin rather than by a member of the team.

PR #575, opened 2020-03-04
Updated buildings

tabascq:


ServerCore/Pages/Resources/PH21/RulesPartial.cshtml
```diff
@@ -38,7 +38,7 @@
...
         <li>If any travel is required, it will take place on Microsoft property in or near East Campus, West Campus, North Campus, and/or RedWest. In other words, east of 148th Avenue NE, north of NE 24th Street, northwest of Bel-Red Road, and south of NE 60th Street. Some of these zones may have more activity than others. You are strongly encouraged to have a conference room in this general area.<p>
         <li id="OffLimits">
-            Within the area described above, buildings <b>1, 2, 6, 8, 9, 10, 11, 11A, 22, 24, 26, 30, 33, 34, 50, 87, 111, 114, 120, 123, 124, 126, 127, Studio A, Studio B, and Studio C</b> are <b>off limits</b> and should not be entered, other than to go your private office if it is located in one of these buildings.<p>
+            Within the area described above, buildings <b>4, 21, 22, 30, 33, 34, 40, 41, 42, 43, 44, 50, 87, 111, 114, 120, 123, 124, 125, 126, 127, and Studios A through H</b> are <b>off limits</b> and should not be entered, other than to go your private office if it is located in one of these buildings.<p>
```
jenetlan: The e-mail also says that the cafe buildings for 40-44 are also off limits - should we note that here?
The e-mail also mentions the redmond town center but since PH is only on main campus I assume that's why those aren't here

jenetlan: oh lol read the line above it - ignore the redmond town center comment

tabascq: I think we can leave them off since no teams will have rooms there. Thanks for the review!

jaylorch:
Looks good to me, thanks!

PR #577, opened 2020-04-28
Create pd2020 folder and update rules page with a rough draft.

Spartan97:
Creating the pd2020 folder and updated the rules page for our online event. Placeholders for the other two files.

I hope this works... been a while since I made a PR on github.


ServerCore/wwwroot/css/pd2020/styles.css
```diff
@@ -0,0 +1,105 @@
+/* For logo*/
```
tabascq, line 1: Not sure whether this CSS will work or not. The PD2019 one works because of some code inserted into bundleconfig.json. Not sure whether it's necessary to do a similar trick for this version; let me see if I can figure out how to check out this branch and build it...

tabascq: Yeah, the CSS is not working. Puttering...

PR #586, opened 2020-06-20
Remove the team room from the create page

tabascq:


ServerCore/Pages/Teams/Create.cshtml
```diff
@@ -50,11 +50,13 @@ else
             }
             else
             {
+                <!--
```
morganbr, line 53: It'd be nice to have config for this so we don't forget about it next year

tabascq: I don't know how to do it without database changes and I don't want to deploy any database changes during active registration time.

PR #598, opened 2020-08-14
File import improvements

morganbr:
Fixes issues in importing files across events:
1. Moves the file copying to background threads and updates the database as each file copies. This was previously taking long enough that it would cause request timeouts and the entire import would fail. 
2. Doesn't re-randomize urls when copying blobs to the new event. This maintains the relationship between files that were uploaded in the same directory

### DIFF
--- a/ServerCore/FileManager.cs
+++ b/ServerCore/FileManager.cs
@@ -27,7 +27,7 @@ namespace ServerCore
         /// <returns>Url of the file in blob storage</returns>
         public static async Task<Uri> UploadBlobAsync(string fileName, int eventId, Stream contents)
         {
-            CloudBlockBlob blob = await CreateNewBlob(fileName, eventId, true);
+            CloudBlockBlob blob = await CreateNewBlob(fileName, eventId, GetRandomDirectoryName());
 
             await blob.UploadFromStreamAsync(contents);
             return blob.Uri;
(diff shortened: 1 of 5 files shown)


ServerCore/FileManager.cs
```diff
@@ -103,16 +107,24 @@ public static async Task DeleteBlobAsync(Uri fileUri)
         /// Gets a reference to a random directory
         /// </summary>
         /// <param name="eventId">Event the directory should be associated with</param>
-        private static async Task<CloudBlobDirectory> GetRandomDirectoryAsync(int eventId)
+        private static async Task<CloudBlobDirectory> GetRandomDirectoryAsync(int eventId, bool randomize)
```
tabascq, line 110: I think it would be better to not pass a flag here, and instead move the shortcut behavior directly into UploadBlobsAsync. Otherwise, "randomize without randomize" is hard to process.

ServerCore/FileManager.cs
```diff
@@ -136,4 +148,32 @@ private static CloudStorageAccount StorageAccount
...
+        public void CloneInBackground(ContentFile newFile, string fileName, int eventId, Uri sourceUri)
+        {
+            Task<Uri> t = FileManager.CloneBlobAsync(fileName, eventId, sourceUri);
+            t.ContinueWith((finishedTask) =>
+            {
+                newFile.Url = finishedTask.Result;
+                lock (_context)
+                {
+                    _context.ContentFiles.Add(newFile);
+                    _context.SaveChanges();
```
tabascq, line 174: Is it a problem that this makes hundreds of single property edits and commits them? Would it be better to batch up all ContentFile objects and make all the Add calls in one FinishBackgroundProcessing() call?

tabascq: Maybe add all clone tasks to a List<Task> and then gate the final step on some WhenAll() thing.

morganbr: Ordinarily, you'd be right that this is pretty inefficient. However, even with these fixes, the PD 2020 import still got killed before copying 100% of files, so gating saving all of the entries would have meant that no entries would have saved at all. This way, we save whatever progress we do make.

ServerCore/FileManager.cs
morganbr: Good point, fixed.

PR #600, opened 2021-01-06
Changes for new PH21

tabascq:
Creates a new home/FAQ/Rules for the revised PH21 and links back to it from the original.


jaylorch:
Given that it's going to be a virtual event, I suggest dropping the part saying that Microsoft Security needs to be able to contact teams at all times. I presume we're not going to be involving Microsoft Security at all.

Other than that, it looks good to me. But I'll leave it for others to approve since I can't test it now. (I paved the machine I was using for testing the Puzzlehunt website a while ago, when it lost domain connectivity.)

tabascq:
Agreed; good catch, Jay!

tabascq:
Ping! Can someone please sign off on this so we can deploy it etc.?

asyasky (approved):

PR #603, opened 2021-02-28
Upgrade to .NET 5

morganbr:
Upgrade to .NET 5. Reasons for doing so:
* .NET Core 2.1 goes out of support in August
* Newer versions of .NET Core include Blazor, which should allow more interactive pages without reloads (This will help for PH Hints and would also let us make select pages like answer submission faster)
* Newer versions have improved performance

Changes include:
* Project and package updates to 5.0
* Fixes for breaking changes:
  * Entity Framework no longer does automatic local evaluation, so queries that required local evaluation have been refactored. In most cases, this should have either equivalent or better performance
  * AuthorizationFilterContext is gone, replace with passing appropriate objects to the auth helpers
  * Workarounds for EF Core bugs:
    * Some queries that include the name of the type were broken in .NET Core 3.1 (the bug was supposedly fixed in .NET 5, but I haven't verified it). Removed the name of the type from the impacted queries.
    * Navigation properties with the Required attribute set up the wrong kind of foreign key. This would mess with several tables the next time somebody generates a migration. Worked around by adding a ID property to the affected classes (that doesn't have any effect on the DB).

### DIFF
--- a/ServerCore/Pages/Events/FastestSolves.cshtml.cs
+++ b/ServerCore/Pages/Events/FastestSolves.cshtml.cs
@@ -54,10 +54,12 @@ namespace ServerCore.Pages.Events
                                      pspt.SolvedTime <= submissionEnd &&
                                      !pspt.Team.IsDisqualified &&
                                      pspt.Puzzle.Event == Event
-                                     let puzzleToGroup = new { PuzzleID = pspt.Puzzle.ID, PuzzleName = pspt.Puzzle.Name }
+                                     let puzzleToGroup = new { PuzzleID = pspt.Puzzle.ID, PuzzleName = pspt.Puzzle.Name } // Using 'let' to work around EF grouping limitations (https://www.codeproject.com/Questions/5266406/Invalidoperationexception-the-LINQ-expression-for)
                                      group puzzleToGroup by puzzleToGroup.PuzzleID into puzzleGroup
                                      orderby puzzleGroup.Count() descending, puzzleGroup.Max(puzzleGroup => puzzleGroup.PuzzleName) // Using Max(PuzzleName) because only aggregate operators are allowed
                                      select new { PuzzleID = puzzleGroup.Key, PuzzleName = puzzleGroup.Max(puzzleGroup => puzzleGroup.PuzzleName), SolveCount = puzzleGroup.Count() }).ToListAsync();
+
+            // Getting the top 3 requires working around EF limitations translating sorting results within a group to SQL. Workaround from https://github.com/dotnet/efcore/issues/13805
 
             // Filter to solved puzzles
             var psptToQuery = _context.PuzzleStatePerTeam.Where(pspt => pspt.Puzzle.EventID == Event.ID &&
--- a/ServerCore/Pages/Events/FastestSolves.cshtml.cs
+++ b/ServerCore/Pages/Events/FastestSolves.cshtml.cs
@@ -44,50 +44,65 @@ namespace ServerCore.Pages.Events
             names.ForEach(t => teamNameLookup[t.ID] = t.Name);
 
             DateTime submissionEnd = Event.AnswerSubmissionEnd;
-            // get the page data: puzzle, solve count, top three fastest
-            // todo: This pulls all solved puzzles for the event. Is that performance acceptable?
-            var puzzlesData = (await PuzzleStateHelper.GetSparseQuery(_context, this.Event, null, null)
-                .Where(s => (s.SolvedTime != null &&
-                             s.Puzzle.IsPuzzle &&
-                             s.UnlockedTime != null &&
-                             s.SolvedTime <= submissionEnd &&
-                             s.Team.IsDisqualified == false)).ToListAsync())
-                .GroupBy(state => state.Puzzle)
-                .Select(g => new
-                {
-                    Puzzle = g.Key,
-                    SolveCount = g.Count(),
-                    Fastest = g.OrderBy(s => s.SolvedTime - s.UnlockedTime).Take(3).Select(s => new { s.Team.ID, Time = s.SolvedTime - s.UnlockedTime }),
-                    IsSolvedByUserTeam = g.Where(s => s.Team == this.CurrentTeam).Any()
-                })
-                .OrderByDescending(p => p.SolveCount).ThenBy(p => p.Puzzle.Name).ToList();
 
-            var unlockedData = this.CurrentTeam == null ? null : (new HashSet<int>(
+            // Get the page data: puzzle, solve count, top three fastest
+            // Puzzle solve counts
+            var solveCounts = await (from pspt in _context.PuzzleStatePerTeam
+                                     where pspt.SolvedTime != null &&
+                                     pspt.Puzzle.IsPuzzle &&
+                                     pspt.UnlockedTime != null &&
+                                     pspt.SolvedTime <= submissionEnd &&
+                                     !pspt.Team.IsDisqualified &&
+                                     pspt.Puzzle.Event == Event
+                                     let puzzleToGroup = new { PuzzleID = pspt.Puzzle.ID, PuzzleName = pspt.Puzzle.Name }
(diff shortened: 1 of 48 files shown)


ServerCore/Pages/Events/FastestSolves.cshtml.cs
```diff
@@ -45,21 +45,22 @@ public async Task OnGetAsync(SortOrder? sort, PuzzleStateFilter? stateFilter)
...
             // get the page data: puzzle, solve count, top three fastest
-            var puzzlesData = await PuzzleStateHelper.GetSparseQuery(_context, this.Event, null, null)
+            // todo: This pulls all solved puzzles for the event. Is that performance acceptable?
+            var puzzlesData = (await PuzzleStateHelper.GetSparseQuery(_context, this.Event, null, null)
                 .Where(s => (s.SolvedTime != null &&
                              s.Puzzle.IsPuzzle &&
                              s.UnlockedTime != null &&
                              s.SolvedTime <= submissionEnd &&
-                             s.Team.IsDisqualified == false))
+                             s.Team.IsDisqualified == false)).ToListAsync())
```
tabascq, line 54: What are the perf implications of converting to a list in mid-query rather than at the end? Looks hazardous but I don't really understand the optimization engine.

ServerCore/Pages/Puzzles/Checklist.cshtml.cs
```diff
@@ -48,62 +48,48 @@ public async Task<IActionResult> OnGetAsync()
         {
             List<Puzzle> puzzles = await PuzzleHelper.GetPuzzles(_context, Event, LoggedInUser, EventRole);
 
-            Dictionary<int, List<string>> puzzleAuthors = await (from author in _context.PuzzleAuthors
-                                                                        where author.Puzzle.Event == Event
-                                                                        group author by author.PuzzleID into authorList
-                                                                        select new { Puzzle = authorList.Key, Authors = (from a in authorList select a.Author.Name).ToList() }).ToDictionaryAsync(x => x.Puzzle, x => x.Authors);
+            ILookup<int, string> puzzleAuthors = (from author in _context.PuzzleAuthors
+                                                  where author.Puzzle.Event == Event
+                                                  select author).ToLookup(author => author.PuzzleID, author => author.Author.Name);
```
tabascq, line 53: Is it a problem that there's no async here?

ServerCore/Pages/Puzzles/Checklist.cshtml.cs
```diff
@@ -48,62 +48,48 @@ public async Task<IActionResult> OnGetAsync()
...
-                                                                   select new { Puzzle = responseList.Key, Responses = new ResponseData { ResponseCount = responseList.Count(), HasAnswer = (from r in responseList where r.IsSolution select r).Count() > 0 } }).ToDictionaryAsync(x => x.Puzzle, x => x.Responses);
-            Dictionary<int, IEnumerable<int>> puzzleHints = await (from hint in _context.Hints
-                                                      where hint.Puzzle.Event == Event
-                                                      group hint by hint.Puzzle.ID into hints
-                                                      select new { Puzzle = hints.Key, Hints = (from h in hints select h.Cost) }).ToDictionaryAsync(x => x.Puzzle, x => x.Hints);
+            ILookup<int, string> puzzlePrereqs = (from prerequisite in _context.Prerequisites
+                                                  where prerequisite.Puzzle.Event == Event
+                                                  select prerequisite).ToLookup(prerequisite => prerequisite.PuzzleID, prerequisite => prerequisite.Prerequisite.Name);
```
tabascq, line 62: OK to lose async?

ServerCore/Pages/Puzzles/Checklist.cshtml.cs
```diff
@@ -48,62 +48,48 @@ public async Task<IActionResult> OnGetAsync()
...
+
+            HashSet<int> puzzlesWithSolutions = (from response in _context.Responses
+                                        where response.Puzzle.Event == Event && response.IsSolution
+                                        select response.PuzzleID).ToHashSet();
```
tabascq, line 66: OK to lose async?

ServerCore/Pages/Puzzles/Checklist.cshtml.cs
```diff
@@ -48,62 +48,48 @@ public async Task<IActionResult> OnGetAsync()
...
-                puzzleHints.TryGetValue(puzzle.ID, out IEnumerable<int> hints);
+                IEnumerable<string> prereqs = puzzlePrereqs[puzzle.ID];
+                puzzleResponses.TryGetValue(puzzle.ID, out int responses);
+                puzzleHints.TryGetValue(puzzle.ID, out var hints);
 
                 int totalHintCostThisPuzzle = 0;
                 int hintsCountThisPuzzle = 0;
-                if (hints != null)
+                
+                if(hints != null)
```
tabascq, line 89: nit: no need to delete this space :-)

ServerCore/Pages/Puzzles/Checklist.cshtml.cs
```diff
@@ -48,62 +48,48 @@ public async Task<IActionResult> OnGetAsync()
...
+                                        where response.Puzzle.Event == Event && response.IsSolution
+                                        select response.PuzzleID).ToHashSet();
+            Dictionary<int, int> puzzleResponses = await (from response in _context.Responses
+                                                          where response.Puzzle.Event == Event
+                                                          group response by response.PuzzleID into responseList
+                                                          select new { Puzzle = responseList.Key, ResponseCount = responseList.Count() }).ToDictionaryAsync(x => x.Puzzle, x => x.ResponseCount);
+            var puzzleHints = await (from hint in _context.Hints
+                                     where hint.Puzzle.Event == Event
+                                     group hint by hint.Puzzle.ID into hints
+                                     select new { Puzzle = hints.Key, Count = hints.Count(), TotalDiscount = hints.Min(hint => hint.Cost), TotalCost = hints.Sum(hint => Math.Abs(hint.Cost)) }).ToDictionaryAsync(x => x.Puzzle);
```
tabascq, line 74: This looks like a functionality change for TotalCost, but I'd need to test some examples to be sure...

ServerCore/Pages/Events/FastestSolves.cshtml.cs
morganbr: This was probably what was happening before -- EF has never been able to translate this kind of query, so at some point, it would have pulled data locally. That said, this page has also always been slow, so I've made an attempt to actually compute this in the db.

ServerCore/Pages/Puzzles/Checklist.cshtml.cs
morganbr: It's probably not a huge deal on the checklist page -- essentially async frees up thread pool threads while the query is running and we're probably not going to run a bunch of checklists in parallel. Doing an async fetch first before converting to other collections is fine though, switching to that.

ServerCore/Pages/Puzzles/Checklist.cshtml.cs
morganbr: Indeed. Rescued.

ServerCore/Pages/Puzzles/Checklist.cshtml.cs
morganbr: You're probably right and I don't really understand this logic. Switched to a DB query that has the same result as the old one and restored the discount logic.

tabascq (approved):
Thanks! Just checking, do we need to take this right now?

morganbr:
I'm hoping we can -- the Freeform change is based off this branch (although it probably doesn't strictly speaking need it in its current revision). I'm also hoping to build some quality of life improvements on top of this. Checking in now lets us catch any issues in PH betas+PD alpha. Any concerns?

tabascq:
I have concerns with things happening so soon before the beta that something will break and we will not catch it. Can we get all code changes done today so we can bang on the playground event a bit tomorrow?

morganbr:
Yes, we can get this and Freeform in today.

PR #604, opened 2021-03-06
Sort pieces better for author and player

tabascq:
First sort key: number of solves required
Second sort key: alphabetical order


jaylorch:
Given that server load is a critical resource (and limits the number of teams we can accommodate), it seems better to go without this change. That way, the sorting effort can be offloaded to Javascript in the client, where compute resources aren't as critical.

tabascq:
We do all kinds of sorting on the server in the puzzle list, leaderboards, and more, and SimpleMeta.html is designed for modest numbers of pieces (< 50). I get why you (wisely) invested in client-side work for PH19 when you were doing all kinds of background polling etc., but it's overkill here. Given that I'm just trying to improve the predictability of these pieces for a PH21 module, I'd prefer to keep it simple.

morganbr (approved):

PR #605, opened 2021-03-11
Set MayBeAdminOrAuthor on context object

asyasky:
MayBeAdminOrAuthor was being set on the local instance of user, not on the object that was being updated in the context. Moved the set up a couple lines so that it gets update in the DB.


tabascq (approved):

morganbr:
I don't think this ordering change has any effect. Please confirm that it helps.

asyasky:
Morgan says I'm totally wrong about how the EF object refrences work, so I'm pulling this PR & going to run things manually to make sure anyone added as an author recently is unblocked while I do some more digging.

PR #606, opened 2021-03-19
No-op database migration

morganbr:
The .NET 5 tools generate a differently formatted database model snapshot file, so put in a migration that has no effect so we don't have to comb through the meaningless diffs for real migrations.


tabascq (approved):
Ick :-)

PR #607, opened 2021-03-19
Freeform puzzles

morganbr:
Add a concept of "freeform" puzzle answers -- puzzles that allow submitting arbitrary responses (e.g. for open-ended challenges). Authors/admins get a freeform evaluation queue where they can accept or reject submissions. Accepting a submissions marks the puzzle correct. Authors/admins can also optionally provide a response with their evaluation ("You broke rule 3" or "Nice job!") and the results will be emailed to the submitter.


ServerCore/Pages/Submissions/FreeformQueue.cshtml.cs
```diff
@@ -0,0 +1,130 @@
...
+
+        private async Task PopulateUI(int? puzzleId)
+        {
+            if (puzzleId != null)
+            {
+                Puzzle = await _context.Puzzles.Where(m => m.ID == puzzleId).FirstOrDefaultAsync();
+            }
+
+            // todo: think about a way to go back to old submissions and redo them (e.g. to fix a forgotten comment)
+            // todo: should admins be able to see anything they're not an author of?
```
tabascq, line 67: I believe the answer to this question is yes. Admins should be able to see everything, and in a module event should be trusted to be disciplined enough not to cheat. I mean, a player-admin could just as easily look up the answer to the puzzle they need hints for, mark their own status as solved, etc. Not a blocker for tonight if it slows things down.

ServerCore/Pages/Submissions/FreeformQueue.cshtml.cs
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        private async Task PopulateUI(int? puzzleId)
+        {
+            if (puzzleId != null)
+            {
+                Puzzle = await _context.Puzzles.Where(m => m.ID == puzzleId).FirstOrDefaultAsync();
+            }
+
+            // todo: think about a way to go back to old submissions and redo them (e.g. to fix a forgotten comment)
```
tabascq, line 66: This also needs to have an audit trail, by which I only mean a way to check up on a grader and confirm that they are grading as you would expect. Also selection of greatest hits like Safari does; consider adding a "Favorite" checkbox that can be filtered later. Not a blocker for tonight if it slows things down.

ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -102,12 +104,20 @@ public async Task<IActionResult> OnPostAsync(int puzzleId, string submissionText
...
+            if (Puzzle.IsFreeform)
+            {
+                submission.UnformattedSubmissionText = submissionText;
+            }
+            else
+            {
+                submission.SubmissionText = submissionText;
+            }
+
             submission.Response = await _context.Responses.Where(
```
tabascq, line 121: This response means nothing in the freeform case, why look it up?

ServerCore/Pages/Submissions/FreeformQueue.cshtml.cs
```diff
@@ -0,0 +1,130 @@
...
+                                               join PuzzleAuthors author in _context.PuzzleAuthors on sub.PuzzleID equals author.PuzzleID
+                                               where sub.ID == SubmissionID &&
+                                               sub.FreeformAccepted == null &&
+                                               author.Author == LoggedInUser
+                                               select sub).FirstOrDefaultAsync();
+
+                if (submission != null)
+                {
+                    submission.FreeformAccepted = accepted;
+                    submission.FreeformResponse = FreeformResponse;
```
tabascq, line 110: If FreeformResponse is null, this should just be either "Accepted" or "Rejected". Better than blank.

tabascq (approved):
A couple small things plus some advice for the future. Thanks!

ServerCore/Pages/Submissions/FreeformQueue.cshtml.cs
morganbr: Good idea. Done.

ServerCore/Pages/Submissions/Index.cshtml.cs
morganbr: I had the general thought that there could be freeform puzzles that still have a single final answer like Twice from PD2020. Does that sound reasonable? (Also looking at this again, I realized I wasn't formatting freeform answers for this check; fixed that)

ServerCore/Pages/Submissions/FreeformQueue.cshtml.cs
morganbr: Great ideas. I'll start thinking about that for a followup. Filed #608

ServerCore/Pages/Submissions/Index.cshtml.cs
tabascq: With the formatting pass, makes sense.

ServerCore/Pages/Submissions/FreeformQueue.cshtml.cs
morganbr: Sounds reasonable, done.

PR #614, opened 2021-03-21
Small quality of life improvements for freeform puzzles

morganbr:
Small quality of life improvements for freeform puzzles:
* Turn submisisons that are only a URL into a link
* Limit queue length to make sure the page won't get overwhelmed
* Default to not showing freeform submissions on the submissions page and hide them entirely on SubmisisonsWithoutResponses

### DIFF
--- a/ServerCore/Pages/Submissions/FreeformQueue.cshtml.cs
+++ b/ServerCore/Pages/Submissions/FreeformQueue.cshtml.cs
@@ -52,6 +52,8 @@ namespace ServerCore.Pages.Submissions
 
         public List<SubmissionView> Submissions { get; set; }
 
+        public int FullQueueSize { get; set; }
+
         public async Task<IActionResult> OnGetAsync(int? puzzleId)
         {
             if (puzzleId != null)
@@ -93,6 +95,8 @@ namespace ServerCore.Pages.Submissions
             }
 
             Submissions = await submissionQuery.Take(50).ToListAsync();
+
+            FullQueueSize = await submissionQuery.CountAsync();
             
             foreach(SubmissionView submission in Submissions)
             {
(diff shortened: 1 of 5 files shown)


ServerCore/Pages/Submissions/FreeformQueue.cshtml.cs
```diff
@@ -91,7 +92,15 @@ orderby submission.TimeSubmitted
                 submissionQuery = submissionQuery.Where(submissionView => submissionView.PuzzleId == puzzleId);
             }
 
-            Submissions = await submissionQuery.ToListAsync();
+            Submissions = await submissionQuery.Take(50).ToListAsync();
```
tabascq, line 95: This does the 50 oldest, yes? And it does nice things when queue size is under 50, yes?

Would be nice to report the entire queue size, though. Maybe query the count when the take actually gets 50?

tabascq (approved):

ServerCore/Pages/Submissions/FreeformQueue.cshtml.cs
morganbr: Yes, the queue is oldest first (sorted above) and Take is a maximum. Added a query to get the full size.

asyasky (approved):

ServerCore/Pages/Submissions/FreeformQueue.cshtml.cs
```diff
@@ -93,6 +95,8 @@ orderby submission.TimeSubmitted
             }
 
             Submissions = await submissionQuery.Take(50).ToListAsync();
+
+            FullQueueSize = await submissionQuery.CountAsync();
```
tabascq, line 100: faster version:
    FullQueueSize = (Submissions.Count < 50) ? Submissions.Count : await submissionQuery.CountAsync();

PR #618, opened 2021-03-25
Hint reporting update

tabascq:
Two things:
1. Renamed "tickets" to tokens
2. Now showing tokens earned in the standings as well as tokens spent

### DIFF
--- a/ServerCore/Pages/Events/Standings.cshtml
+++ b/ServerCore/Pages/Events/Standings.cshtml
@@ -36,8 +36,13 @@
                     @if (!Model.Event.HideHints)
                     {
                         <th>
+                            <a asp-page="./Standings" asp-route-sort="@(Model.SortForColumnLink(StandingsModel.SortOrder.HintsEarnedAscending, StandingsModel.SortOrder.HintsEarnedDescending))">
+                                Hint Tokens Earned
+                            </a>
+                        </th>
+                        <th>
                             <a asp-page="./Standings" asp-route-sort="@(Model.SortForColumnLink(StandingsModel.SortOrder.HintsUsedAscending, StandingsModel.SortOrder.HintsUsedDescending))">
-                                Hint Tickets Used
+                                Hint Tokens Used
                             </a>
                         </th>
                     }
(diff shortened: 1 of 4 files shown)


philipzloh:
> Renamed "tickets" to tokens

I wonder if the better plan is to call them coins, because that's what they are in code and also because "tokens" are a different thing that do appear in the UI

ServerCore/Pages/Events/Standings.cshtml
```diff
@@ -35,9 +35,14 @@
                     </th>
                     @if (!Model.Event.HideHints)
                     {
+                        <th>
+                            <a asp-page="./Standings" asp-route-sort="@(Model.SortForColumnLink(StandingsModel.SortOrder.HintsEarnedAscending, StandingsModel.SortOrder.HintsEarnedDescending))">
```
philipzloh, line 39: Is the first entry the default sort when you sort by the column? I'd say descending for HintsEarned makes more sense

ServerCore/Pages/Events/Standings.cshtml.cs
```diff
@@ -112,6 +112,12 @@ public async Task OnGetAsync(SortOrder? sort)
                 case SortOrder.ScoreDescending:
                     teamsFinal = teamsFinal.OrderByDescending(ts => ts.Score).ThenByDescending(ts => ts.Rank).ThenByDescending(ts => ts.Team.Name).ToList();
                     break;
+                case SortOrder.HintsEarnedAscending:
+                    teamsFinal = teamsFinal.OrderBy(ts => ts.Team.HintCoinsEarned).ThenBy(ts => ts.Rank).ThenBy(ts => ts.Team.Name).ToList();
```
philipzloh, line 116: OrderBy ascending HintCoinsEarned (weaker to stronger team order)
ThenBy ascending Rank (stronger to weaker team order)

I think this file has a few of these ThenBys that break "conventional logic"

philipzloh (changes requested):

ServerCore/Pages/Events/Standings.cshtml
tabascq: Agree - fixed this and several others.

philipzloh (approved):
"more hints used" would usually map to "weaker team" in my mind, but no biggie

PR #619, opened 2021-03-25
The discount earned from negative hints should only apply to other ne…

tabascq:
…gative hints. This way, you can have a final extraction hint that does not become free when other hints are used. Several authors have assumed that hints already work this way.

### DIFF
--- a/ServerCore/Pages/Puzzles/Checklist.cshtml.cs
+++ b/ServerCore/Pages/Puzzles/Checklist.cshtml.cs
@@ -93,16 +93,15 @@ namespace ServerCore.Pages.Puzzles
                     hintsCountThisPuzzle = hints.Count();
 
                     // positive hints cost what they say they cost.
-                    // negative hints apply discounts to each other, find the max negative hint.
+                    // negative hints apply discounts to each other, find the most negative hint.
                     foreach (int cost in hints)
                     {
                         totalDiscount = Math.Min(totalDiscount, cost);
                         totalHintCostThisPuzzle += Math.Max(0, cost);
                     }
 
-                    // totalDiscount is 0 or negative. Add that cost (flipped to positive)
-                    // as it must be paid in order to reduce the cost of the others.
-                    totalHintCostThisPuzzle += -totalDiscount;
+                    // add the total available discount to the cost.
+                    totalHintCostThisPuzzle += Math.Abs(totalDiscount);
                 }
 
                 PuzzleData.Add(new PuzzleView()
--- a/ServerCore/Pages/Puzzles/Checklist.cshtml.cs
+++ b/ServerCore/Pages/Puzzles/Checklist.cshtml.cs
@@ -91,22 +91,18 @@ namespace ServerCore.Pages.Puzzles
                     int totalDiscount = 0;
 
                     hintsCountThisPuzzle = hints.Count();
+
+                    // positive hints cost what they say they cost.
+                    // negative hints apply discounts to each other, find the max negative hint.
                     foreach (int cost in hints)
                     {
                         totalDiscount = Math.Min(totalDiscount, cost);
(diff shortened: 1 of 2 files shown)


philipzloh (changes requested):
Please update checklist

ServerCore/Pages/Teams/Hints.cshtml.cs
```diff
@@ -63,17 +63,22 @@ private async Task<IList<HintWithState>> GetAllHints(int puzzleID, int teamID)
...
                 {
-                    hint.Discount = discount;
+                    if (allHintsFree)
+                    {
+                        hint.Discount = -hint.BaseCost;
+                    }
+                    else if (hint.Hint.Cost < 0)
+                    {
+                        hint.Discount = discount;
+                    }
```
philipzloh, line 81: missing the `else` which should cause `hint.Discount` to be undefined later

ServerCore/Pages/Puzzles/Checklist.cshtml.cs
```diff
@@ -91,22 +91,18 @@ public async Task<IActionResult> OnGetAsync()
                     int totalDiscount = 0;
 
                     hintsCountThisPuzzle = hints.Count();
+
+                    // positive hints cost what they say they cost.
+                    // negative hints apply discounts to each other, find the max negative hint.
                     foreach (int cost in hints)
                     {
                         totalDiscount = Math.Min(totalDiscount, cost);
+                        totalHintCostThisPuzzle += Math.Max(0, cost);
```
philipzloh, line 100: unassigned variable

ServerCore/Pages/Puzzles/Checklist.cshtml.cs
```diff
@@ -91,22 +91,18 @@ public async Task<IActionResult> OnGetAsync()
                     int totalDiscount = 0;
 
                     hintsCountThisPuzzle = hints.Count();
+
+                    // positive hints cost what they say they cost.
+                    // negative hints apply discounts to each other, find the max negative hint.
```
philipzloh, line 96: `max negative hint` --> brain hurt 🤣

ServerCore/Pages/Puzzles/Checklist.cshtml.cs
```diff
@@ -91,22 +91,18 @@ public async Task<IActionResult> OnGetAsync()
...
-                    totalHintCostThisPuzzle = -totalDiscount;
-
-                    foreach (int cost in hints)
-                    {
-                        // Negative cost hints will be ignored because Max(0,negative) is 0.
-                        // Positive cost hints will only be counted for the cost above the discount.
-                        // (The discount is counted against each other hint.)
-                        totalHintCostThisPuzzle += Math.Max(0, cost + totalDiscount);
-                    }
+                    totalHintCostThisPuzzle += -totalDiscount;
```
philipzloh, line 105: idk if it's more readable or also syntactically safer in case we screw up elsewhere in the future since we don't have tests to use `Math.Abs`

philipzloh (changes requested):
some notes

ServerCore/Pages/Teams/Hints.cshtml.cs
tabascq: Not needed in C# - see [the language reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/classes#field-initialization).

The default value is 0.

ServerCore/Pages/Puzzles/Checklist.cshtml.cs
tabascq: assigned to 0 on line 85

philipzloh (approved):

PR #624, opened 2021-04-20
Improve Import speed

tabascq:
The slowest part of Import turns out to be the file copy, by far. This change removes the file copy responsibilities from the import task entirely; the storage can simply be duplicated in Azure Storage Explorer by right-clicking and using the Clone... command, then providing the folder name that the import step adds to URLs for this event.

### DIFF
--- a/ServerCore/Pages/Events/Import.cshtml.cs
+++ b/ServerCore/Pages/Events/Import.cshtml.cs
@@ -157,7 +157,14 @@ namespace ServerCore.Pages.Events
                         ContentFile newFile = new ContentFile(contentFile);
                         newFile.EventID = Event.ID;
                         newFile.PuzzleID = puzzleCloneMap[contentFile.Puzzle.ID].ID;
-                        _backgroundUploader.CloneInBackground(newFile, contentFile.ShortName, Event.ID, contentFile.Url);
+                        
+                        // new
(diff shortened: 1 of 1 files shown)


ServerCore/Pages/Events/Import.cshtml.cs
```diff
@@ -130,25 +131,17 @@ public async Task<IActionResult> OnPostImportAsync()
...
                             _context.HintStatePerTeam.Add(new HintStatePerTeam() { Hint = destHint, TeamID = team.ID });
                         }
                     }
 
                     // PuzzleStatePerTeam
-                    foreach (Team team in _context.Teams.Where(t => t.Event == Event))
+                    foreach (Team team in destTeams)
                     {
                         int newPuzzleId = puzzleCloneMap[sourcePuzzle.ID].ID;
-                        bool hasPuzzleStatePerTeam = await (from pspt in _context.PuzzleStatePerTeam
```
tabascq, line 143: Note that the puzzle in puzzleCloneMap is always 100% new, see lines 87-91 above. That's why I believe this check can be removed. Also, there's no parallel existence check for hints in the block above...

morganbr (approved):

tabascq:
Closed in favor of #659.

Issue #608, opened 2021-03-19
Freeform history


morganbr:
We should have a way to revisit freeform submissions after they've been evaluated. That could include:
* A way to change decisions
* A record of who made the decision
* Marking favorites for later (e.g. closing slides)

PR #628, opened 2021-04-28
Freeform sharing and fixes

### LINKED ISSUE #622: Freeform text sumissions get canonicalized
Freeform text submissions get converted to all caps and spaces removed, which makes it hard to read them. They should probably be stored in their original form

morganbr:
Add share permission and author/admin favoriting to freeform submissions and an author view of favorite freeform submissions. This supports showing off selected submisisons at closing. Also includes freeform queue autorefresh and  tracking who graded a submission, but doesn't expose it visually yet and a couple of freeform fixes. Fixes #622 and progress toward #608 

In a later change, we may also add a page for players to see sharable submissions after the event is over.


morganbr:
For anyone planning to merge this, note the DB changes need to be deployed

Data/DataModel/Submission.cs
```diff
@@ -73,6 +73,21 @@ public string SubmissionText
         /// </summary>
         public bool? FreeformAccepted { get; set; }
 
+        /// <summary>
```
tabascq, line 76: Most submissions are not freeform. How much of a perf issue do all of these extra columns add?

ServerCore/Pages/Submissions/AuthorIndex.cshtml
```diff
@@ -21,18 +21,26 @@
...
+        <a asp-page="./AuthorIndex" asp-route-puzzleId="@Model.Puzzle?.ID" asp-route-teamId="@Model.Team?.ID" asp-route-hideFreeform="false" asp-route-sort="@Model.Sort">Show Freeform</a> 
     }
     else
     {        
-        <a asp-page="./AuthorIndex" asp-route-puzzleId="@Model.Puzzle?.ID" asp-route-teamId="@Model.Team?.ID" asp-route-hideFreeform="true" asp-route-sort="@Model.Sort">Hide Freeform</a>
+        <a asp-page="./AuthorIndex" asp-route-puzzleId="@Model.Puzzle?.ID" asp-route-teamId="@Model.Team?.ID" asp-route-hideFreeform="true" asp-route-sort="@Model.Sort">Hide Freeform</a> 
+    }
+    @if (!Model.FavoritesOnly || Model.HideFreeform)
+    {
+        <a asp-page="./AuthorIndex" asp-route-puzzleId="@Model.Puzzle?.ID" asp-route-teamId="@Model.Team?.ID" asp-route-hideFreeform="false" asp-route-favoritesOnly="true" asp-route-sort="@Model.Sort">| Freeform Favorites</a>
```
tabascq, line 39: If we're adding favorites, might as well allow favoriting of wrong answers in non-freeform cases?

Data/DataModel/Submission.cs
tabascq, line 76: I will admit to some anxiety about changing the DB this close to the event with no betas in between. How confident are you?

ServerCore/Pages/Submissions/AuthorIndex.cshtml
tabascq: Though any tweaks here can wait if it's the only thing to address (and if it is indeed advisable at all). I'm more concerned about the shape of the data and the risk level in touching the DB at all, just because of the timeframe.

tabascq: BTW, "favoriting of wrong answers" is exactly the workflow I want (as an author) in place of the "secret achievements" discussion we are having. I mark as many as I like, I am not in competition with other puzzle authors to use them, feedback can be immediate, etc.

tabascq: (Not that I think we have runway to do it today, mind you)

Data/DataModel/Submission.cs
morganbr: Not a significant one. It's a few bytes per submission. It's possible we should have had this stuff in a side table, but I didn't realize we'd want more metadata when I started.

ServerCore/Pages/Submissions/AuthorIndex.cshtml
morganbr: Agreed it sounds good. Made the freeform and favorite filtering orthoganal on this page, but didn't try to add a way to favorite wrong answers

morganbr: Not renaming the DB column though to avoid any risk of messing up the migration

tabascq: Also, to fully support favoriting of wrong answers/freeform, the player index should show what has been favorited.

Data/DataModel/Submission.cs
morganbr: Discussed offline and we're taking precautions including DB backup, extra code review and extra testing

asyasky (approved):

Data/DataModel/Submission.cs
```diff
@@ -84,7 +84,7 @@ public string SubmissionText
         public bool AllowFreeformSharing { get; set; }
 
         /// <summary>
-        /// True if this freeform submission was marked as a favorite
+        /// True if this submission was marked as a favorite
         /// </summary>
         public bool FreeformFavorited { get; set; }
```
tabascq, line 89: If it no longer applies only to freeform, will we be able to rename this later?

PR #650, opened 2021-06-27
Update base-styles.css

philipzloh:


ServerCore/wwwroot/css/pd2021/resources/base-styles.css
```diff
@@ -186,9 +186,16 @@ h1.solution-h1::before {
...
 
-.answer {
+.answer-red {
   color: var(--ruby-red);
   text-transform: uppercase;
+  font-weight: bold;
+}
+
+.answer-blue {
+  color: var(--ruby-red);
```
tabascq: probably should be blue

tabascq (approved):

PR #652, opened 2021-07-05
Add team password url parameter

morganbr:
Allow authors to add a team password to a puzzle url so the puzzle can verify that a link is associated with a particular team

### DIFF
--- a/ServerCore/Helpers/PuzzleHelper.cs
+++ b/ServerCore/Helpers/PuzzleHelper.cs
@@ -33,16 +33,16 @@ namespace ServerCore.Helpers
 
         public static string GetFormattedUrl(Puzzle puzzle, int eventId)
         {
-            return GetFormattedUrl(puzzle.CustomURL, puzzle.ID, eventId);
+            return GetFormattedUrl(puzzle.CustomURL, puzzle.ID, eventId, null);
         }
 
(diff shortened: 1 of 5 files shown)


asyasky (approved):

tabascq:
Can you give a usage example of this?

morganbr:
@tabascq the idea is to link to an externally hosted puzzle with the team password autofilled as a url parameter e.g. https://example.com/mypuzzle?teamId={teamPass} which would turn into https://example.com/mypuzzle?teamId=27bbc66c-5604-4c36-904b-9fb8217ebf92. The externally hosted puzzle can call back to (e.g.) https://puzzlehunt.azurewebsites.net/api/puzzleapi/27bbc66c-5604-4c36-904b-9fb8217ebf92 to confirm it's a real team id. This ensures a team can only see one instance of the external site.

Vroo:
This introduces a security bug. Any puzzle author can get access to any team's password. And sending passwords in urls is never a good idea. A better way to do this is to have the puzzle author provide a secret value and then encrypt the team id + the puzzle id with that secret. Then the externally hosted puzzle can decrypt that to get the ids.

morganbr:
@Vroo "password" is a misnomer for what that field is. It's a randomly generated value that is only used for player invite links. The only think an author could do with that field is join the team (if registration is still open and the team isn't full).

PR #655, opened 2021-07-09
Add custom page loader and submit answer api endpoint

asyasky:


ServerCore/Pages/PuzzleApiController.cs
```diff
@@ -38,5 +43,27 @@ public async Task<ActionResult<bool>> GetValidateTeamPasswordAsync(string teamPa
...
         }
+
+        [HttpPost]
+        [Authorize(Policy = "PlayerCanSeePuzzle")]
+        [Route("api/puzzleapi/submitanswer/{eventId}/{puzzleId}")]
+        public async Task<SubmissionResponse> PostSubmitAnswerAsync([FromBody] AnswerSubmission submission, [FromRoute] string eventId, [FromRoute] int puzzleId)
+        {
+            Event currentEvent = await EventHelper.GetEventFromEventId(context, eventId);
+
+            PuzzleUser user = await PuzzleUser.GetPuzzleUserForCurrentUser(context, User, userManager);
```
morganbr, line 54: Is this possible with auth enabled?

ServerCore/Helpers/SubmissionEvaluator.cs
```diff
@@ -0,0 +1,356 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore;
+using ServerCore.DataModel;
+
+namespace ServerCore.Helpers
```
morganbr, line 8: Comments at least at the file and class level would be helpful if the idea is to move to this going forward

ServerCore/Helpers/SubmissionEvaluator.cs
```diff
@@ -0,0 +1,356 @@
...
+        public Submission Submission { get; set; }
+        public Response Response { get; set; }
+        public string SubmitterName { get; set; }
+        public bool IsFreeform { get; set; }
+        public string FreeformReponse { get; set; }
+    }
+
+    public class SubmissionEvaluator
+    {
+        private static PuzzleStatePerTeam puzzleState;
```
morganbr, line 44: Should all of these fields be locals in EvaluateSubmission? They all look like they'd have race conditions as fields since they're statics

ServerCore/Helpers/SubmissionEvaluator.cs
```diff
@@ -0,0 +1,356 @@
...
+        private static IList<Submission> submissions;
+
+        public static List<SubmissionView> submissionViews;
+
+        public static PuzzleUser loggedInUser;
+
+
+        public static async Task<SubmissionResponse> EvaluateSubmission(PuzzleServerContext context, PuzzleUser loggedInUser, Event anEvent, int puzzleId, string submissionText, bool allowFreeformSharing)
+        {
+            thisEvent = anEvent;
```
morganbr, line 65: Should the parameter just be named ```thisEvent```?

ServerCore/Helpers/SubmissionEvaluator.cs
```diff
@@ -0,0 +1,356 @@
...
+            }
+
+            team = await UserEventHelper.GetTeamForPlayer(context, thisEvent, loggedInUser);
+
+            if (loggedInUser == null || team == null)
+            {
+                return new SubmissionResponse() { ResponseCode = SubmissionResponseCode.Unauthorized };
+            }
+
+            puzzle = await context.Puzzles.Where(
```
morganbr, line 116: This is already done on line 69

ServerCore/Helpers/SubmissionEvaluator.cs
```diff
@@ -0,0 +1,356 @@
...
+            {
+                return new SubmissionResponse() { ResponseCode = SubmissionResponseCode.PuzzleLocked };
+            }
+
+            if (DateTime.UtcNow < thisEvent.EventBegin)
+            {
+                return new SubmissionResponse() { ResponseCode = SubmissionResponseCode.Unauthorized };
+            }
+
+            puzzleState = await (PuzzleStateHelper
```
morganbr, line 129: This is already done on line 72

ServerCore/Helpers/SubmissionEvaluator.cs
```diff
@@ -0,0 +1,356 @@
...
+            puzzlesCausingGlobalLockout = await PuzzleStateHelper.PuzzlesCausingGlobalLockout(context, thisEvent, team).ToListAsync();
+
+            if (puzzlesCausingGlobalLockout.Count != 0 && !puzzlesCausingGlobalLockout.Contains(puzzle))
+            {
+                return new SubmissionResponse() { ResponseCode = SubmissionResponseCode.TeamLockedOut };
+            }
+
+            duplicateSubmission = (from sub in submissions
+                                   where sub.SubmissionText == Response.FormatSubmission(submissionText)
+                                   select sub).Any();
```
morganbr, line 156: Isn't there an AnyAsync?

ServerCore/Helpers/SubmissionEvaluator.cs
```diff
@@ -0,0 +1,356 @@
...
+                        thisEvent,
+                        submission.Puzzle,
+                        submission.Team,
+                        true);
+
+                    var authors = await context.PuzzleAuthors.Where((pa) => pa.Puzzle == submission.Puzzle).Select((pa) => pa.Author.Email).ToListAsync();
+
+                    MailHelper.Singleton.SendPlaintextBcc(authors,
+                        $"{thisEvent.Name}: Team {submission.Team.Name} is in email mode for {submission.Puzzle.Name}",
+                        "");
```
morganbr, line 218: Should this return a locked result or is this the last submission permitted before the lock?

ServerCore/Helpers/SubmissionEvaluator.cs
```diff
@@ -0,0 +1,356 @@
...
+                    }
+                }
+            }
+
+            context.Submissions.Add(submission);
+            await context.SaveChangesAsync();
+
+            if (submission.Response != null && submission.Response.IsSolution)
+            {
+                return new SubmissionResponse() { ResponseCode = SubmissionResponseCode.Correct };
```
morganbr, line 245: Shouldn't correct submission still have the response set (e.g. "Correct! The meta clues are in Building 7")

ServerCore/Helpers/SubmissionEvaluator.cs
```diff
@@ -0,0 +1,356 @@
...
+            }
+
+            if (submission.Response != null && !submission.Response.IsSolution)
+            {
+                return new SubmissionResponse() { ResponseCode = SubmissionResponseCode.Partial, CompleteResponse = submission.Response };
+            }
+
+            if (!puzzle.IsFreeform && submission.Response == null && thisEvent.IsAnswerSubmissionActive)
+            {
+                return new SubmissionResponse() { ResponseCode = SubmissionResponseCode.Freeform, FreeformResponse = submission.FreeformResponse };
```
morganbr, line 255: This doesn't look right... it's checking for non-Freeform puzzles with incorrect submissions during answer submission and returning the Freeform code. Also, if this falls through, it returns Unauthorized. Somewhere in here, it should return Incorrect for non-freeform and Freeform for freeform

ServerCore/Helpers/SubmissionEvaluator.cs
```diff
@@ -0,0 +1,356 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore;
+using ServerCore.DataModel;
+
+namespace ServerCore.Helpers
+{
+
```
morganbr, line 10: Lots of bonus blank lines in this file. A quick formatting pass would be nice, but isn't critical

ServerCore/Pages/CustomHtmlPageLoader.cs
```diff
@@ -0,0 +1,103 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Identity;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.EntityFrameworkCore;
+using ServerCore.DataModel;
+using ServerCore.Helpers;
+
+namespace ServerCore.Pages
+{
+    public class CustomHtmlPageLoader : Controller
```
morganbr, line 11: Class comment, please

ServerCore/Pages/CustomHtmlPageLoader.cs
```diff
@@ -0,0 +1,103 @@
...
+
+        public CustomHtmlPageLoader(PuzzleServerContext context, UserManager<IdentityUser> userManager)
+        {
+            this.context = context;
+            this.userManager = userManager;
+        }
+
+        // GET api/custom
+        [HttpGet]
+        [Route("{eventId}/{puzzleId}/api/custom/")]
```
morganbr, line 24: "api" is odd in the route for a page that isn't an api

ServerCore/Pages/CustomHtmlPageLoader.cs
```diff
@@ -0,0 +1,103 @@
...
+        public async Task<IActionResult> Index(string eventId, int puzzleId)
+        {
+            Event currentEvent = await EventHelper.GetEventFromEventId(context, eventId);
```
morganbr, line 27: Can some/all of these checks be replaced by the Authorize attribute with an appropriate policy?

ServerCore/Pages/CustomHtmlPageLoader.cs
```diff
@@ -0,0 +1,103 @@
...
+                fileContents = await wc.DownloadStringTaskAsync(materialUrl);
+            }
+
+            // Return the file contents to the user.
+            return Content(fileContents, "text/html");
+        }
+
+        // POST api/custom/submit
+        [HttpPost]
+        [Route("{eventId}/{puzzleId}/api/custom/submit/")]
```
morganbr, line 87: This is dead code, right?

morganbr (changes requested):

tabascq:
I know I wasn't asked to review this, but I get the mails anyway so I took a glance. I found that I _can't_ review this because I do not understand the use case at all. I don't know:

- How this is used (All puzzles? One puzzle? Some puzzles?)
- What will happen if it has bugs (this is rather a lot of new code to be calling into service < 36 hours before event start)
- Does it replace any existing code paths (increasing risk enormously)
- Why does the code at the end of the review use the team password but also get the logged in user (meaning the team is already known, why do you need the password, highly confused)

If others know the answers to these questions, please carry on w/o me...

tabascq:
Oops, I think my last question (the password one) is completely off base, I was looking at code in two separate methods. Scratch that question please :-)

ServerCore/Helpers/SubmissionEvaluator.cs
asyasky: I don't see anything in the logic that blocks submissions after setting this up, so I think it goes through this time.

ServerCore/Pages/PuzzleApiController.cs
asyasky: In what way?

morganbr: Comment is off by a line. The question is whether user would ever be null given that auth is enabled

asyasky:
> 
> 
> I know I wasn't asked to review this, but I get the mails anyway so I took a glance. I found that I _can't_ review this because I do not understand the use case at all. I don't know:
> 
>     * How this is used (All puzzles? One puzzle? Some puzzles?)
Right now this is one puzzle - the Westing Game, later I'm going to clean it up and refactor so that other puzzles can use it as well.

> 
>     * What will happen if it has bugs (this is rather a lot of new code to be calling into service < 36 hours before event start)
I've tested the common paths in it and it's almost all directly copied from the existing logic - the only new logic here is the api return paths, which will be consumed by my html puzzle file (which I can update if needed). This is also being used for a puzzle that will have continuous coverage throughout the weekend, so bugs will be spotted quickly.

> 
>     * Does it replace any existing code paths (increasing risk enormously)
None, that was an intentional choice - any issues will only affect my puzzle, everything else stays on the existing code path.

> 
>     * Why does the code at the end of the review use the team password but also get the logged in user (meaning the team is already known, why do you need the password, highly confused)
Can you be more specific? There shouldn't be anywhere I'm using the password so that's an error.

> 
> 
> If others know the answers to these questions, please carry on w/o me...

asyasky:
> 
> 
> Oops, I think my last question (the password one) is completely off base, I was looking at code in two separate methods. Scratch that question please :-)

Then scratch my last question for you :)

philipzloh:
> > * How this is used (All puzzles? One puzzle? Some puzzles?)
> > ```
> 
> Right now this is one puzzle - the Westing Game, later I'm going to clean it up and refactor so that other puzzles can use it as well.
> 
> > ```
> > * What will happen if it has bugs (this is rather a lot of new code to be calling into service < 36 hours before event start)
> > ```
> 
> I've tested the common paths in it and it's almost all directly copied from the existing logic - the only new logic here is the api return paths, which will be consumed by my html puzzle file (which I can update if needed). This is also being used for a puzzle that will have continuous coverage throughout the weekend, so bugs will be spotted quickly.

Um, hold on, what does this change actually enable? I thought we tested and ran The Westing Game without issues?

asyasky:
> Um, hold on, what does this change actually enable? I thought we tested and ran The Westing Game without issues?

This will allow players to pick a sub-challenge from a drop down on the puzzle page and submit to that specific puzzle instead of using the five umbrella puzzles. This will make it so that the puzzle list doesn't have to have all of the individual puzzles (like we did for Puzzle Hunt) and the staff don't have to guess which challenge they're trying to submit for (in the beta no one followed the instructions to include the name in the picture - I don't want to spend my weekend sending submissions back for not following that instruction, I've already rejected 50% of the submissions for not follow more important instructions, namely the ones that say "you can't use internal sharepoint because I can't see it"). The puzzle file itself will have a drop down for the challenge and a text box for the submission text, solving both problems.

philipzloh:
> > Um, hold on, what does this change actually enable? I thought we tested and ran The Westing Game without issues?
> 
> This will allow players to pick a sub-challenge from a drop down on the puzzle page and submit to that specific puzzle instead of using the five umbrella puzzles. This will make it so that the puzzle list doesn't have to have all of the individual puzzles (like we did for Puzzle Hunt) and the staff don't have to guess which challenge they're trying to submit for (in the beta no one followed the instructions to include the name in the picture - I don't want to spend my weekend sending submissions back for not following that instruction, I've already rejected 50% of the submissions for not follow more important instructions, namely the ones that say "you can't use internal sharepoint because I can't see it"). The puzzle file itself will have a drop down for the challenge and a text box for the submission text, solving both problems.

I see the value. Can this be deployed somewhere testable now?

asyasky:
> 
> 
> > > Um, hold on, what does this change actually enable? I thought we tested and ran The Westing Game without issues?
> > 
> > 
> > This will allow players to pick a sub-challenge from a drop down on the puzzle page and submit to that specific puzzle instead of using the five umbrella puzzles. This will make it so that the puzzle list doesn't have to have all of the individual puzzles (like we did for Puzzle Hunt) and the staff don't have to guess which challenge they're trying to submit for (in the beta no one followed the instructions to include the name in the picture - I don't want to spend my weekend sending submissions back for not following that instruction, I've already rejected 50% of the submissions for not follow more important instructions, namely the ones that say "you can't use internal sharepoint because I can't see it"). The puzzle file itself will have a drop down for the challenge and a text box for the submission text, solving both problems.
> 
> I see the value. Can this be deployed somewhere testable now?

Yep, I've been working against it locally but I can also put the html file in the sandbox when I get it cleaned up & this is deployed.

philipzloh:
Ideally we can deploy somewhere before merging and deploying to the main site, yea?

ServerCore/Pages/CustomHtmlPageLoader.cs
```diff
@@ -0,0 +1,60 @@
...
+        private readonly PuzzleServerContext context;
+        private readonly UserManager<IdentityUser> userManager;
+
+        public CustomHtmlPageLoader(PuzzleServerContext context, UserManager<IdentityUser> userManager)
+        {
+            this.context = context;
+            this.userManager = userManager;
+        }
+
+        // GET api/custom
```
morganbr, line 26: Comment is no longer accurate

morganbr (approved):

tabascq:
Who is doing the merge/deploy of this? If it's me, I need a clear signal when it is ready. Note that if one of you does the merge/deploy, there have been other changes today so be sure to get those. Either is fine w/me...

ServerCore/Helpers/SubmissionEvaluator.cs
```diff
@@ -0,0 +1,348 @@
...
+                                                              Response = response,
+                                                              SubmitterName = user.Name,
+                                                              FreeformReponse = sub.FreeformResponse,
+                                                              IsFreeform = puzzle.IsFreeform
+                                                          }).ToListAsync();
+
+            List<Submission> submissions = new List<Submission>(submissionViews.Count);
+            foreach (SubmissionView submissionView in submissionViews)
+            {
+                submissions.Add(submissionView.Submission);
```
tabascq, line 93: Why not just get the submissions directly if this is all you want? submissionViews is unused beyond this point.

ServerCore/Helpers/SubmissionEvaluator.cs
```diff
@@ -0,0 +1,348 @@
...
+            }
+
+            List<Puzzle> puzzlesCausingGlobalLockout = await PuzzleStateHelper.PuzzlesCausingGlobalLockout(context, thisEvent, team).ToListAsync();
+
+            // Return early for cases when there's obviously nothing we can do with the submission
```
tabascq, line 98: Perf nit: These return-early cases could be before the submissions query.

asyasky: done

ServerCore/Helpers/SubmissionEvaluator.cs
asyasky: That's something I'll address in post-event clean up, I wasn't confident enough in my sql syntax to rewrite the queries.

PR #659, opened 2022-01-21
Improve Import Perf

tabascq:
Improve the perf of import such that it succeeds by omitting file copy, and provide user instructions for how to complete the job by clining the container in Azure Storage Explorer.

The model for the container display is very kludgy; I tried to do it the "right" way by creating a separate model item within Import.cshtml.cs, but SelectList was crashing internally and not explaining why. Apologies!

### DIFF
--- a/Data/DataModel/Event.cs
+++ b/Data/DataModel/Event.cs
@@ -21,6 +21,14 @@ namespace ServerCore.DataModel
         public string EventID => UrlString ?? ID.ToString();
 
         /// <summary>
+        /// I was unable to put this in a separate ImportEvent model in import.cshtml.cs;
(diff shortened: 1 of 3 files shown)


morganbr:
Do we also need #624?

morganbr:
LGTM, but I seem to have lost approval permissions. I'll look into it.

morganbr (approved):

tabascq:
LOL, I completely forgot about #624 existing at all. I'd been holding off on this until I added instructions as to how to perform the container clone and assumed I had never put it up for review.

I ran this code to perform today's clone and it took maybe 5 seconds without bumping the server, so I don't think the rest is necessary. I'll just close that one.

Thanks for the review!

PR #661, opened 2022-02-09
Users/tabascq/puzzle weights and event text

tabascq:
Some small changes I've been thinking about for a while to increase site flexibility. They aren't really connected, but futzing with the migrations is annoying so I'm batching them together.

Jay: I haven't fully plumbed SyncController for places to change because I don't really know how to test it other than via the SimpleMeta shell I wrote around it (which I did exercise, and the changes work there).


tabascq:
Ping

jaylorch (commented):
Would it make sense to have the piece meta tag filter be a regular expression rather than an exact match?  I agree with your implicit suggestion that SyncController.cs should also use the new piece filtering stuff; we can chat offline about a good way to test that.  Now that we can customize piece filtering, would it make sense to also customize the solve-value requirements, instead of hardcoding them as "must be >= 10 and <50"?

tabascq:
@jaylorch , I like the idea to filter with a regular expression; I'll make this change. I'd like to get this change in by this weekend since there's an event starting up and we should keep the DB nice and stable during signups etc, so I'd like to save the deeper SyncController.cs changes for a separate PR afterwards. Not keen on changing the default solve-value requirements for fear of breaking old events in some unforeseen way, but I'll look into whether I can easily make that behavior change when a filter is specified.

tabascq:
Now, when a filter is specified, it ignores score. Good idea, thanks!

tabascq:
If there is no further activity in the next 24 hours, I'm just going to merge this. If there are new comments before then, I will certainly address them.

asyasky (approved):
I want to do a little research into whether we need to be doing anything to boost security around loading raw html, but since we're doing it anyway no need to hold the change on it. :) This looks good to me.

tabascq:
Thanks! I've been feeling that since only admins/authors can use raw HTML, it's safe enough. I don't love it, though, and would support a markdown-based approach if an appropriate library/etc could be located.

Vroo:
Puzzle files can be html which has the same risks. And markdown has limited
formatting capabilities.

tabascq:
FYI, this is now deployed. Thanks all!

PR #663, opened 2022-02-28
More Bulk Edits

tabascq:
More Bulk Edits

Support bulk creation of hints, and also support deletion of existing responses/hints during bulk creation.

Also, support multi-puzzle bulk creation of hints/responses. Multi-puzzle is largely the same as single-puzzle except that it has an extra column for puzzle names, and it does not send mail to players.

I know there are some chunks of similar code here, but refactoring seemed a little tricky.

### DIFF
--- a/ServerCore/Pages/Hints/CreateBulk.cshtml.cs
+++ b/ServerCore/Pages/Hints/CreateBulk.cshtml.cs
@@ -60,10 +60,10 @@ namespace ServerCore.Pages.Hints
                 _context.Hints.RemoveRange(hintsToRemove);
             }
 
-            StringReader descriptionReader = new StringReader(Description ?? string.Empty);
-            StringReader contentReader = new StringReader(HintContent ?? string.Empty);
-            StringReader costReader = new StringReader(Cost ?? string.Empty);
-            StringReader displayOrderReader = new StringReader(DisplayOrder ?? string.Empty);
+            using StringReader descriptionReader = new StringReader(Description ?? string.Empty);
+            using StringReader contentReader = new StringReader(HintContent ?? string.Empty);
+            using StringReader costReader = new StringReader(Cost ?? string.Empty);
+            using StringReader displayOrderReader = new StringReader(DisplayOrder ?? string.Empty);
 
             while (true)
             {
--- a/ServerCore/Pages/Hints/CreateBulk.cshtml.cs
+++ b/ServerCore/Pages/Hints/CreateBulk.cshtml.cs
@@ -1,0 +1,138 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Identity;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.EntityFrameworkCore;
+using ServerCore.DataModel;
+using ServerCore.ModelBases;
+
+namespace ServerCore.Pages.Hints
+{
+    [Authorize(Policy = "IsEventAdminOrAuthorOfPuzzle")]
+    public class CreateBulkModel : EventSpecificPageModel
+    {
+        public CreateBulkModel(PuzzleServerContext serverContext, UserManager<IdentityUser> userManager) : base(serverContext, userManager)
+        {
+        }
+
+        [BindProperty]
+        public bool DeleteExisting { get; set; }
+
+        [BindProperty]
+        public string Description { get; set; }
+
+        [BindProperty]
+        public string HintContent { get; set; }
+
+        [BindProperty]
+        public string Cost { get; set; }
+
+        [BindProperty]
+        public string DisplayOrder { get; set; }
+
+        public int PuzzleId { get; set; }
+
+        public Puzzle Puzzle { get; set; }
+
+        public async Task<IActionResult> OnGetAsync(int puzzleId)
+        {
+            PuzzleId = puzzleId;
+            Puzzle = await _context.Puzzles.Where(m => m.ID == puzzleId).FirstOrDefaultAsync();
+            return Page();
+        }
+
+        public async Task<IActionResult> OnPostAsync(int puzzleId)
+        {
+            if (!ModelState.IsValid)
+            {
+                return Page();
+            }
+
+            if (DeleteExisting)
+            {
+                Hint[] hintsToRemove = await (from Hint h in _context.Hints
+                                                      where h.PuzzleID == puzzleId
+                                                      select h).ToArrayAsync();
+                _context.Hints.RemoveRange(hintsToRemove);
+            }
+
+            StringReader descriptionReader = new StringReader(Description ?? string.Empty);
+            StringReader contentReader = new StringReader(HintContent ?? string.Empty);
+            StringReader costReader = new StringReader(Cost ?? string.Empty);
+            StringReader displayOrderReader = new StringReader(DisplayOrder ?? string.Empty);
+
+            while (true)
+            {
+                string description = descriptionReader.ReadLine();
+                string content = contentReader.ReadLine();
+                string cost = costReader.ReadLine();
+                string displayOrder = displayOrderReader.ReadLine();
+
+                // TODO probably clearer ways to validate but I honestly do not understand how validation works
+                if (description == null)
+                {
+                    if (content != null)
+                    {
+                        ModelState.AddModelError("HintContent", "Unmatched Content without Description");
+                    }
+                    if (cost != null)
+                    {
+                        ModelState.AddModelError("Cost", "Unmatched Cost without Description");
+                    }
+                    if (displayOrder != null)
+                    {
+                        ModelState.AddModelError("DisplayOrder", "Unmatched DisplayOrder without Description");
+                    }
+
+                    // we're done
+                    break;
+                }
+
+                int costInt;
+                int displayOrderInt;
+
+                if (content == null)
+                {
+                    ModelState.AddModelError("HintContent", $"Content must be present");
+                }
+
+                // int checks also handle the null case
+                if (!int.TryParse(cost, out costInt))
+                {
+                    ModelState.AddModelError("Cost", $"Cost of '{cost}' must be an integer");
+                    break;
+                }
+
+                if (!int.TryParse(displayOrder, out displayOrderInt))
+                {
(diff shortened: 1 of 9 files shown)


tabascq:
Ping!

ServerCore/Pages/Hints/CreateBulk.cshtml.cs
```diff
@@ -0,0 +1,138 @@
...
+
+            if (DeleteExisting)
+            {
+                Hint[] hintsToRemove = await (from Hint h in _context.Hints
+                                                      where h.PuzzleID == puzzleId
+                                                      select h).ToArrayAsync();
+                _context.Hints.RemoveRange(hintsToRemove);
+            }
+
+            StringReader descriptionReader = new StringReader(Description ?? string.Empty);
```
morganbr, line 63: StringReader is disposable. I'm not sure if it needs it in this case, but just to be safe, please add using to it. I think with .NET 5, you can just write

```using StringReader = new StringReader(Description ?? string.Empty);```

File: ServerCore/Pages/Hints/CreateBulkMulti.cshtml [hunk 2]
```diff
@@ -0,0 +1,75 @@
+﻿@page "/{eventId}/{eventRole}/Hints/CreateBulkMulti"
```
morganbr, line 1: What's the use case for multi-puzzle bulk hinting?

ServerCore/Pages/Hints/CreateBulkMulti.cshtml.cs
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        public async Task<IActionResult> OnPostAsync()
+        {
+            if (!ModelState.IsValid)
+            {
+                return Page();
+            }
+
+            StringReader puzzleNameReader = new StringReader(PuzzleName);
```
morganbr, line 53: Does this also need a null check?

ServerCore/Pages/Hints/CreateBulk.cshtml.cs
```diff
@@ -0,0 +1,138 @@
...
+    public class CreateBulkModel : EventSpecificPageModel
+    {
+        public CreateBulkModel(PuzzleServerContext serverContext, UserManager<IdentityUser> userManager) : base(serverContext, userManager)
+        {
+        }
+
+        [BindProperty]
+        public bool DeleteExisting { get; set; }
+
+        [BindProperty]
```
morganbr, line 25: You might be able to add [Required] on these fields to have the user required to enter something and get rid of a couple of checks below

ServerCore/Pages/Hints/CreateBulkMulti.cshtml.cs
```diff
@@ -0,0 +1,174 @@
...
+    public class CreateBulkMultiModel : EventSpecificPageModel
+    {
+        public CreateBulkMultiModel(PuzzleServerContext serverContext, UserManager<IdentityUser> userManager) : base(serverContext, userManager)
+        {
+        }
+
+        [BindProperty]
+        public bool DeleteExisting { get; set; }
+
+        [BindProperty]
```
morganbr, line 26: Same comment on possibly being able to add [Required] to these

ServerCore/Pages/Hints/CreateBulk.cshtml.cs
```diff
@@ -0,0 +1,138 @@
...
+                    Cost = costInt,
+                    DisplayOrder = displayOrderInt
+                };
+
+                _context.Hints.Add(hint);
+            }
+
+            if (!ModelState.IsValid)
+            {
+                return await OnGetAsync(puzzleId);
```
morganbr, line 130: Does this need to return Page() to show the model errors?

ServerCore/Pages/Hints/CreateBulkMulti.cshtml.cs
```diff
@@ -0,0 +1,174 @@
...
+                    Cost = costInt,
+                    DisplayOrder = displayOrderInt
+                };
+
+                _context.Hints.Add(hint);
+            }
+
+            if (!ModelState.IsValid)
+            {
+                return OnGet();
```
morganbr, line 166: Same question on whether this needs to return Page()

ServerCore/Pages/Responses/CreateBulk.cshtml.cs
```diff
@@ -49,6 +52,14 @@ public async Task<IActionResult> OnPostAsync(int puzzleId)
                 return Page();
             }
 
+            if (DeleteExisting)
+            {
+                Response[] responsesToRemove = await (from Response r in _context.Responses
+                                                      where r.PuzzleID == puzzleId
+                                                      select r).ToArrayAsync();
+                _context.Responses.RemoveRange(responsesToRemove);
```
morganbr, line 60: If I recall correctly, removing a response that a team has already submitted breaks a DB constraint. Should this try to fix it or tell the user what's wrong instead of throwing?

ServerCore/Pages/Hints/CreateBulkMulti.cshtml.cs
```diff
@@ -0,0 +1,174 @@
...
+                string content = contentReader.ReadLine();
+                string cost = costReader.ReadLine();
+                string displayOrder = displayOrderReader.ReadLine();
+
+                int puzzleId = -1;
+
+                if (puzzleName != null && !puzzleTitleLookup.TryGetValue(puzzleName, out puzzleId))
+                {
+                    Puzzle puzzle = await (from Puzzle p in _context.Puzzles
+                                           where p.Name == puzzleName && p.EventID == this.Event.ID
```
morganbr, line 74: I'm not sure if there's much we can do, but this will be case and white-space sensitive

ServerCore/Pages/Responses/CreateBulkMulti.cshtml.cs
```diff
@@ -0,0 +1,199 @@
...
+    public class CreateBulkMultiModel : EventSpecificPageModel
+    {
+        public CreateBulkMultiModel(PuzzleServerContext serverContext, UserManager<IdentityUser> userManager) : base(serverContext, userManager)
+        {
+        }
+
+        [BindProperty]
+        public bool DeleteExisting { get; set; }
+
+        [BindProperty]
```
morganbr, line 26: Same comment on [Required]

ServerCore/Pages/Responses/CreateBulkMulti.cshtml.cs
```diff
@@ -0,0 +1,199 @@
...
+                {
+                    ModelState.AddModelError("SubmittedText", "Unmatched Puzzle without Response");
+                    break;
+                }
+
+                isSolution = isSolution == null ? string.Empty : isSolution.ToLower();
+
+                if (puzzleId == -1)
+                {
+                    ModelState.AddModelError("PuzzleName", "Bug in puzzleId lookup");
```
morganbr, line 173: Throwing might be helpful here so we can get a crash dump

ServerCore/Pages/Responses/CreateBulkMulti.cshtml.cs
```diff
@@ -0,0 +1,199 @@
...
+                if (submittedText == null)
+                {
+                    ModelState.AddModelError("SubmittedText", "Unmatched Puzzle without Submission");
+                    break;
+                }
+
+                string submittedTextFormatted = ServerCore.DataModel.Response.FormatSubmission(submittedText);
+
+                // Ensure that the submission text is unique for this puzzle.
+                async Task<bool> isSubmittedTextUnique(string text)
```
morganbr, line 137: I wonder if it would be more efficient to select all responses for the puzzle the first time we see it and add them to submissionsLookup. (A DB constraint would work too, but I assume you're trying to be more user friendly here)

ServerCore/Pages/Responses/CreateBulkMulti.cshtml.cs
```diff
@@ -0,0 +1,199 @@
...
+                    Note = note,
+                    IsSolution = isSolution == "y" || isSolution == "yes" || isSolution == "t" || isSolution == "true" || isSolution == "1"
+                };
+
+                _context.Responses.Add(response);
+            }
+
+            if (!ModelState.IsValid)
+            {
+                return OnGet();
```
morganbr, line 191: Same question about Page()

morganbr (approved):

ServerCore/Pages/Hints/CreateBulkMulti.cshtml
tabascq: PH22 is planning to distribute response/hint editing across fewer people than module hunts have traditionally used, so it's a timesaver there..

ServerCore/Pages/Hints/CreateBulk.cshtml.cs
tabascq: Neat, thanks!

ServerCore/Pages/Hints/CreateBulkMulti.cshtml.cs
tabascq: Yes, oops

ServerCore/Pages/Hints/CreateBulk.cshtml.cs
tabascq: The issue is that these fields are multiline and each line needs to be validated. [Required] isn't strong enough.

ServerCore/Pages/Hints/CreateBulkMulti.cshtml.cs
tabascq: Agree, though it'll generate errors in those cases that the author can respond to.

ServerCore/Pages/Responses/CreateBulkMulti.cshtml.cs
tabascq: The existing non-multi versions return await OnGetAsync(puzzleId) and I was keeping the same pattern as those. If all four bulk pages can return Page() I can fix them all but I'm not super clear on the difference.

ServerCore/Pages/Responses/CreateBulkMulti.cshtml.cs
tabascq: I think you're right about this; I'll change it.

tabascq:
Thanks for the review!

PR #664, opened 2022-02-28
Support PieceMetaTagFilter within SyncController

tabascq:
One extra bit of code that didn't make it into the data model change. I didn't do this initially because I thought the config info needed to go into the request object, but now I think it's just this simple.


jaylorch (commented):
Looks good to me!

PR #669, opened 2022-04-19
Refine Raw HTML support

tabascq:
New algorithm that lets the "Html.Raw(" part be prefixed with any text you desire, to support sorting order. Also, moved all the logic into a static method so it wouldn't be strewn all over.

### DIFF
--- a/ServerCore/Helpers/RawHtmlHelper.cs
+++ b/ServerCore/Helpers/RawHtmlHelper.cs
@@ -1,0 +1,26 @@
+﻿using Microsoft.AspNetCore.Html;
+using Microsoft.AspNetCore.Mvc.Rendering;
+
+namespace ServerCore.Helpers
+{
+    public static class RawHtmlHelper
+    {
+        /// <summary>
+        /// Helper for extracting raw HTML from a string if present.
+        /// The syntax we look for is: {anything}Html.Raw({raw html})
+        /// </summary>
+        /// <param name="html">html to search for raw HTML in</param>
+        public static IHtmlContent GetRawHtmlIfPresent<T>(string html, IHtmlHelper<T> helper)
+        {
+            if(html != null && html.EndsWith(")") && html.Contains("Html.Raw("))
+            { 
+                int start = html.IndexOf("Html.Raw(") + 9;
+                return helper.Raw(html.Substring(start, html.Length - start - 1));
+            }
+            else
+            {
+                return helper.DisplayFor(m => html);
+            }
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


ServerCore/Helpers/RawHtmlHelper.cs
```diff
@@ -0,0 +1,26 @@
...
+namespace ServerCore.Helpers
+{
+    public static class RawHtmlHelper
+    {
+        /// <summary>
+        /// Helper for extracting raw HTML from a string if present.
+        /// The syntax we look for is: {anything}Html.Raw({raw html})
+        /// </summary>
+        /// <param name="html">html to search for raw HTML in</param>
+        public static IHtmlContent GetRawHtmlIfPresent<T>(string html, IHtmlHelper<T> helper)
```
Vroo, line 13: The name seems a bit off to me. `IfPresent` suggests it doesn't do anything if html isn't present. Also as used it's `....Helper...RawHtmlHelper...RawHtml...` which is long and redundant. I'd suggest maybe something like `Helpers.Display.ForTextOrRawHtml` or `Helpers.Display.ForOptionalRawHtml` since the `Display...` functions return `IHtmlHelper` objects.

ServerCore/Helpers/RawHtmlHelper.cs
```diff
@@ -0,0 +1,26 @@
+﻿using Microsoft.AspNetCore.Html;
+using Microsoft.AspNetCore.Mvc.Rendering;
+
+namespace ServerCore.Helpers
+{
+    public static class RawHtmlHelper
+    {
+        /// <summary>
+        /// Helper for extracting raw HTML from a string if present.
+        /// The syntax we look for is: {anything}Html.Raw({raw html})
```
Vroo, line 10: I'd add:
```
if the string does not contain "Html.Raw", it is returned as rendered text.
```

ServerCore/Helpers/RawHtmlHelper.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// </summary>
+        /// <param name="html">html to search for raw HTML in</param>
```
Vroo, line 12: I'd rename this parameter `text` with the explanation:
```
<param name="text">text which might contain raw HTML</param>
```
and add
```
/// <returns>a A new IHtmlContent containing the rendered text or raw HTML</returns>

Vroo (approved):
Thanks Kenny.

ServerCore/Helpers/RawHtmlHelper.cs
tabascq: The name made sense before I refactored all the IHtmlHelper logic into the method, and I forgot to change the method name. I agree it's redundant, I'm just going to call the method "Display".

PR #674, opened 2022-05-10
Speculative fix for CORS error

jbodner09:
Puzzles in the webification repo can't access the font file stored here, so I'm going to see if storing it in the same location as the puzzles will help the access issues for the online puzzles.  Unfortunately, I can't test this since the issue is with one website talking to another, so if this doesn't work, I'll need to try something else.  

For local testing, I also updated the instructions in the webification repo to better describe how to change the link to the local copy to local-file-talking-to-local-file testing purposes, which should avoid the other issue of the local files not being able to access the font file stored online.


ServerCore/wwwroot/css/pd2022/base-styles.css
```diff
@@ -23,7 +23,7 @@
 
 @font-face {
   font-family: "pd2022-font-title";
-  src: url("resources/LTtilt.ttf");
+  src: url("https://puzzleserverteststore.blob.core.windows.net/evt4/fHdElKHm5bEGdqMFPRvZjA==/LTtilt.ttf");
```
tabascq, line 26: evt4??? which is that - unsure what is happening here and how it keeps from breaking stuff. If this is the pdalpha event, this means that everything will break when we clean it for PD2023.

tabascq (approved):

PR #679, opened 2022-05-20
Add swag filter

asyasky:
Adds a filter for swag (tshirt  & lunch) independent from the IsIntern flag.


tabascq (changes requested):
For future cleanliness, I think this change should also update the Event.IsInternEvent checks in:
- EventSpecificPageModel.HasSwag()
- Teams\Index.cshtml

tabascq (approved):
Good catch on the other case as well!

PR #681, opened 2022-05-30
Sync Client Support

tabascq:
Make a custom sync client aware of the event ID and puzzle ID; this keeps us from having to update the client when migrating from event to event.


jaylorch:
Looks good to me, thanks!

jaylorch (approved):

PR #682, opened 2022-05-31
Added link to feedback submission form from answer submission page

jbodner09:


ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -232,6 +232,12 @@
     <p>A pre-populated email form will appear and you should add details to the message body. Remember: the author may only know specifics about this puzzle and not others in the event.</p>
 }
 
+@if (Model.AllowFeedback)
```
tabascq, line 235: If you use Model.Event.AllowFeedback here, then you do not need to add a new property.

jbodner09: Good catch!

tabascq (approved):

PR #684, opened 2022-05-31
Add infrastructure for constructing help links on PD2022 webified puzzles

jbodner09:


ServerCore/wwwroot/js/pd2022/help.js
```diff
@@ -0,0 +1,20 @@
...
+	for (let i = 0; i < authordiv.childNodes.length; i++) {
+		if ((authordiv.childNodes[i].nodeType == 3) && (authordiv.childNodes[i].textContent.trim().length > 0)) {
+			childindex = i;
+		}
+	}
+	
+	let helpspan = document.createElement("span");
+	helpspan.classList.add("help-link");
+	let helplink = document.createElement("a");
+	helplink.href = "mailto:puzzhunt@microsoft.com?subject=%5B" +
```
tabascq, line 13: Since this mail does not identify the player's team, it'll make support a mess if implemented this way. How necessary is this link?

If it's important enough, we may need to do this in a way akin to how SyncController works, where it loads the file and then does string replacements based on info the server knows.

tabascq: Talked with Josh offline. We might evolve this to work like SyncController or we might evolve this to pass the team name in the query string, but this is a start regardless and it only impacts a single event. So I'm going to approve it although it makes me VERY nervous.

tabascq (approved):

PR #688, opened 2022-06-13
Add User Perf Fix

tabascq:
As the number of players has increased, the add-user list has gotten slower and slower. Making two attempts to fix it:

1. Requesting less data from the server (just ID/Name/Email, not all user fields)

2. Simplifying the query since we know that the set of users to skip will always be very small

If the true bottleneck is the amount of HTML transmitted to the client, this won't help, but that's unlikely since the page doesn't progressively render.

I also removed a Tuple<> from the add-team-member list that wasn't doing anything (Item2 was always -1).


ServerCore/Properties/ServiceDependencies/puzzlehunt - Web Deploy/profile.arm.json
```diff
@@ -0,0 +1,113 @@
+{
```
morganbr, line 1: Extra file?

ServerCore/Pages/Teams/AddMember.cshtml.cs
```diff
@@ -34,12 +34,7 @@ public async Task<IActionResult> OnGetAsync(int teamId)
                 return NotFound("Could not find team with ID '" + teamId + "'. Check to make sure you're accessing this page in the context of a team.");
             }
 
-            Users = await (from user in _context.PuzzleUsers
```
morganbr, line 37: What's the query this generates?

ServerCore/Helpers/AddUserHelper.cs
```diff
@@ -0,0 +1,85 @@
...
+                                             orderby teamMember.Member.ID
+                                             select teamMember.Member.ID).ToListAsync();
+
+            return await GetAllUsersExcept(puzzleServerContext, memberUserIDs);
+        }
+
+        private static async Task<List<PuzzleUserView>> GetAllUsersExcept(PuzzleServerContext puzzleServerContext, List<int> userIDsToSkip)
+        {
+            // get all users (many)
+            List<PuzzleUserView> users = await (from user in puzzleServerContext.PuzzleUsers
```
morganbr, line 47: This looks pretty expensive since it pulls all users from all events ever. It's likely more performant to do an antijoin like this https://stackoverflow.com/a/57217382

morganbr:
We can probably use app insights to find out more about why this is slow too

morganbr (changes requested):

ServerCore/Properties/ServiceDependencies/puzzlehunt - Web Deploy/profile.arm.json
tabascq: Shoot - I've been pointedly not adding this file for many many consecutive PRs and finally forgot. Will remove!

ServerCore/Pages/Teams/AddMember.cshtml.cs
tabascq: I don't know how to tell. Can we look at this together?

ServerCore/Helpers/AddUserHelper.cs
tabascq: At this time the user may not be associated with the event at all. All of these lists currently show "all users from all events ever", minus maybe 25 at most. The number of users we are excluding from the list is very small - either the number of admins for the event, number of authors for the event, or number of players on the team. I think all the SQL overhead here is likely to be overkill since SQL has no way of knowing that removing entries from "all users ever" is the exceptional case for these queries.

tabascq: You should be able to see this in the three queries I'm _removing_, which select all users but a very small subset:

```
                Users = await (from user in _context.PuzzleUsers
                               where !((from eventAdmin in _context.EventAdmins
                                       where eventAdmin.Event == Event
                                       where eventAdmin.Admin == user
                                       select eventAdmin).Any())
                               select user).ToListAsync();

                Users = await (from user in _context.PuzzleUsers
                               where !((from eventAuthor in _context.EventAuthors
                                       where eventAuthor.Event == Event
                                       where eventAuthor.Author == user
                                       select eventAuthor).Any())
                               select user).ToListAsync();

                Users = await (from user in _context.PuzzleUsers
                            where !((from teamMember in _context.TeamMembers
                                    where teamMember.Team.Event == Event
                                    where teamMember.Member == user
                                    select teamMember).Any())
                            select new Tuple<PuzzleUser, int>(user, -1)).ToListAsync();
```

tabascq: [Correction: I guess the prior select team member one ignores all members of _all_ teams, not just the selected team, but even so that is only going to be like 10% of the user list and dropping]

ServerCore/Pages/Teams/AddMember.cshtml.cs
morganbr: Generated queries show up in the debugger output when the query runs. Happy to look at this with you as well

tabascq: Current:

```
SELECT [p].[ID], [p].[Email], [p].[EmployeeAlias], [p].[IdentityUserId], [p].[IsGlobalAdmin], [p].[MayBeAdminOrAuthor], [p].[Name], [p].[PhoneNumber], [p].[TShirtSize], [p].[VisibleToOthers]
FROM [PuzzleUsers] AS [p]
WHERE NOT (EXISTS (
    SELECT 1
    FROM [EventAdmins] AS [e]
    INNER JOIN [Events] AS [e0] ON [e].[EventID] = [e0].[ID]
    INNER JOIN [PuzzleUsers] AS [p0] ON [e].[AdminID] = [p0].[ID]
    WHERE ([e0].[ID] = @__entity_equality_Event_0_ID) AND ([p0].[ID] = [p].[ID])))
```

Proposed: (Note: the INNER JOIN is completely unnecessary and easily removed by doing an EventID equality test instead of an Event equality test; I don't know why Entity Framework is doing this)

```
SELECT [e].[AdminID]
FROM [EventAdmins] AS [e]
INNER JOIN [Events] AS [e0] ON [e].[EventID] = [e0].[ID]
WHERE [e0].[ID] = @__entity_equality_thisEvent_0_ID
ORDER BY [e].[AdminID]

SELECT [p].[ID], [p].[Name], [p].[Email]
FROM [PuzzleUsers] AS [p]
ORDER BY [p].[ID]
```

tabascq: Proposed after next commit:

```
SELECT [e].[AdminID]
FROM [EventAdmins] AS [e]
WHERE [e].[EventID] = @__thisEvent_ID_0
ORDER BY [e].[AdminID]

SELECT [p].[ID], [p].[Name], [p].[Email]
FROM [PuzzleUsers] AS [p]
ORDER BY [p].[ID]
```

morganbr:
@morganbr  to review

morganbr (approved):

PR #689, opened 2022-06-15
GetMailInfo method to assist automatic email creation

tabascq:
Note: This doesn't work via CORS despite many attempts to get that to work. Accordingly, it only works when something like CustomHtmlPageLoader is used.

Also includes a change to the FilesController to pass arguments along, which will be necessary should anyone figure out the CORS part.


tabascq:
The many lines of changes to PuzzleApiController is due to a line endings issue. It doesn't look that extensive when whitespace changes are omitted. The only changes I made are to add the new GetMailInfoAsync method and the MailInfo class.

morganbr:
Can you please fix your line endings? It's hard to review and messes up history

tabascq:
I *did* fix line endings (because VS asked me to on opening this file). If you'd rather me leave them in a non-Windowsy state, I can do that. As noted above, the only things I actually changed here are to add the GetMailInfoAsync method and add the MailInfo class.

ServerCore/Pages/PuzzleApiController.cs
```diff
@@ -1,65 +1,97 @@
...
+        [HttpGet]
+        [Route("api/puzzleapi/getmailinfo")]
+        public async Task<ActionResult<MailInfo>> GetMailInfoAsync(string eventId, int puzzleId)
+        {
+            Event currentEvent = await EventHelper.GetEventFromEventId(context, eventId);
+            Team team = await UserEventHelper.GetTeamForCurrentPlayer(context, currentEvent, User, userManager);
+
+            MailInfo mailInfo = new MailInfo();
+
+            mailInfo.PuzzleName = await (from puzzle in context.Puzzles
```
morganbr, line 61: It would be good to check access to the puzzle

ServerCore/Pages/PuzzleApiController.cs
```diff
@@ -1,65 +1,97 @@
...
+                                         where puzzle.ID == puzzleId
+                                         select puzzle.Name).FirstOrDefaultAsync();
+            mailInfo.TeamName = team.Name;
+
+            // replace commas with semicolons for better email support
+            mailInfo.TeamContactEmail = team.PrimaryContactEmail.Replace(',', ';');
```
morganbr, line 67: Does this need a null check? I thought this was optional

ServerCore/Pages/PuzzleApiController.cs
```diff
@@ -1,65 +1,97 @@
...
+        /// Allows external puzzles to validate a team password
+        /// </summary>
+        /// <param name="teamPassword">Potential team password</param>
+        /// <returns>True if the password belongs to a team, false otherwise</returns>
+        [HttpGet]
+        [Route("api/puzzleapi/getmailinfo")]
+        public async Task<ActionResult<MailInfo>> GetMailInfoAsync(string eventId, int puzzleId)
+        {
+            Event currentEvent = await EventHelper.GetEventFromEventId(context, eventId);
+            Team team = await UserEventHelper.GetTeamForCurrentPlayer(context, currentEvent, User, userManager);
```
morganbr, line 57: What happens if the player isn't on a team in these scenarios?
1. Not signed up for this event
2. Author
3. Admin

morganbr (approved):

ServerCore/Pages/PuzzleApiController.cs
tabascq: Added some checks, thanks for the reminder!

ServerCore/Pages/PuzzleApiController.cs
tabascq: Probably, just added one

ServerCore/Pages/PuzzleApiController.cs
tabascq: Now verifies that you're on a team. No special handling of authors/admins, no need for such things.

PR #691, opened 2022-06-19
Create entries for Puzzle Safari 22.

davidggarber:


ServerCore/ServerCore.csproj
```diff
@@ -50,6 +50,7 @@
   <ItemGroup>
     <Folder Include="Areas\Identity\Data\" />
     <Folder Include="Content\" />
+    <Folder Include="wwwroot\images\ps22\" />
```
corob-msft, line 53: I don't think this should be necessary, or at least, I don't see anything similar that was required before.

davidggarber: It was auto-generated by VS when I created the folder for the image. I've removed it, and the image has no trouble.

tabascq (approved):

tabascq:
Merged and deployed!

PR #701, opened 2022-06-30
Notify us on team merge

tabascq:
BCC the event contact email on team merges so we have a record.

### DIFF
--- a/ServerCore/Helpers/MailHelper.cs
+++ b/ServerCore/Helpers/MailHelper.cs
@@ -71,17 +71,17 @@ your email as the contact address for a team, then you also need to remove it on
     /// <summary>
     /// Send plaintext mail to a single person.
     /// </summary>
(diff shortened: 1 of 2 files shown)


ServerCore/Helpers/MailHelper.cs
```diff
@@ -71,17 +71,17 @@ private MailHelper(IConfiguration configuration, bool isDev)
     /// <summary>
     /// Send plaintext mail to a single person.
     /// </summary>
-    public void SendPlaintextOneAddress(string recipient, string subject, string body)
+    public void SendPlaintextOneAddress(string recipient, string subject, string body, string extraBcc = null)
```
morganbr, line 74: Consider IEnumerable in case we want to copy groups on mails (e.g. all authors of a puzzle or admins of the event)

morganbr (approved):

PR #704, opened 2022-07-02
Import admins when events are imported

tabascq:
We almost always want this, so why not make it easy?

### DIFF
--- a/ServerCore/Pages/Events/Import.cshtml.cs
+++ b/ServerCore/Pages/Events/Import.cshtml.cs
@@ -48,6 +48,7 @@ namespace ServerCore.Pages.Events
                 return Forbid();
             }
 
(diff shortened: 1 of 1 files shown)


ServerCore/Pages/Events/Import.cshtml.cs
```diff
@@ -56,6 +57,20 @@ public async Task<IActionResult> OnPostImportAsync()
 
             using (var transaction = _context.Database.BeginTransaction(System.Data.IsolationLevel.Serializable))
             {
+                // Step 0: Make sure all admins exist
+                foreach (var sourceEventAdmin in sourceEventAdmins)
+                {
+                    var destEventAdmin = await _context.EventAdmins.Where((e) => e.Event == Event && e.Admin == sourceEventAdmin.Admin).FirstOrDefaultAsync();
```
morganbr, line 63: It might be better to get the list of destination event admins first to avoid the n^2 queries

morganbr (approved):

PR #705, opened 2022-07-02
Add Puzzle.SupportEmailAlias to GetMailInfo

tabascq:
For easier puzzle support by...me! (and others of course lol)

### DIFF
--- a/ServerCore/Pages/PuzzleApiController.cs
+++ b/ServerCore/Pages/PuzzleApiController.cs
@@ -91,6 +91,7 @@ namespace ServerCore.Pages
             MailInfo mailInfo = new MailInfo();
 
             mailInfo.PuzzleName = puzzle.Name;
+            mailInfo.PuzzleSupportAlias = puzzle.SupportEmailAlias ?? currentEvent.ContactEmail ?? "puzzhunt@microsoft.com";
             mailInfo.TeamName = team.Name;
 
             // replace commas with semicolons for better email support
@@ -121,6 +122,7 @@ namespace ServerCore.Pages
     public class MailInfo
     {
         public string PuzzleName { get; set; }
+        public string PuzzleSupportAlias { get; set; }
         public string TeamName { get; set; }
         public string TeamContactEmail { get; set; }
     }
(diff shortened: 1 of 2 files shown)


ServerCore/Pages/PuzzleApiController.cs
```diff
@@ -91,6 +91,7 @@ public async Task<ActionResult<MailInfo>> GetMailInfoAsync(string eventId, int p
             MailInfo mailInfo = new MailInfo();
 
             mailInfo.PuzzleName = puzzle.Name;
+            mailInfo.PuzzleSupportAlias = puzzle.SupportEmailAlias ?? currentEvent.ContactEmail ?? "puzzhunt@microsoft.com";
```
morganbr, line 94: Are we switching from all mail going to puzzhunt to sending it to atuhors?

tabascq: It's a per-puzzle choice. We did this last year for puzzles supported by people who couldn't get into the puzzhunt account, and in the Discord you can see a discussion about doing it this year as well. This code mirrors line 18 of Submissions/Index.cshtml, so that the new in-puzzle link will work the same as the website-based link.

jbodner09: And also, I see you've got puzzhunt in the support list for your puzzles, so it'll go there anyway because of that.

tabascq: Yes, but that's just our Puzzleday convention. In module-based events, for example, the authors would not put puzzhunt in the list. I know that help.js is PD2022-specific at this point, but this change allows any other event to build a similar experience if they wish.

morganbr: What do we want for puzzleday? To: author; puzzhunt? Just the author seems dangerous

tabascq: The convention we are using is to put puzzhunt as well as the author (if external) as well as anyone who is supporting the author (if external). But that is a policy issue that has nothing to do with the code change...

morganbr (approved):

PR #709, opened 2022-07-06
Email small teams

morganbr:
Add email small teams button to admin teams page and list emails on the build mailer page

### DIFF
--- a/ServerCore/Pages/Events/Mailer.cshtml.cs
+++ b/ServerCore/Pages/Events/Mailer.cshtml.cs
@@ -20,7 +20,8 @@ namespace ServerCore.Pages.Events
         {
             Players,
             PrimaryContacts,
-            NonSolvers
+            NonSolvers,
+            SmallTeams
         }
 
         [BindProperty]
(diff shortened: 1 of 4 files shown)


ServerCore/Pages/Events/Mailer.cshtml.cs
```diff
@@ -129,11 +145,36 @@ public async Task<IActionResult> OnPostAsync()
 
                     addresses = primariesSplit;
                     break;
-            }
 
-            MailHelper.Singleton.SendPlaintextBcc(addresses, MailSubject, MailBody);
+                case MailGroup.SmallTeams:
+                    List<string> smallTeamPlayers = new List<string>();
```
tabascq, line 150: unused?

morganbr: Oops, yes, removed.

tabascq (approved):
I think sending bulk mail based on this output is inadvisable because the mails will go to junk and not be acted on, but this is a great way to harvest the list of people to send mail to.

PR #716, opened 2022-07-08
Organize groups in state map

tabascq:
Allow arbitrary customization of groups in the state map, with a pattern like "Meta|*|Timing" which would put the Meta puzzles first, the Timing puzzles last, and all the other puzzles in the middle sorted by group. Supports arbitrary group names.

### DIFF
--- a/ServerCore/Pages/Events/Map.cshtml.cs
+++ b/ServerCore/Pages/Events/Map.cshtml.cs
@@ -21,15 +21,26 @@ namespace ServerCore.Pages.Events
 
         public StateStats[,] StateMap { get; private set; }
 
+        public string Groups { get; set; }
+
         public MapModel(PuzzleServerContext serverContext,
                         UserManager<IdentityUser> userManager)
             : base(serverContext, userManager) { }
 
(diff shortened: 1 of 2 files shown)


ServerCore/Pages/Events/Map.cshtml.cs
```diff
@@ -21,17 +21,28 @@ public class MapModel : EventSpecificPageModel
...
                 Refresh = refresh;
             }
 
+            Groups = groups ?? "Meta|*";
+
+            Dictionary<string, int> groupSortIndex = new Dictionary<string, int>();
+            string[] groupList = Groups.Split("|");
+            for (int i = 0; i < groupList.Length; i++)
+            {
+                groupSortIndex[groupList[i]] = i;
```
morganbr, line 43: Consider trimming the strings so there isn't whitespace confusion. Maybe also case/space insensitive comparison?

morganbr (approved):

ServerCore/Pages/Events/Map.cshtml.cs
tabascq: Good suggestions, done!

PR #733, opened 2022-08-11
Allow puzzle names to be raw HTML

tabascq:
The puzzle name appears in many more places than this, but this should cover all the player-facing cases as well as a couple of critical author/admin cases.

### DIFF
--- a/ServerCore/Helpers/RawHtmlHelper.cs
+++ b/ServerCore/Helpers/RawHtmlHelper.cs
@@ -15,7 +15,7 @@ namespace ServerCore.Helpers
         /// <returns>IHtmlContent containing either the raw html processed by Html.Raw or the entire string processed by Html.DisplayFor</returns>
         public static IHtmlContent Display<T>(string text, int eventId, IHtmlHelper<T> helper)
         {
-            if(text != null && text.EndsWith(")") && text.Contains("Html.Raw("))
+            if (text != null && text.EndsWith(")") && text.Contains("Html.Raw("))
             {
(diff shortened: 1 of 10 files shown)


Vroo (commented):
This would be nice. Just to check I understand it.

For this line in Puzzles/Edit.cshtml
```
<h3>@Model.Puzzle.Name: Properties <input type="submit" value="Save properties" class="btn btn-default btn-primary" /><a class="btn btn-default btn-link" asp-page="./Delete" asp-route-puzzleId="@Model.Puzzle.ID">Delete this puzzle</a></h3>
```
it will render the puzzle name as html.

When rendering the prereqs it will render the Plaintext name or the full unfiltered name?

This menu:
```
<select class="form-control" asp-for="NewPrerequisiteID" asp-items="@(new SelectList(Model.PotentialPrerequisites, "ID", "Name"))"></select>
```
really should use the Plaintext name and I don't think it does.

I know it's probably not conventional, but adding this to DataModel/Puzzle.cs might be helpful:
```
[NotMapped]
public Uri PlaintextName
{
     get { return RawHtmlHelper.Plaintext(Name, Event.Id); }
}
```
might be very helpful. You could also do that for `HtmlName`.

tabascq:
Agree that more could be done here (and I alluded to as much in my initial comment), but I am traveling all day tomorrow and then you have a beta coming up this weekend. I'm not opposed to doing more later.

Vroo:
Agree. Just couldn't tell for sure it wasn't sticking rendered html in a menu.

Vroo (approved):

PR #737, opened 2022-08-23
Fix Bulk Hints

tabascq:
I wasn't aware that there was a table that needs to have entries added. Oops!

I also missed that the hints page, being player-visible, needs to have Html.Raw support for the puzzle name.


Vroo (approved):
I don't see any issues but since I wouldn't have known this was required either, better to have someone else review it.

Is a change like this necessary for uploading bulk responses? It did work for me but I did see some server errors.

tabascq:
There is no similar table for bulk responses. I can look up any recent server errors you have, though I need timestamps.

Vroo:
I don't remember when it happened but I'll note the time next time I see a server error.

PR #741, opened 2022-09-03
indicate creditable hints with a ★

Vroo:


ServerCore/Pages/Teams/Hints.cshtml
```diff
@@ -68,6 +68,12 @@ You have <b>
         <text>no</text>
     }
 </b> hint coins.
+@if (Model.Hints.Any(h => h.Hint.Cost < 0)) {
+    <text>
+        Hints marked ★ are creditable hints.
+        If you unlock these hints, you will receive credit for the cost of that hint against the cost of other creditable hints.
```
tabascq, line 74: Both lines display along with the line before on a single line of text. For slightly higher understanding of what "these hints" refers to, I suggest either adding a <br/> before line 73, or combining 73/74 and replacing "hints. If" with "hints; if".

Vroo: Took both suggestions although I used a colon instead of a semicolon (since the second sentence is defining a term in the first). Thanks.

tabascq (approved):

PR #742, opened 2022-09-04
improve reporting of duplicate submissions

Vroo:
When a user enters a duplicate submission, rather than just telling them they already submitted it, it will now tell them what the response was. This is helpful in data confirmation puzzles, especially when multiple people are working together.

I made a very small change to Submissions/Index.cshtml.cs but it shows as a large change because GitHub normalized the line endings. The actual changes are on lines 39, 100, 102.

### DIFF
--- a/ServerCore/Pages/Submissions/Index.cshtml.cs
+++ b/ServerCore/Pages/Submissions/Index.cshtml.cs
@@ -97,7 +97,7 @@ namespace ServerCore.Pages.Submissions
             // Soft enforcement of duplicates to give a friendly message in most cases
             DuplicateSubmission = (from sub in Submissions
                                    where sub.SubmissionText == ServerCore.DataModel.Response.FormatSubmission(submissionText)
-                                   select sub).FirstOrDefault(null).?ResponseText;
+                                   select sub).FirstOrDefault(null)?.ResponseText;
 
             if (DuplicateSubmission != null)
             {
--- a/ServerCore/Pages/Submissions/Index.cshtml.cs
+++ b/ServerCore/Pages/Submissions/Index.cshtml.cs
@@ -36,7 +36,7 @@ namespace ServerCore.Pages.Submissions
 
         public IList<Puzzle> PuzzlesCausingGlobalLockout { get; set; }
 
-        public bool DuplicateSubmission { get; set; }
+        public string DuplicateSubmission { get; set; }
 
         [BindProperty]
         public bool AllowFreeformSharing { get; set; }
@@ -97,9 +97,9 @@ namespace ServerCore.Pages.Submissions
             // Soft enforcement of duplicates to give a friendly message in most cases
             DuplicateSubmission = (from sub in Submissions
                                    where sub.SubmissionText == ServerCore.DataModel.Response.FormatSubmission(submissionText)
-                                   select sub).Any();
+                                   select sub).FirstOrDefault(null).?ResponseText;
 
-            if (DuplicateSubmission)
+            if (DuplicateSubmission != null)
             {
                 return Page();
             }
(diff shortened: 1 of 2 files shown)


ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -1,352 +1,352 @@
...
+            // Don't accept posted submissions when a puzzle is causing lockout
+            if (PuzzlesCausingGlobalLockout.Count != 0 && !PuzzlesCausingGlobalLockout.Contains(Puzzle))
+            {
+                return Page();
+            }
+
+            // Soft enforcement of duplicates to give a friendly message in most cases
+            DuplicateSubmission = (from sub in Submissions
+                                   where sub.SubmissionText == ServerCore.DataModel.Response.FormatSubmission(submissionText)
+                                   select sub).FirstOrDefault(null).?ResponseText;
```
tabascq, line 100: Should be `?.`, not `.?`.

[Also, the `null` is not strictly required here, but it's harmless to provide it so no worries either way.]

Vroo: Fixed. I didn't replace `null` because `FirstOrDefault` actually returns `default(T)` if it's unspecified and there are cases where that's not `null`. So good habit to be explicit here.

tabascq: Also, line 100 needs to be
`select sub).FirstOrDefault(null)?.Response.ResponseText;`

When I went to my copy of this file in the master branch to test out the line above, it did not complain about line endings, so I think that's an issue with whatever editor you're using. Do you mind if I make my own version of this PR with proper line endings?

Vroo: Yes, go ahead and replicate the changes in your own repo.

The inconsistent line endings are on lines 80-82 and 118-124. These lines are separated by `\r\n` rather than just `\n`, I did the edits directly on github.com although emacs also sees the extra characters.

tabascq: In working to replicate these changes, I find that they do not work for the case where the submission has no response (i.e. it's not a partial, it's just wrong).

Unless there is some reason this issue is more important for PH22 than it's been for previous events, I think the safest thing to do is to wait on this. This has never cropped up as an issue before.

Vroo: Except for freeform responses, this would just be replacing null with Incorrect. I don't want to break freeform responses although we're not using them in PH22.

The motivation to do this is that beta teams complained that it was tedious to have to search the page for a duplicate response. I don't know how important it is but it seemed easy. But apparently it isn't.

I've actually written a prototype of a new data checker that can run either in ajax mode with a server or independently without a server. You can check it out here. https://vroo.site44.com/demo/checker.html After this hunt, I'd like to see if I could splice it into puzzleserver.

tabascq: The issue is that the Response object is null, which crashes when I simply use Response.ResponseText. This can surely be worked around with some minor effort, but since it is trickier than expected to get the logic right, I am asking whether this is somehow more important than it has been for previous events. Trying to make an informed decision between benefit and risk.

Vroo: It's probably not worth any risk you think is significant. Just because beta teams complain about something doesn't mean we should jump. I thought it was easy and low risk. I will defer to your judgment that it isn't.

digit01Wave: @tabascq , I ended up solving this in my other PR
https://github.com/PuzzleServer/mainpuzzleserver/pull/758

tabascq:
Going to close this in favor of #758 which (once cleaned up) will do this and more.

PR #743, opened 2022-09-06
More Plaintext

tabascq:
Plaintext for the checklist and for response update emails.

### DIFF
--- a/Data/DataModel/Response.cs
+++ b/Data/DataModel/Response.cs
@@ -70,6 +70,21 @@ namespace ServerCore.DataModel
         public string Note { get; set; }
 
         /// <summary>
+        /// The plaintext responseText of the puzzle, which is a subset of the responseText when the responseText is of the form:
+        /// {plaintextResponseText}Html.Raw({htmlResponseText})
+        /// </summary>
+        public string GetPlaintextResponseText(int eventId)
+        {
+            string responseText = this.ResponseText;
+            if (responseText != null && responseText.EndsWith(")") && responseText.Contains("Html.Raw("))
+            {
+                responseText = responseText.Replace("{eventId}", $"{eventId}");
+                responseText = responseText.Substring(0, responseText.IndexOf("Html.Raw(")).TrimEnd();
+            }
+            return string.IsNullOrEmpty(responseText) ? this.ResponseText : responseText;
+        }
+        
+        /// <summary>
         /// Converts the submission to the common spaceless uppercase format used by the website
         /// </summary>
         /// <param name="submission">The submission text</param>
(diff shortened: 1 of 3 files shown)


Data/DataModel/Response.cs
```diff
@@ -69,6 +69,21 @@ public string SubmittedText
...
+        {
+            string responseText = this.ResponseText;
+            if (responseText != null && responseText.EndsWith(")") && responseText.Contains("Html.Raw("))
+            {
+                responseText = responseText.Replace("{eventId}", $"{eventId}");
+                responseText = responseText.Substring(0, responseText.IndexOf("Html.Raw(")).TrimEnd();
+            }
+            return string.IsNullOrEmpty(responseText) ? this.ResponseText : responseText;
+        }
+
```
royleban, line 86: This doesn't seem quite right to me. It seems to me that {eventId} should never be left behind, because the only reason it will be there is if it's in a URL, even if Html.Raw isn't used. Note that this doesn't affect my usage.
```
        public string GetPlaintextResponseText(int eventId)
        {
            string responseText = this.ResponseText;
            if (responseText != null) {
                responseText = responseText.Replace("{eventId}", $"{eventId}");
                if (responseText.EndsWith(")") && responseText.Contains("Html.Raw("))
                {
                    responseText = responseText.Substring(0, responseText.IndexOf("Html.Raw(")).TrimEnd();
                }
            }
            return string.IsNullOrEmpty(responseText) ? this.ResponseText : responseText;
        }
```

royleban (commented):
I would approve but others are more qualified to determine if this new method should be called anywhere else. One minor concern that needn't block this from getting in, but perhaps for later cleanup (and other places may have the same issue).

Data/DataModel/Response.cs
tabascq: This matches the pattern used elsewhere in the code. If it's ever deemed important to get that scenario working, it'd need a change broader than I'd want to make this close to PH22.

royleban:
I'm fine with that. Long run, I think we should never expose {eventId} or
anything similar.

Data/DataModel/Response.cs
Vroo: I agree with both of you. Let's hold off on that more complex change.

It would probably make sense to move GetPlaintextResponseText into ServerCore.Helpers.RawHtmlHelper so that the places that know about `Html.Raw(...)` are all in the same place. (But not now.)

Vroo (approved):

tabascq:
It's actually a lot trickier than I expected to place this code, due to ASP reasons. Probably need to make a set of prerequisite view objects to do better.

Issue #746, opened 2022-09-20
Chrome blocks login cookies when running locally


beckbria:
**Repro Steps:**

1) Run local instance of ServerCore via VS2019
2) Connect to http://localhost:44319/ in Google Chrome
3) Click "Login/Register" and go through the Microsoft Account Flow

**Expected:** Log into site

**Actual:** 

```
System.Exception: Correlation failed.`

Exception: An error was encountered while handling the remote login.
Microsoft.AspNetCore.Authentication.RemoteAuthenticationHandler<TOptions>.HandleRequestAsync()
Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware.Invoke(HttpContext context)
```

**Root Cause:**

This  is caused by [SameSite cookie policy](https://stackoverflow.com/a/64874175) in Chrome. 

**Workaround:**

Testing in Firefox locally currently works.

**Fix:**

Add the following to `Startup.Configure`:

```
            if (env.IsDevelopment())
            {
                app.UseCookiePolicy(new CookiePolicyOptions()
                {
                    MinimumSameSitePolicy = SameSiteMode.Lax
                });
            }
```

PR #747, opened 2022-09-20
Allow SameSite cookie use during local development

beckbria:
This allows cookies to be shared between localhost and the site name in local sessions.  Without this change Chromium-based browsers will reject attempts to sign in with a Microsoft account.

https://github.com/PuzzleServer/mainpuzzleserver/issues/746


tabascq:
@jbodner09 You had this change as well, but I like how this version is development-only. Much safer this way!

tabascq (approved):

PR #749, opened 2022-09-27
Change All/Unsolved Filter from radio buttons to link

### LINKED ISSUE #748: The Unsolved/All filter does not navigate when clicking on the radio button
Repro steps: 

1) Go to either the Puzzles page or the Fastest Solves page
2) Click on the radio button (not the text)

Expected: Filter changes
Actual: No navigation triggered

3) Click on the text for "Unsolved" or "All" (not the radio button)

Observed: navigation and the filter changes

beckbria:
Radio buttons are traditionally used as selectors, not as links.  Their behavior as links is unpredicatable, and in this case only clicking on the text triggered a navigation.  Clicking on the radio buttons did not, which is confusing to the user.  Eliminate the confusion by removing the radio button and showing a navigation link.

This affects the puzzles list and fastest solves pages. A search of the codebase indicated that all other uses of radio buttons were as input controls that did not trigger navigation, so they were left as is.

New UI: 
image

Old UI for comparison:
image

Fixes #748

### DIFF
--- a/ServerCore/Pages/Events/FastestSolves.cshtml
+++ b/ServerCore/Pages/Events/FastestSolves.cshtml
@@ -15,19 +15,10 @@
 <h2>Fastest Solves</h2>
 <br/>
 <div>
-    <a class="radioButton"
-       asp-page="./FastestSolves"
+    <a asp-page="./FastestSolves"
        asp-route-sort="@Model.Sort"
-       asp-route-stateFilter="@FastestSolvesModel.PuzzleStateFilter.All">
-
-        <input id="StateFilter" name="StateFilter" value="All" type="radio" @(unsolvedFilter ? "" : "checked")> All
-    </a>
-    <a class="radioButton"
-       asp-page="./FastestSolves"
-       asp-route-sort="@Model.Sort"
-       asp-route-stateFilter="@FastestSolvesModel.PuzzleStateFilter.Unsolved">
-
-        <input id="StateFilter" name="StateFilter" value="Unsolved" type="radio" @(unsolvedFilter ? "checked" : "")> Unsolved
+       asp-route-stateFilter=@(unsolvedFilter ? FastestSolvesModel.PuzzleStateFilter.All : FastestSolvesModel.PuzzleStateFilter.Unsolved)>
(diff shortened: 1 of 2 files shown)


ServerCore/Pages/Events/FastestSolves.cshtml
```diff
@@ -15,19 +15,10 @@
...
-
-        <input id="StateFilter" name="StateFilter" value="All" type="radio" @(unsolvedFilter ? "" : "checked")> All
-    </a>
-    <a class="radioButton"
-       asp-page="./FastestSolves"
-       asp-route-sort="@Model.Sort"
-       asp-route-stateFilter="@FastestSolvesModel.PuzzleStateFilter.Unsolved">
-
-        <input id="StateFilter" name="StateFilter" value="Unsolved" type="radio" @(unsolvedFilter ? "checked" : "")> Unsolved
+       asp-route-stateFilter=@(unsolvedFilter ? FastestSolvesModel.PuzzleStateFilter.All : FastestSolvesModel.PuzzleStateFilter.Unsolved)>
```
beckbria, line 20: Another option:

I could change this code to something like:

```
static class PuzzleStateFilterMethods { 
    PuzzleStateFilter opposite(this PuzzleStateFilter psf) {
        switch (psf) {
            case All: 
                return Unsolved;
            case Unsolved:
            case default:
                return All;
        }
    }
}

FastestSolvedModel.PuzzleStateFilter oppositeFilter = Model.StateFilter.opposite();

asp-route-stateFilter=@(oppositeFilter)

Show @(nameof(oppositeFilter)) Puzzles
```

...or even add a uiString() extension method instead of just using nameof if we were worried about future compatability.

Advantages: looks cleaner
Disadvantages: if we ever add a third state we'd need to replace all this logic

tabascq (approved):
Wish I'd thought of this approach several events ago - I was focused on how to get radio buttons to work instead of how to remove them :-)

Thanks!

ServerCore/Pages/Events/FastestSolves.cshtml
tabascq: Nah no need

PR #750, opened 2022-09-29
Allow score adjustment by event admins

tabascq:
Allows admins to create invisible puzzles with positive or negative scores, so team scores can be adjusted without affecting puzzle counts.

### DIFF
--- a/ServerCore/Pages/Events/Standings.cshtml.cs
+++ b/ServerCore/Pages/Events/Standings.cshtml.cs
@@ -28,8 +28,8 @@ namespace ServerCore.Pages.Events
             Sort = sort;
 
             var puzzleData = await _context.Puzzles
-                .Where(p => p.Event == Event && p.IsPuzzle)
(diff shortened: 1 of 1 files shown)


royleban (approved):
Look solid.

Vroo (approved):
This does what it says it does although this condition 
```
.Where(pspt => pspt.SolvedTime != null && pspt.SolvedTime <= submissionEnd)
```
means that adjustments can only be done before the event ends (or by doing retroactive changes).

tabascq:
I'm fine with needing retroactive changes right now since there's a risk that if we modify the event cutoff, something else will go wrong given the fact that players can solve puzzles after the event is over.

royleban:
I agree with Kenny. But, it might be nice to create a future feature that allows for changing a puzzle's SolvedTime to 1 second before the event ended (or perhaps a specified time) to avoid having to write custom SQL to make retroactive changes.

tabascq:
FYI, this is deployed.

PR #758, opened 2023-01-31
Fixed submission answers to show alerts when partial or incorrect

digit01Wave:
Currently, when a user submits a partial or an incorrect answer, they have to scroll down to see what happened.

Now, if the player submits an incorrect answer, a red alert box will pop up to say "Incorrect".
image

If the player submits a partially correct answer, a yellow alert box will pop up to give them the partial.
image

image

### DIFF
--- a/ServerCore/Pages/Submissions/Index.cshtml.cs
+++ b/ServerCore/Pages/Submissions/Index.cshtml.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
 using System.Linq;
+using System.Net.NetworkInformation;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Identity;
@@ -36,7 +37,9 @@ namespace ServerCore.Pages.Submissions
 
         public IList<Puzzle> PuzzlesCausingGlobalLockout { get; set; }
 
-        public bool DuplicateSubmission { get; set; }
+        public string AnswerRedAlertMessage { get; set; }
+
+        public string AnswerYellowAlertMessage { get; set; }
 
         [BindProperty]
         public bool AllowFreeformSharing { get; set; }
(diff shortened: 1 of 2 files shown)


ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -131,57 +140,29 @@ public async Task<IActionResult> OnPostAsync(int puzzleId, string submissionText
...
-                    Event,
-                    submission.Puzzle,
-                    submission.Team,
-                    submission.TimeSubmitted);
-
-                AnswerToken = submission.SubmissionText;
+            if (submission.Response == null)
+            {
+                // Is incorrect
+                AnswerRedAlertMessage = $"\"{submission.SubmissionText}\" is incorrect";
```
tabascq, line 146: This will mark all freeform submissions as incorrect, won't it?

tabascq (changes requested):

ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -95,12 +102,14 @@ public async Task<IActionResult> OnPostAsync(int puzzleId, string submissionText
...
-            DuplicateSubmission = (from sub in Submissions
+            Submission duplicatedSubmission = (from sub in Submissions
                                    where sub.SubmissionText == ServerCore.DataModel.Response.FormatSubmission(submissionText)
-                                   select sub).Any();
+                                   select sub).FirstOrDefault();
 
-            if (DuplicateSubmission)
+            if (duplicatedSubmission != null)
             {
+                string response = duplicatedSubmission.Response != null ? duplicatedSubmission.Response.ResponseText : IndexModel.IncorrectResponseText;
```
tabascq, line 111: This also doesn't handle freeform correctly; you should probably just have a message like "you already submitted this" if the puzzle is freeform.

Beyond that, this is a shorter way to write the non-freeform parts:

string response = duplicatedSubmission.Response?.ResponseText ?? IndexModel.IncorrectResponseText;

That will use IndexModel.IncorrectResponseText if either duplicatedSubmission.Response is null or duplicatedSubmission.Response.ResponseText is null.

digit01Wave: Fixed

PR #759, opened 2023-02-01
do not send mails to no one, take 2

tabascq:
Alternate version of #426

### DIFF
--- a/ServerCore/Helpers/MailHelper.cs
+++ b/ServerCore/Helpers/MailHelper.cs
@@ -182,14 +182,24 @@ your email as the contact address for a team, then you also need to remove it on
             AddRecipients(msg.Bcc, extraBcc);
         }
 
-        SmtpClient client = new SmtpClient("in.mailjet.com", 587);
-        // Unlike port 25 which is for general SMTP, port 587 is for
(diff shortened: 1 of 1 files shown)


ServerCore/Helpers/MailHelper.cs
```diff
@@ -182,14 +182,24 @@ private void SendInternal2(IEnumerable<string> recipients, bool bcc, string subj
...
+            // MSA (Message Submission Agent) use and is authenticated.
+            client.DeliveryMethod = SmtpDeliveryMethod.Network;
+            client.EnableSsl = true;
+            client.UseDefaultCredentials = false;
+            client.Credentials = new NetworkCredential(PublicSecret, PrivateSecret);
+            client.SendMailAsync(msg);
+        }
+        else
+        {
+            if (!IsDev)
```
Vroo, line 198: Why are you writing a debug log message only in production? I would think you would do the reverse.

Vroo (approved):

ServerCore/Helpers/MailHelper.cs
tabascq: I copied and pasted this code from #426 but I can reverse it.

PR #771, opened 2023-02-18
Add page to experiment with iframe embedding of puzzles directly on the website

jbodner09:
Route to test page is {event}/{role}/Submissions/SubmitProto.  Note that the allow-same-origin permission is required so that we pass the same origin test when we're running on the server, since it has the same origin policy turned on.  The allow-scripts permission is self-explanatory since puzzles use scripts for interactivity.  The credentialless attribute is a brand new feature in literally THE latest version of Chrome and Edge that lifts some of the cross-origin restrictions on loading content AND improves security by restricting access the iframe content has to content from the surrounding page:  https://developer.mozilla.org/en-US/docs/Web/Security/IFrame_credentialless.

### DIFF
--- a/PuzzleServer.sln
+++ b/PuzzleServer.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.30907.101
+# Visual Studio Version 17
+VisualStudioVersion = 17.4.33403.182
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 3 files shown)


tabascq:
Are you saying that this prototype only works if you have the absolute latest Chrome/Edge? What do you expect to see if you don't have that?

jbodner09:
Excellent question.  credentialless, when seen on a browser like Firefox that doesn't support it, will merely do nothing.  It is only there to add EXTRA security.  When I use Firefox in my local environment, everything still works the same:  
image

tabascq (approved):

PR #772, opened 2023-02-21
Add prototype list of puzzles that links to prototype submission page instead of regular submission page

jbodner09:
Minimum POC.  After proving this works, these separate files will be removed and the necessary changes will be ported to the existing puzzle lists and submission pages.  The main test that needs to happen on the server is for PDF files, since I couldn't figure out how to upload PDFs on my local test instance.  

For PDFs, it may also be useful to have a download link below the frame in case the browser has trouble rendering the PDF and they want to use a different PDF viewer.  For puzzles that have no content, the message saying so also won't be necessary in production.  

The real changes will also create a new event property, which will be disabled by default, to use the embedded puzzle format, since not all events may want it.

### DIFF
--- a/ServerCore/Pages/Teams/PlayProto.cshtml
+++ b/ServerCore/Pages/Teams/PlayProto.cshtml
@@ -1,0 +1,31 @@
+﻿@page "/{eventId}/{eventRole}/Teams/{teamId}/PlayProto"
+@model ServerCore.Pages.Teams.PlayProtoModel
+
+@{
+    @using Helpers;
+
+    ViewData["Title"] = "Puzzles";
+    ViewData["AdminRoute"] = "/Puzzles/Index";
+    ViewData["AuthorRoute"] = "/Puzzles/Index";
+}
+
+<h2>Puzzles</h2>
+<table class="table">
+    <thead>
+        <tr>
+            <th>
+                <div>Puzzle</div>
+            </th>
+        </tr>
+    </thead>
+    <tbody>
+        @foreach (var puz in Model.PuzzleViews)
+        {
+            <tr>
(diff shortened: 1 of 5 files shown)


ServerCore/Pages/Teams/PlayProto.cshtml
```diff
@@ -0,0 +1,31 @@
...
+                <div>Puzzle</div>
+            </th>
+        </tr>
+    </thead>
+    <tbody>
+        @foreach (var puz in Model.PuzzleViews)
+        {
+            <tr>
+                <td>
+                    <a asp-page="/Submissions/SubmitProto" asp-route-puzzleId="@puz.ID">@RawHtmlHelper.Display(puz.Name, Model.Event.ID, Html)</a>
```
tabascq, line 26: Is this the right `asp-page` value? Looks like you deleted `Submissions` in the first line of this review.

jbodner09: It goes to the right place in my local testing.  From what I can tell, that @page path at the top of the file is what the URL _shows_, but the path to the file based on our directory structure is what this link uses to locate what file to actually go to.  They don't necessarily have to match.  In fact, when I had this link use the @page path, it didn't work!  You can also see that this is exactly what our existing pages do:  the Play page goes to the Index page according to the link, but the actual Index page itself never shows the word Index in its @page path, which is why I modelled this POC off of that.

jbodner09: Oops, hope that guy I accidentally pinged with the @ ignores this!

tabascq (approved):

PR #774, opened 2023-02-21
Use object instead of iframe for PDFs in prototype

jbodner09:
Apparently Chromium-based browsers don't like showing PDFs in iframes, so use an object-embed sandwich for PDFs instead.  Firefox worked fine with iframes and still works fine with this.  Also added a download link just in case something still goes wrong with the browser's built-in PDF viewer.

### DIFF
--- a/ServerCore/Pages/Submissions/SubmitProto.cshtml
+++ b/ServerCore/Pages/Submissions/SubmitProto.cshtml
@@ -16,6 +16,10 @@
         width: 100%;
         height: calc(100vh - 265px);
     }
+    #pdf-frame {
+        width: 100%;
+        height: calc(100vh - 285px);
+    }
 </style>
 @if (Model.Puzzle.CustomURL != null)
 {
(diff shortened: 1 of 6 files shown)


ServerCore/Pages/Submissions/SubmitProto.cshtml
```diff
@@ -16,15 +16,22 @@
...
+    }
 </style>
 @if (Model.Puzzle.CustomURL != null)
 {
     <iframe id="puzzle-frame" title="@Model.Puzzle.Name" sandbox="allow-same-origin allow-scripts" src="@PuzzleHelper.GetFormattedUrl(Model.Puzzle, Model.Event.ID)"></iframe>
 }
 else if (Model.PuzzleFile != null)
 {
-    <iframe id="puzzle-frame" title="@Model.Puzzle.Name" sandbox="allow-same-origin allow-scripts" src="@Model.PuzzleFile.Url.AbsoluteUri"></iframe>
+    <object id="pdf-frame" title="@Model.Puzzle.Name" data="@Model.PuzzleFile.Url.AbsoluteUri" type="application/pdf">
```
tabascq, line 30: Not important for a prototype, but this should eventually do some extension checking.

tabascq (approved):

PR #778, opened 2023-03-14
send prerequisites to the correct place

digit01Wave:
Currently, clicking on a puzzle in the prerequisite list just links you back to the top of the current puzzle.
This fix makes sure the link will actually take you to the prerequisite puzzle

The problem was that the link was creating /admin/Puzzles/Edit/[currentPuzzleId]?Id=[prerequisitePuzzleId]

What we wanted was /admin/Puzzles/Edit/[prerequisitePuzzleId]

For example, for the GetHoppingPuzzle (id1) we want all the other links to link to the edits of the other puzzle ids
image

### DIFF
--- a/ServerCore/Pages/Puzzles/Edit.cshtml
+++ b/ServerCore/Pages/Puzzles/Edit.cshtml
@@ -210,7 +210,7 @@
                     {
                         <tr>
                             <td>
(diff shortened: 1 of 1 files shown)


tabascq (approved):
Thanks!

Issue #769, opened 2023-02-14
Use a lot less of @Html.DisplayFor


tabascq:
Turns out that this is causing lots of runtime JITting, and removing this from the add-an-admin page turns the page load time from ~7s to ~1s.

We should do a systematic purge of this construct except for places where we actually use it. Those places should be places that have [Display....] attributes.

tabascq:
Skip this for properties with attributes [DisplayFormat] or [Display]

tabascq:
Also handle @Html.DisplayNameFor

digit01Wave:
> Html.DisplayNameFor

@tabascq , for these, why can't we just hard code the name into the html rather than using the Html.DisplayNameFor? Is it just so that the display name change can be shown in one place?

tabascq:
You can hardcode it - we'll lose a little flexibility, but we aren't using said flexibility.

digit01Wave:
Ok. I will hardcode the Display but will leave display format

PR #779, opened 2023-03-14
Fix display name html

digit01Wave:
Remove usage of Html.DisplayNameFor and Html.DisplayFor except for items with [Display(...)] or [DisplayFormat(...)]

Addresses issue
#769 

Also fixed delete page
Before fix
image

After fix
image

### DIFF
--- a/ServerCore/Pages/Events/Delete.cshtml
+++ b/ServerCore/Pages/Events/Delete.cshtml
@@ -27,10 +27,13 @@
             UrlString
         </dt>
         <dd>
-            HomePartial
-            </dt>
-        <dd>
             @Model.Event.UrlString
+        </dd>
+        <dt>
+            HomePartial
+        </dt>
+        <dd>
+            @Model.Event.HomePartial
         </dd>
         <dt>
             MaxNumberOfTeams
--- a/ServerCore/Pages/Events/Delete.cshtml
+++ b/ServerCore/Pages/Events/Delete.cshtml
@@ -18,127 +18,127 @@
     <hr />
     <dl class="dl-horizontal">
         <dt>
-            @Html.DisplayNameFor(model => model.Event.Name)
+            Name
         </dt>
         <dd>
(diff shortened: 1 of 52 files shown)


ServerCore/Pages/Puzzles/Feedback.cshtml
```diff
@@ -24,8 +24,8 @@ else
 }
 <div>
     <br />
-    <h4><b>Avg Fun Rating:</b> @Html.DisplayFor(x => x.FunScore) / @FeedbackModel.FeedbackMax</h4>
-    <h4><b>Avg Difficulty Rating:</b> @Html.DisplayFor(x => x.DiffScore) / @FeedbackModel.FeedbackMax</h4>
+    <h4><b>Avg Fun Rating:</b> @Model.FunScore / @FeedbackModel.FeedbackMax</h4>
```
tabascq, line 27: These two have a [DisplayFormat] attribute, so something was lost here.

ServerCore/Pages/Events/Delete.cshtml
```diff
@@ -18,127 +18,127 @@
...
-            @Html.DisplayNameFor(model => model.Event.Name)
+            Name
         </dt>
         <dd>
-            @Html.DisplayFor(model => model.Event.Name)
+            @Model.Event.Name
         </dd>
         <dt>
-            @Html.DisplayNameFor(model => model.Event.UrlString)
+            UrlString
```
tabascq, line 27: You didn't break anything here, but it looks like something is off WRT UrlString and HomePartial as they do not match the dt-followed-by-dd pattern used by the other properties. Not an important fix as we rarely delete events anyway...

tabascq (changes requested):

ServerCore/Pages/Puzzles/Feedback.cshtml
digit01Wave: Ah good catch. Thanks

tabascq (approved):

PR #785, opened 2023-04-13
Fixed redirect issue

digit01Wave:
#784 
Fixes redirect issue. Now the login and logout should redirect back to the previous page the user was on.


tabascq (approved):
Neat! Thanks so much!

PR #790, opened 2023-05-13
Regenerated ContextModel with new EntityFramework version

asyasky:
The EF version was updated with .Net but the first time a migration is created on the new version it's going to regenerate the ContextModel with the new standard syntax. I'm separating out that change to make sure it doesn't have any unintended effects. I didn't spot any breaking changes in the code or docs, but I want to get a few eyes on it to be sure.

### DIFF
--- a/Data/Migrations/PuzzleServerContextModelSnapshot.cs
+++ b/Data/Migrations/PuzzleServerContextModelSnapshot.cs
@@ -6,6 +6,8 @@ using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
 using ServerCore.DataModel;
 
+#nullable disable
+
 namespace Data.Migrations
 {
     [DbContext(typeof(PuzzleServerContext))]
@@ -15,9 +17,13 @@ namespace Data.Migrations
         {
 #pragma warning disable 612, 618
             modelBuilder
-                .UseIdentityColumns()
-                .HasAnnotation("Relational:MaxIdentifierLength", 128)
-                .HasAnnotation("ProductVersion", "5.0.2");
+                .HasAnnotation("ProductVersion", "7.0.2")
+                .HasAnnotation("Proxies:ChangeTracking", false)
+                .HasAnnotation("Proxies:CheckEquality", false)
+                .HasAnnotation("Proxies:LazyLoading", true)
+                .HasAnnotation("Relational:MaxIdentifierLength", 128);
+
+            SqlServerModelBuilderExtensions.UseIdentityColumns(modelBuilder);
 
             modelBuilder.Entity("Microsoft.AspNetCore.Identity.IdentityRole", b =>
                 {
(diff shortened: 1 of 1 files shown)


Data/Migrations/PuzzleServerContextModelSnapshot.cs
```diff
@@ -15,9 +17,13 @@ protected override void BuildModel(ModelBuilder modelBuilder)
         {
 #pragma warning disable 612, 618
             modelBuilder
-                .UseIdentityColumns()
-                .HasAnnotation("Relational:MaxIdentifierLength", 128)
-                .HasAnnotation("ProductVersion", "5.0.2");
+                .HasAnnotation("ProductVersion", "7.0.2")
+                .HasAnnotation("Proxies:ChangeTracking", false)
```
morganbr, line 21: How sure are you that we didn't use change tracking?

Data/Migrations/PuzzleServerContextModelSnapshot.cs
```diff
@@ -15,9 +17,13 @@ protected override void BuildModel(ModelBuilder modelBuilder)
         {
 #pragma warning disable 612, 618
             modelBuilder
-                .UseIdentityColumns()
-                .HasAnnotation("Relational:MaxIdentifierLength", 128)
-                .HasAnnotation("ProductVersion", "5.0.2");
+                .HasAnnotation("ProductVersion", "7.0.2")
+                .HasAnnotation("Proxies:ChangeTracking", false)
+                .HasAnnotation("Proxies:CheckEquality", false)
```
morganbr, line 22: Same question here

Data/Migrations/PuzzleServerContextModelSnapshot.cs
```diff
@@ -43,15 +49,16 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                         .HasDatabaseName("RoleNameIndex")
                         .HasFilter("[NormalizedName] IS NOT NULL");
 
-                    b.ToTable("AspNetRoles");
+                    b.ToTable("AspNetRoles", (string)null);
```
morganbr, line 52: What happens with a null schema?

Data/Migrations/PuzzleServerContextModelSnapshot.cs
```diff
@@ -43,15 +49,16 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
 
             modelBuilder.Entity("Microsoft.AspNetCore.Identity.IdentityRoleClaim<string>", b =>
                 {
                     b.Property<int>("Id")
                         .ValueGeneratedOnAdd()
-                        .HasColumnType("int")
-                        .UseIdentityColumn();
+                        .HasColumnType("int");
+
+                    SqlServerPropertyBuilderExtensions.UseIdentityColumn(b.Property<int>("Id"));
```
morganbr, line 61: Is this an identical result to the UseIdentityColumn above?

morganbr (commented):
Added my questions for the change

asyasky:
No longer needed

Issue #720, opened 2022-07-19
Feature Request:  Make pregame puzzles solvable per individual, not per team


jbodner09:
Ideally, pregame puzzles could be solvable even if you're not on a team, and then the amount you've solved could also be seen by other teams in case they want to add players who are actually serious about the event.  And then, once you're on a team, other teammates' solves of pregame puzzles won't mark the puzzle as solved for you, so you can still have a chance to submit an answer even once you're on a team.  @philipzloh can also explain more.

vidapour:
When flag is on, show submissions per player instead of per team.

vidapour:
@metapzl starting discussion/planning of this item.

digit01Wave:
**How this should work:**
- If we can make it so that each person has their own list
then these could just have no points no problem
- There's a set of puzzles called your team puzzles. Then there's a set of puzzles called your individual puzzles
- For an event you'll have the team puzzles be the event puzzles
- The individual puzzles will be the pre-event puzzles. Those won't score team points. Any account that ever exists can have individual puzzles, they don't need a team to see them or submit

jbodner09:
Closing since an event has successfully done this thanks to the above updates.

PR #791, opened 2023-05-15
SinglePlayerPuzzles Part1: Data migration

digit01Wave:
#720 (part 1)
We want to introduce a concept of a SinglePlayerPuzzle, which can be solved as an individual player rather than as a team.
The player does not have to be on a team.

This PR is the first part that adds the necessary new tables. These parts generally split up this massive code change (https://github.com/PuzzleServer/mainpuzzleserver/pull/792)

This PR adds the following 4 tables to the database:
- `SinglePlayerPuzzleHintStatePerPlayer` - every row is the hint state of a specific player for a specific single player puzzle
- `SinglePlayerPuzzleStatePerPlayer` - every row is the puzzle state (solve time, lockout, etc.) of a specific player for a specific single player puzzle
- `SinglePlayerPuzzleSubmission` - every row is is a submission to a single player puzzle
- `SinglePlayerPuzzleUnlockState` - every row is the unlock state of a given SinglePlayerPuzzle. <br />Note that this is different from the regular PuzzleStatePerTeam that has an unlock property that the author/admin can set to manually unlock the puzzle for a specific team. When it comes to single player puzzles, the author/admin will only be able to unlock ALL or lock ALL. This is because we don't want to keep adding all users to the SinglePlayerPuzzleStatePerPlayer every time someone new logs in.


Data/Migrations/PuzzleServerContextModelSnapshot.cs
```diff
@@ -15,9 +17,13 @@ protected override void BuildModel(ModelBuilder modelBuilder)
         {
 #pragma warning disable 612, 618
             modelBuilder
-                .UseIdentityColumns()
-                .HasAnnotation("Relational:MaxIdentifierLength", 128)
-                .HasAnnotation("ProductVersion", "5.0.2");
+                .HasAnnotation("ProductVersion", "7.0.2")
+                .HasAnnotation("Proxies:ChangeTracking", false)
```
morganbr, line 21: Please coordinate with @asyasky on this and #790

ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -98,7 +98,7 @@
                 <span>You've submitted too many wrong responses. Please <a href="@mailtoUrl">use this link to email us your answer for @Model.Puzzle.PlaintextName or ask for help</a>.</span>
             </div>
         }
-        else if (Model.PuzzleState.IsTeamLockedOut)
+else if (Model.PuzzleState.IsLockedOut)
```
morganbr, line 101: Nit: It looks like something happened to indentation here

morganbr (approved):
Looks good once the EF migration questions are solved

ServerCore/Pages/Submissions/Index.cshtml
digit01Wave: fixed in newest iteration

Data/DataModel/SinglePlayerPuzzleUnlockState.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.ComponentModel.DataAnnotations.Schema;
+
+namespace ServerCore.DataModel
+{
+    /// <summary>
+    /// The unlock state of a single player puzzle.
+    /// Unlike puzzles for teams, which allows admins/authors to unlock the puzzle for a specific team,
+    /// they will only be able to either lock or unlock for ALL players.
```
tabascq, line 9: How will we test these puzzles then?

tabascq: Will we treat them as standard team puzzles and switch them over late in the process?

digit01Wave: So. I was thinking that we would lock the puzzle behind a pre-requisite and we can mark as solved the prerequisite to do this. A little hacky but better than constantly checking for every player that logs in.

digit01Wave: Another way of doing it is to add a button that manually adds a player to the SinglePlayerPuzzleStatePerPlayer and allowing unlocks the typical way. Which I actually like more actually

tabascq: If the way I suggested is less work then I think it's fine as all puzzles will start their lives that way.

digit01Wave: sounds good to me!

PR #798, opened 2023-06-13
SinglePlayerPuzzles Part3: Create single player puzzles

digit01Wave:
#720 part3
This PR allows for the creation of single player puzzles and allows the author/admin to switch the puzzle between SinglePlayerPuzzles and Team puzzles.

Note that we have still not allowed for players to see the puzzles or for the author/admin to unlock single player puzzles. That will be in the next PR


ServerCore/Pages/Puzzles/Edit.cshtml.cs
```diff
@@ -153,6 +159,41 @@ public async Task<IActionResult> OnPostAsync(int puzzleId)
...
+            if (oldPuzzleView.IsForSinglePlayer != Puzzle.IsForSinglePlayer)
+            {
+                if (Puzzle.IsForSinglePlayer) // If switch from team puzzle to single player puzzle
+                {
+                    if (!_context.SinglePlayerPuzzleUnlockStates.Where(unlockState => unlockState.PuzzleID == oldPuzzleView.ID).Any())
+                    {
+                        _context.SinglePlayerPuzzleUnlockStates.Add(new SinglePlayerPuzzleUnlockState() { PuzzleID = oldPuzzleView.ID });
+                    }
+
+                    List<PuzzleStatePerTeam> allPsptsToDelete = await (from pspt in _context.PuzzleStatePerTeam
```
tabascq, line 171: Do we *have* to delete this state? This makes it so accidents lose solver data. Same question for removing single-player state.

digit01Wave: We do not have to delete this state. But if we don't, we will then have to make sure any references to context.PlayerStatePerTeam needs to also check that IsSinglePlayerPuzzle is not true. I thought that might be hard to remember to do for newer devs. I am fine with doing that of course. What do you think?

digit01Wave: I realize the other issue is that referencing PuzzleStatePerTeam.Puzzle.IsSinglePlayerPuzzle would be an access to another table for some queries that previously only queried TeamID or PuzzleID, so it also may affect performance.

Another option is to not let authors edit this property. That is, if they want a single player puzzle, they would have to make a whole another puzzle and delete the previous one.

tabascq (approved):

PR #801, opened 2023-06-15
SinglePlayerPuzzles Part4: Let players see SinglePlayerPuzzles

digit01Wave:
#720 part4
This PR lets players see the list of single player puzzles even if they are not yet registered for the event.

It also allows the author or admin to unlock the puzzle for the player.

Note that the player still cannot actually submit answers or take hints. That will happen in the next PR.

- Admin/Author status view when locked
image
- Admin/Author status view when click unlock
image
image
- Player view when not registered
image
- Player view when registered
image


asyasky (approved):
LGTM. Out of curiosity, why did you decide to make a new page instead of a filtered version of the existing one?

digit01Wave:
> version

I made them two separate pages for two reasons
1. The current team puzzles assume awareness of team (path is Team/<teamID/play) and inherits from 
2. I was worried that players would not notice the filter

PR #810, opened 2023-06-20
SinglePlayerPuzzles Part5: Handle unlocks and submission 

digit01Wave:
#720 part 5

Final mandatory part of single player puzzles.
We can now unlock single player puzzles and players can also submit answers to them.

Note that this still does not include hints.


ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -173,16 +192,27 @@ public async Task<IActionResult> OnPostAsync(int puzzleId, string submissionText
...
+                            Team,
+                            true);
+                    }
 
                     var authors = await _context.PuzzleAuthors.Where((pa) => pa.Puzzle == submission.Puzzle).Select((pa) => pa.Author.Email).ToListAsync();
-
+                    string affectedEntity = IsPuzzleForSinglePlayer ? $"User {LoggedInUser.Name ?? LoggedInUser.Email}" : $"Team {Team.Name}";
                     MailHelper.Singleton.SendPlaintextBcc(authors,
-                        $"{Event.Name}: Team {submission.Team.Name} is in email mode for {submission.Puzzle.Name}",
+                        $"{Event.Name}: Team {affectedEntity} is in email mode for {submission.Puzzle.Name}",
```
tabascq, line 215: You don't want "Team" here, right?

ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -233,38 +284,70 @@ private async Task SetupContext(int puzzleId)
...
+                                         select new SubmissionView()
+                                         {
+                                             Submission = submission,
+                                             Response = response,
+                                             SubmitterName = user.Name,
+                                             FreeformReponse = submission.FreeformResponse,
+                                             IsFreeform = Puzzle.IsFreeform
+                                         }).ToListAsync();
+
+                PuzzlesCausingGlobalLockout = await SinglePlayerPuzzleStateHelper.PuzzlesCausingGlobalLockout(_context, Event, LoggedInUser.ID).ToListAsync();
```
tabascq, line 311: Pairing the global lockout story per puzzle type feels odd, but since we aren't using it for PD2023 it's not super critical.

ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -233,38 +284,70 @@ private async Task SetupContext(int puzzleId)
...
+            List<SubmissionView> submissionViews = new List<SubmissionView>();
+            if (Puzzle.IsForSinglePlayer)
+            {
+                PuzzleState = await SinglePlayerPuzzleStateHelper.GetOrAddStateIfNotThere(_context,
+                    Event,
+                    Puzzle,
+                    LoggedInUser.ID);
+
+                SubmissionViews = await (from submission in _context.SinglePlayerPuzzleSubmissions
+                                         join user in _context.PuzzleUsers on submission.Submitter equals user
```
tabascq, line 297: Don't we just want `where submission.Submitter == LoggedInUser.ID`? No join needed if so...

tabascq (changes requested):

ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -233,38 +284,70 @@ private async Task SetupContext(int puzzleId)
...
+                        _context,
+                        Event,
+                        Puzzle,
+                        Team))
+                    .FirstAsync();
+
+                SubmissionViews = await (from submission in _context.Submissions
+                                         where submission.Team == Team
+                                            && submission.Puzzle == Puzzle
+                                            && submission.Submitter.ID == LoggedInUser.ID
```
tabascq, line 328: I think you went too far when changing the single-player query - I should be able to see all submissions from my teammates, so the team query was good as-is.

tabascq (approved):

ServerCore/Pages/Submissions/Index.cshtml.cs
digit01Wave: Fixed in newest commit

PR #814, opened 2023-06-22
Admin page to see team lunch orders

morganbr:
Shows a summary report and individual team lunch orders. Includes defaulting for lunches that were not chosen

### DIFF
--- a/ServerCore/Pages/Swag/TeamReport.cshtml.cs
+++ b/ServerCore/Pages/Swag/TeamReport.cshtml.cs
@@ -1,0 +1,120 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Identity;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.EntityFrameworkCore;
+using ServerCore.DataModel;
+using ServerCore.ModelBases;
+
+namespace ServerCore.Pages.Swag
+{
+    /// <summary>
+    /// Model for the admin's view of the swag items for a team
+    /// </summary>
+    [Authorize(Policy = "IsEventAdmin")]
+    public class TeamReportModel : EventSpecificPageModel
+    {
+        public TeamReportModel(PuzzleServerContext serverContext, UserManager<IdentityUser> userManager) : base(serverContext, userManager)
+        {
+        }
+
+        /// <summary>
+        /// Individual order for a team
+        /// </summary>
+        public class SwagView
+        {
+            public string ContactEmail { get; set; }
+            public string TeamName { get; set; }
+            public string Lunch { get; set; }
+            public int TeamId { get; set; }
+        }
+
+        /// <summary>
+        /// Aggregate orders for an item
+        /// </summary>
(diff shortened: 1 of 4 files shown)


tabascq:
Can you explain more precisely what this is displaying? I can't follow the query logic through all the grouping etc. Example questions:

- Does it show orders per team? Does it show how many of those orders are default?
- Does it show cumulative orders for the event? Does it show how many of those orders are default?
- Does it keep track of teams that have ordered more pizzas than they are permitted, due to team member cancellations? (Is this even possible?)

Some sample output would help.

ServerCore/Pages/Swag/TeamReport.cshtml.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        public List<SwagView> SwagViews { get; set; }
+        public List<SummaryItem> Summary { get; set; }
+
+        public async Task<IActionResult> OnGetAsync()
+        {
+            // Fetch lunches the team chose
+            SwagViews = await (from teamLunch in _context.TeamLunch
+                         join team in _context.Teams on teamLunch.TeamId equals team.ID
```
digit01Wave, line 51: question: Does this need a join if Team is already a virtual navigation property of TeamLunch?

ServerCore/Pages/Swag/TeamReport.cshtml.cs
```diff
@@ -0,0 +1,120 @@
...
+            // Find out how many lunches weren't ordered and should be default:
+            // 1. Find out how many got ordered
+            var teamLunchCounts = from team in SwagViews
+                             group team by team.TeamId into teamGroup
+                             select new { TeamId = teamGroup.Key, Count = teamGroup.Count() };
+
+            // 2. Find out how many lunches the team is allowed to order
+            var teamEligibleMembers = await (from team in _context.Teams
+                                      join member in _context.TeamMembers on team equals member.Team
+                                      join player in _context.PlayerInEvent on member.Member equals player.Player
```
digit01Wave, line 71: [question]: Does this also need a join if Player is a navigation property of TeamMembers?

morganbr:
Here's a sample screenshot where a team has chosen one pizza and has failed to choose another and is getting the default of cheese
image

morganbr:
>     * Does it show orders per team? Does it show how many of those orders are default?
It shows totals and each team's orders

>     * Does it show cumulative orders for the event? Does it show how many of those orders are default?
Yes and yes

>     * Does it keep track of teams that have ordered more pizzas than they are permitted, due to team member cancellations? (Is this even possible?)
It could show that, but doesn't currently. Do we need it?

> Some sample output would help.
See above

ServerCore/Pages/Swag/TeamReport.cshtml.cs
morganbr: Using the navigation property also does a join, this is just more explicit

ServerCore/Pages/Swag/TeamReport.cshtml.cs
morganbr: Similarly, these navigation properties do the same kind of join

asyasky (approved):

PR #826, opened 2023-06-28
Embed cleanup

tabascq:
- Don't crash when there are no authors
- Don't show previous submissions when there are no previous submissions

### DIFF
--- a/ServerCore/Pages/Submissions/Index.cshtml
+++ b/ServerCore/Pages/Submissions/Index.cshtml
@@ -509,7 +509,7 @@ else
     <br />
     <div class="centerwrapper">
         <div class="inlineblockwrapper">
-            <div id="submissionHistory" class="hideablepanel">
+            <div id="submissionHistory" style="display: none" class="hideablepanel">
                 Click here to view previous submissions:<span class="caret"></span>
                 <div class="panelcontents" style="display: none">
                     <div class="gridwrapper">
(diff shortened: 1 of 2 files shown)


ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -509,7 +509,7 @@ else
     <br />
     <div class="centerwrapper">
         <div class="inlineblockwrapper">
-            <div id="submissionHistory" class="hideablepanel">
+            <div id="submissionHistory" style="display: none" class="hideablepanel">
```
jbodner09, line 512: This won't be as simple as just hiding by default.  If you navigate to the page and there are already submissions, we do need to show the panel upon first load.  So, you'll want to remove this style tag here, and instead add some JS to optionally add the display:none if there are no submissions.  Line 626 right after the code to make the panel open and close should be a good place to do that.

jbodner09 (changes requested):

ServerCore/Pages/Submissions/Index.cshtml
tabascq: Oh yeah I had the right model in mind but I skipped a step when implementing. I meant to only add display:none if the model knows there are no submissions; then I'm guaranteed that the page won't jerk. I'll wrap this line in the appropriate check...

digit01Wave (approved):

jbodner09 (approved):

PR #834, opened 2023-06-28
Update hidden properties for PuzzleUser editing pages

asyasky:
PuzzleUser editing pages weren't updated when the AdminOrAuthor bool was added so it was getting overwritten whenever the PuzzleUser was updated.

### DIFF
--- a/ServerCore/Areas/Identity/Pages/Account/Manage/Index.cshtml.cs
+++ b/ServerCore/Areas/Identity/Pages/Account/Manage/Index.cshtml.cs
@@ -71,11 +71,12 @@ namespace ServerCore.Areas.Identity.Pages.Account.Manage
 
             var thisPuzzleUser = await PuzzleUser.GetPuzzleUserForCurrentUser(_context, User, _userManager);
 
-            // enforce access rights, do not let these change!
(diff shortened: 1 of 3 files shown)


ServerCore/Areas/Identity/Pages/Account/Manage/Index.cshtml.cs
```diff
@@ -75,6 +75,7 @@ public async Task<IActionResult> OnPostAsync()
             Input.ID = thisPuzzleUser.ID;
             Input.IsGlobalAdmin = thisPuzzleUser.IsGlobalAdmin;
             Input.IdentityUserId = thisPuzzleUser.IdentityUserId;
+            Input.MayBeAdminOrAuthor = thisPuzzleUser.MayBeAdminOrAuthor;
```
tabascq, line 78: Is there a reason that T-Shirt size is not present here?

asyasky: Not especially, I'll add it in so that it's consistent across pages.

tabascq (approved):

PR #836, opened 2023-06-29
Fixed single player state auth check

digit01Wave:


ServerCore/Areas/Identity/UserAuthorizationPolicy/AuthorizationHelper.cs
```diff
@@ -193,6 +193,12 @@ public async Task PlayerCanSeePuzzleCheck(AuthorizationHandlerContext authContex
                     {
                         authContext.Succeed(requirement);
                     }
+
+                    IQueryable<SinglePlayerPuzzleStatePerPlayer> statePerPlayerQ = SinglePlayerPuzzleStateHelper.GetFullReadOnlyQuery(dbContext, thisEvent, puzzle.ID, puzzleUser.ID);
```
tabascq, line 197: put all this in an else clause so it doesn't execute if the above succeeds?

tabascq (approved):

PR #842, opened 2023-07-02
Make metas embeddable

jbodner09:


ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -72,24 +72,37 @@
...
     }
+
     if (puzzleFilePath.Length > 0)
     {
-        if (puzzleFilePath.EndsWith(".pdf"))
-        {
-            embedType = 1;
-        }
-        else if ((puzzleFilePath.EndsWith(".htm")) || (puzzleFilePath.EndsWith(".html")))
+        if (!puzzleFilePath.ToLowerInvariant().StartsWith("https://puzzlehunt.azurewebsites.net"))
```
tabascq, line 78: This isn't right - should be:

```
string lowerPath = puzzleFilePath.ToLowerInvariant();
if (!lowerPath.StartsWith("https://puzzleserverteststore.blob.core.windows.net/") && !lowerPath.StartsWith("localhost"))
```

since the endpoint on puzzlehunt.azurewebsites.net is one of the ones we _want_.

jbodner09: I did add the localhost check you mentioned here since that's also good to have in there

tabascq (approved):

PR #845, opened 2023-07-04
Quick & dirty simplified state map

tabascq:
Could be further improved but thsi does the important stuff for PD2023


tabascq:
image

digit01Wave (approved):

PR #847, opened 2023-07-06
Add PD2023 lunch FAQs

NateD-MSFT:
Adding questions related to PD2023 pizzas.

### DIFF
--- a/ServerCore/Pages/Resources/pd2023/FAQPartial.cshtml
+++ b/ServerCore/Pages/Resources/pd2023/FAQPartial.cshtml
@@ -98,6 +98,20 @@ else
         </p>
 
         <h3>
+            Will food be provided?
+        </h3>
+        <p>
+            Pizza will be provided for all teams.  If you or a team member cannot eat pizza, please reach out to us and we will provide an alternative.
+        </p>
+
+        <h3>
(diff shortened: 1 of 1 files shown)


jbodner09 (approved):

ServerCore/Pages/Resources/pd2023/FAQPartial.cshtml
```diff
@@ -97,6 +97,20 @@ else
...
+        </h3>
+        <p>
+            Pizza will be provided for all teams.  If you or a team member cannot eat pizza, please reach out to us and we will provide an alternative.
+        </p>
+
+        <h3>
+            When and how do we get our food?
+        </h3>
+        <p>
+            Watch your inbox! Your team will receive an email letting you know when food is ready with instructions on where to go to pick it up.
```
tabascq, line 111: Will this mail come from puzzhunt or will it come from the bulk mailer on the site? If the latter, we need to also encourage them to monitor junk mail here.

NateD-MSFT: Great question.  I'll check with Morgan/Megan today.  I expect it's the bulk mailer.

PR #848, opened 2023-07-06
Add custom author field

jbodner09:
If there's custom author text, it will take precedence over the Puzzle Authors assigned to that puzzle.  I also removed the hard-coded "by" on the page so that, if for some reason there's no custom author text AND none of the authors have a visible display name, there won't be an orphaned "by" just hanging out there all lonely on the page.


ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -343,16 +343,28 @@ orderby submission.TimeSubmitted
...
                 {
-                    if ((PuzzleAuthor != null) && (PuzzleAuthor.Length > 0))
+                    if (CurrentAuthors[i].Name?.Trim().Length > 0)
                     {
-                        PuzzleAuthor += ", ";
+                        if ((PuzzleAuthor != null) && (PuzzleAuthor.Length > 0))
+                        {
+                            PuzzleAuthor += ", ";
+                        }
+                        else if ((PuzzleAuthor == null) || ((PuzzleAuthor != null) && (PuzzleAuthor.Length == 0)))
```
tabascq, line 362: The != null check is redundant since that's the only way to be on the right side of the ||.

This would be cleaner overall if you set `PuzzleAuthor = "";` on line 354 above the for loop. Then lines 358-365 boil down to this single line of readable code:

`PuzzleAuthor += (PuzzleAuthor.Length == 0) ? "by " : ", ";`

jbodner09: I like it!

tabascq (approved):

PR #852, opened 2023-07-07
Team Announcements

tabascq:
Adds a TeamAnnouncement property on Event and a ShowTeamAnnouncement property on Team.

Announcement UX only shows when the event has an a TeamAnnouncement set.

When a team has its announcement turned on/off, an email is sent to all team members and an announcement displays on the top of every page.


ServerCore/Pages/Teams/Index.cshtml.cs
```diff
@@ -21,5 +24,40 @@ public async Task<IActionResult> OnGetAsync()
             await LoadTeamDataAsync();
             return Page();
         }
+
+        public async Task<IActionResult> OnGetShowTeamAnnouncementAsync(int teamId, bool value)
+        {
+            var team = await _context.Teams.FirstAsync(m => m.ID == teamId);
```
morganbr, line 30: Should be FirstOrDefaultAsync

ServerCore/Pages/Teams/Index.cshtml.cs
```diff
@@ -21,5 +24,40 @@ public async Task<IActionResult> OnGetAsync()
...
+
+            if (team == null)
+            {
+                return NotFound();
+            }
+
+            team.ShowTeamAnnouncement = value;
```
morganbr, line 37: Check if this changed so we don't redo work/resend mail if the page reloads

morganbr (approved):

PR #853, opened 2023-07-07
Add conference room admin page

asyasky:
Ugly but functional way to add rooms for an event and assign them to teams in bulk.


ServerCore/Pages/Rooms/AddRooms.cshtml
```diff
@@ -0,0 +1,18 @@
...
+@model ServerCore.Pages.Rooms.AddRoomsModel
+
+@{
+    ViewData["Title"] = "Add Rooms";
+    ViewData["AdminRoute"] = "../Rooms/AddRooms";
+}
+<h2>Add Rooms</h2>
+
+<form method="post">
+    <textarea name="inputRooms" style="height:500px"></textarea>
```
tabascq, line 11: Define the format on this page?

tabascq (approved):

ServerCore/Pages/Rooms/AddRooms.cshtml
asyasky: Done

Data/PuzzleServerContext.cs
```diff
@@ -76,6 +77,7 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             modelBuilder.Entity<SinglePlayerPuzzleStatePerPlayer>().HasIndex(pspt => new { pspt.PlayerID });
             modelBuilder.Entity<SinglePlayerPuzzleStatePerPlayer>().HasIndex(pspt => new { pspt.PlayerID, pspt.SolvedTime });
             modelBuilder.Entity<SinglePlayerPuzzleUnlockState>().HasKey(state => new { state.PuzzleID });
+            modelBuilder.Entity<Room>().HasIndex(room => new { room.EventID, room.Building, room.Number }).IsUnique();
```
morganbr, line 80: Need to check for delete issues by ensuring deleting teams doesn't delete rooms and deleting events works

ServerCore/Helpers/TeamHelper.cs
```diff
@@ -46,6 +46,9 @@ public static async Task DeleteTeamAsync(PuzzleServerContext context, Team team,
                               select annotation;
             context.Annotations.RemoveRange(annotations);
 
+            var room = context.Room.Where(r => r.TeamID == team.ID).ToList();
```
morganbr, line 49: Deleting teams shouldn't delete rooms

morganbr: But it should unreserve them

ServerCore/Pages/Rooms/AddRooms.cshtml.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+        }
+
+        public void OnGet()
+        {
+        }
+
+        public async Task<IActionResult> OnPost(string inputRooms, string group) 
+        {
+            string[] rooms = inputRooms.Split("\r\n");
```
morganbr, line 24: It's possible this will have different line breaks on Mac and mobile. Not sure if there's a great way to deal with that

ServerCore/Pages/Rooms/RoomList.cshtml
```diff
@@ -0,0 +1,106 @@
...
+                            @Html.DisplayFor(modelItem => item.Team.Name)
+                        </td>
+                        <td>
+                            @Html.DisplayFor(modelItem => item.CurrentlyOnline)
+                        </td>
+                        <td>
+                            @Html.DisplayFor(modelItem => item.Group)
+                        </td>
+                        <td>
+                            @*                <a asp-page="./Edit" asp-route-id="@item.ID">Edit</a> ||
```
morganbr, line 97: Is this supposed to be here?

ServerCore/Pages/Rooms/RoomList.cshtml.cs
```diff
@@ -0,0 +1,105 @@
...
+                    }
+                }
+
+                await _context.SaveChangesAsync();
+                await transaction.CommitAsync();
+            }
+            return RedirectToPage("/Rooms/RoomList");
+        }
+
+        private async Task FillCollections()
```
morganbr, line 96: Nit -- end async method names with Async so callers know to await them

ServerCore/Pages/Rooms/RoomList.cshtml.cs
```diff
@@ -0,0 +1,105 @@
...
+
+            foreach (Tuple<string, bool> gr in Groups)
+            {
+                if (GroupToToggle.Contains(gr.Item1) && !gr.Item2)
+                {
+                    groupMapping.Add(gr.Item1, true);
+                }
+                else if(!GroupToToggle.Contains(gr.Item1) && gr.Item2)
+                {
+                    // TODO: Add notification or something for rooms going offline that have been assigned to teams
```
morganbr, line 46: Do we need this for this week?

ServerCore/Pages/Rooms/RoomList.cshtml.cs
```diff
@@ -0,0 +1,105 @@
...
+
+                await _context.SaveChangesAsync();
+                await transaction.CommitAsync();
+            }
+            return RedirectToPage("/Rooms/RoomList");
+        }
+
+        private async Task FillCollections()
+        {
+            Room = await _context.Room
```
morganbr, line 98: I don't see anything filtering on the event

ServerCore/Pages/Teams/Details.cshtml.cs
```diff
@@ -146,6 +148,10 @@ orderby lunch.ID
                 SoftMaxLunches = (int)Math.Ceiling(possibleInPersonMembers / (double)PlayersPerLunch);
             }
 
+            // Get team room
+            Room teamRoom = _context.Room.Where(r => r.TeamID == teamId).FirstOrDefault();
+            TeamRoom = $"{teamRoom.Building}/{teamRoom.Number}({teamRoom.Capacity})";
```
morganbr, line 153: This will throw if the team isn't assigned a room

ServerCore/Pages/Rooms/RoomList.cshtml.cs
```diff
@@ -0,0 +1,105 @@
...
+namespace ServerCore.Pages.Rooms
+{
+    [Authorize(Policy = "IsEventAdmin")]
+    public class RoomListModel : EventSpecificPageModel
+    {
+        public RoomListModel(PuzzleServerContext serverContext, UserManager<IdentityUser> manager) : base(serverContext, manager)
+        {
+        }
+
+        public IList<Room> Room { get; set; }
```
morganbr, line 22: If fixing pluralization here and in the db doesn't cause a mess, it would be nice

ServerCore/Pages/Rooms/AddRooms.cshtml.cs
jbodner09: It looks like the bulk response creation uses a StringReader.ReadLine(), which checks for all EOL possibilities.  I highly recommend doing it that way.

asyasky: Looks like Environment.NewLine should handle it without requiring me to rewrite the logic.

ServerCore/Helpers/TeamHelper.cs
asyasky: Done

ServerCore/Pages/Rooms/RoomList.cshtml.cs
asyasky: Done

ServerCore/Pages/Rooms/RoomList.cshtml.cs
asyasky: Probably not, but if we don't have this then the rooms & groups can get out of sync (which should also be ok since I'm just overwriting when there's a difference, but it's messy)

ServerCore/Pages/Rooms/RoomList.cshtml.cs
asyasky: Done

ServerCore/Pages/Rooms/RoomList.cshtml.cs
asyasky: Seems to be ok this time, I'm not seeing whatever I was running into before. Done.

ServerCore/Pages/Teams/Details.cshtml.cs
asyasky: Added null check.

ServerCore/Pages/Rooms/RoomList.cshtml
asyasky: I left it in because that's part of what I want to add next week, but I can find it in the history when I need it.

Data/PuzzleServerContext.cs
asyasky: Deleting teams now sets the teamid to null. Deleting events deletes the rooms, which is intended unless someone tells me otherwise.

ServerCore/Pages/Rooms/RoomList.cshtml.cs
asyasky: Oh, do you mean the notification? I can put that in another PR, but right now there isn't any way to take a specific conference room offline with the UI so we're going to want to handle things manually anyway during the event.

PR #854, opened 2023-07-07
Create a page to record all team merges

jbodner09:
On the list of all teams in the admin view, you'll now find a "Merged Teams" link that lists teams that don't exist anymore and what team they were merged into.  An example (ignore the center alignment, I changed that to left alignment):  

image

image

image

image


ServerCore/Pages/Teams/TeamMergeSummary.cshtml.cs
```diff
@@ -0,0 +1,49 @@
...
+            : base(serverContext, userManager)
+        {
+        }
+
+        public class TeamSummary{
+            public TeamSummary(int teamId, string teamName, string mergedTeams)
+            {
+                TeamID = teamId;
+                TeamName = teamName;
+                MergedTeams = mergedTeams;
```
tabascq, line 28: I would consider decoding here but no biggie

tabascq (approved):

ServerCore/Pages/Teams/TeamMergeSummary.cshtml.cs
jbodner09: Recorded in #855

PR #865, opened 2023-07-18
Map Meta Sort

tabascq:
Automatically move metas more to the left within their groups.


jbodner09 (approved):

ServerCore/Pages/Events/Map.cshtml.cs
```diff
@@ -49,12 +49,14 @@ public async Task<IActionResult> OnGetAsync(int? refresh, string? groups)
             if (EventRole == EventRole.admin)
             {
                 puzzles = await _context.Puzzles.Where(p => p.Event == Event)
+                    .Where(p => !p.IsForSinglePlayer)
```
morganbr, line 52: Why exclude single player puzzles from the state map?

morganbr (approved):

PR #886, opened 2023-07-26
Add Blazor infrastructure

morganbr:
Adds the infrastructure to allow adding Blazor components. Does not include any Blazor components, but an example is in #885 . Also includes a field to disable Blazor on the event in case of performance issues in early usage.


ServerCore/Pages/Shared/_Layout.cshtml
```diff
@@ -16,11 +16,22 @@
...
-    <!-- TODO: Make this actually scalable and correct -->
+    <base href="~/" />
+
     @{
         // Using ev here so that it isn't the same as Event that we bind in the <body>
         // Works fine if we use Event, but VS will yell at you with an error in the IDE
         var ev = (Model as ServerCore.ModelBases.EventSpecificPageModel)?.Event;
+    }
+
+    @if (ev == null || ev.AllowBlazor)
```
tabascq, line 27: For now let's not do this for the null event case so the ripcord is easier to pull.

morganbr: Done

tabascq (approved):

Issue #867, opened 2023-07-18
Merge pages for per-person and per-team puzzles -- two separate lists


tabascq:
There's really no reason for these to be separate views.

vidapour:
Add filter for indv. vs team puzzles

PR #889, opened 2023-07-28
Merge single player and team puzzle pages to one page

digit01Wave:
#867 
Merge the Team puzzle and Single Player puzzle lists into one

image

**What it looks like when there are only single player puzzles**
image

**What it looks like when there are only team puzzles**
image

**What it looks like when there are neither**
image

**What it looks like when you are not on a team**
image

**What it looks like when SinglePlayerPuzzles are not even enabled**
image


ServerCore/Pages/Index.cshtml
```diff
@@ -9,7 +9,7 @@
 <h4>Microsoft Non-Intern Puzzleday 2023</h4>
 <img src="~/images/pd2023/nipd2023logo.png" height="256" width="256" title="Puzzleday 2023 logo" alt="Puzzleday 2023" />
 <p>A one-day rebroadcast for employees and friends that happens every summer. Puzzle solvers of all levels of experience are welcome!</p>
-<a href="https://puzzlehunt.azurewebsites.net/nipd2023/play">Go to the Non-Intern Puzzleday event</a>
```
jbodner09, line 12: Please undo all these changes to the main landing page.  I assume this is just a find-replace gone amok, but these links are to each event's homepage, where play is just the role (as opposed to author or admin).

jbodner09 (changes requested):

ServerCore/Pages/Index.cshtml
digit01Wave: Thanks for the catch! Fixed

ServerCore/Pages/Puzzles/Create.cshtml
```diff
@@ -5,7 +5,7 @@
     ViewData["Title"] = "Create puzzle";
     ViewData["AdminRoute"] = "/Puzzles/Create";
     ViewData["AuthorRoute"] = "/Puzzles/Create";
-    ViewData["PlayRoute"] = "/Teams/Play";
+    ViewData["PlayRoute"] = "/Puzzles/PlayPuzzles";
```
tabascq, line 8: Any reason this can't be renamed to "/Puzzles/Play"?

ServerCore/Pages/Puzzles/PlayPuzzles.cshtml.cs
```diff
@@ -0,0 +1,237 @@
...
+        {
+            foreach (var puzzleView in puzzleViews)
+            {
+                files.TryGetValue(puzzleView.ID, out ContentFile content);
+                puzzleView.Content = content;
+            }
+
+            if (ShowAnswers)
+            {
+                Dictionary<int, ContentFile> answers = await(from file in _context.ContentFiles
```
tabascq, line 225: No reason to do this twice - precompute this in OnGetAsync (it can be null if ShowAnswers is false) and pass as it an argument.

ServerCore/Pages/Puzzles/PlayPuzzles.cshtml
```diff
@@ -0,0 +1,368 @@
...
+    <div class="alert alert-warning" role="alert">
+        <h4>The event is over.</h4>
+        <span>You may check answers, but they will not count toward your ranking.</span>
+    </div>
+}
+
+<h2>Team puzzles</h2>
+@if (Model.Team == null)
+{
+    <p>You cannot see team puzzles because you are not yet on a team!</p>
```
tabascq, line 39: Add a link to /Teams/List?

ServerCore/Pages/Puzzles/PlayPuzzles.cshtml
```diff
@@ -0,0 +1,368 @@
...
+    <p>Looks like there are no Team puzzles unlocked yet.</p>
+}
+else
+{
+    <div>
+        <a asp-page="/Teams/Answers" asp-route-teamId="@Model.Team.ID">View All Correct Team Answers</a>
+    </div>
+    <br />
+    <div>
+        <a asp-page="./PlayPuzzles"
```
tabascq, line 52: Maybe just one of these all/unsolved links, above both lists?

ServerCore/Pages/Puzzles/PlayPuzzles.cshtml.cs
```diff
@@ -0,0 +1,237 @@
...
+        private const SortOrder DefaultSort = SortOrder.GroupAscending;
+
+        public bool ShowAnswers { get; set; }
+
+        public bool AllowFeedback { get; set; }
+
+        public Team Team { get; set; }
+
+        public async Task OnGetAsync(
+            SortOrder? teamPuzzleSort,
```
tabascq, line 45: I don't think we really need two sort parameters here. I think it would be OK to have one parameter and have both column headers do the same thing. I realize this is in some ways the opposite of the advice I'm giving for the unsolved link, but both tables are going to have headers anyway so there's no visual space to save in this case.

ServerCore/Pages/Puzzles/PlayPuzzles.cshtml
```diff
@@ -0,0 +1,368 @@
...
+}
+
+<h2>Team puzzles</h2>
+@if (Model.Team == null)
+{
+    <p>You cannot see team puzzles because you are not yet on a team!</p>
+}
+else if (!Model.VisibleTeamPuzzleViews.Any())
+{
+    <p>Looks like there are no Team puzzles unlocked yet.</p>
```
tabascq, line 43: "available" instead of "unlocked"?

ServerCore/Pages/Puzzles/PlayPuzzles.cshtml
```diff
@@ -0,0 +1,368 @@
...
+        </tbody>
+    </table>
+}
+
+@if (Model.Event.ShouldShowSinglePlayerPuzzles)
+{
+    <h2>@Model.Event.SinglePlayerPuzzleTitle</h2>
+    @if (!Model.VisibleSinglePlayerPuzzleViews.Any())
+    {
+        <p>Looks like there are no puzzles unlocked here yet!</p>
```
tabascq, line 219: "available"?

tabascq:
Oh, one more comment - the errata note now looks like it only applies to pregame puzzles and I would move it above both lists (and consider only showing it when at least one puzzle has errata).

ServerCore/Pages/Puzzles/PlayPuzzles.cshtml.cs
digit01Wave: Wouldn't it be unintuitive if I were to click the "Sort by Title" on the pregame list and it sorts both tables?

ServerCore/Pages/Puzzles/Create.cshtml
digit01Wave: Fixed in newest iteration

ServerCore/Pages/Puzzles/PlayPuzzles.cshtml.cs
digit01Wave: Fixed in newest iteration

ServerCore/Pages/Puzzles/PlayPuzzles.cshtml
digit01Wave: Fixed in newest iteration

ServerCore/Pages/Puzzles/PlayPuzzles.cshtml
digit01Wave: Fixed in newest iteration

ServerCore/Pages/Puzzles/PlayPuzzles.cshtml
digit01Wave: Fixed in newest iteration

ServerCore/Pages/Puzzles/PlayPuzzles.cshtml
digit01Wave: Fixed

digit01Wave:
> Oh, one more comment - the errata note now looks like it only applies to pregame puzzles and I would move it above both lists (and consider only showing it when at least one puzzle has errata).

Fixed

ServerCore/Pages/Puzzles/Play.cshtml.cs
```diff
@@ -0,0 +1,247 @@
...
+            SortOrder? singlePlayerPuzzleSort,
+            PuzzleStateFilter? stateFilter)
+        {
+            TeamPuzzleSort = teamPuzzleSort;
+            SinglePlayerPuzzleSort = singlePlayerPuzzleSort;
+            StateFilter = stateFilter;
+            ShowAnswers = Event.AnswersAvailableBegin <= DateTime.UtcNow;
+            AllowFeedback = Event.AllowFeedback;
+
+            Dictionary<int, ContentFile> answers = await (from file in _context.ContentFiles
```
tabascq, line 55: Something seems wrong here - `answers` is ignored and then the computation of `answerFiles` is going to leave it empty because the `where` clause never gets answer files?

tabascq (changes requested):

ServerCore/Pages/Puzzles/PlayPuzzles.cshtml.cs
tabascq: I'm fine with it as is. I don't think the scenario of looking at both lists is all that common.

ServerCore/Pages/Puzzles/Play.cshtml.cs
digit01Wave: This function does get answer files, but they don't get used for anything because line 240 only includes it if ShowAnswers is available

image

tabascq: That's not what I mean. Unless I am misreading the code, the local variable `answers` has the answers and does not appear to ever be used. The variable `answerFiles`, which is passed to `PopulatePuzzleViewsWithFiles` (the code you highlighted) in both cases, seems like it will always be empty, because it is built on line 71 but the `where` clause on line 61 never includes actual answer files. Am I misreading?

digit01Wave: AH. That was an unused variable that I left in there. Thanks for catching

tabascq: You've addressed half of this issue (`answers` being unused). The other half (`answerFiles` being always empty) should be addressed when you change:

`where file.Event == Event && file.FileType == ContentFileType.Puzzle`

to:

`where file.Event == Event && (file.FileType == ContentFileType.Puzzle || file.FileType == ContentFileType.Answer)`

(unless I am misreading something)

digit01Wave: omg. You are right. What would we do without you. Thanks again!

tabascq (approved):

jbodner09 (approved):

PR #893, opened 2023-08-10
Improve embedded puzzle answer submission response displays

### LINKED ISSUE #856: Improvements to puzzle embedding
- All response types need to show the actual response text in the alert area.  Right now, correct answers (and possibly also incorrect/others?) don't show the full response text and the only place to view it is in the list of previous submissions, which is no longer open by default.  
- List of previous submissions doesn't handle HTML.Raw submissions and just serves straight text.

jbodner09:
Closes #856.  Does the following:

- Shows the system's response for ALL submissions (not just partials) in the immediate display area below the submission box.  
- Correctly displays HTML for responses that use Html.Raw
- Fixes ridiculous whitespace in this file
- Correctly clears the submission box when an answer is submitted

HTML example:
image


ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -700,20 +698,43 @@ else
                     // Response Code enum is in Helpers/SubmissionEvaluator
                     const messageArea = document.querySelector("#initialMessage");
                     let updateVal = "";
+                    let responseString = (responseData.completeResponse ?? "") + (responseData.freeformResponse ?? "");
+                    if (responseString.startsWith("Html.Raw")) {
```
tabascq, line 702: Can you use RawHtmlHelper.Display() here instead?

ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -700,20 +698,43 @@ else
                     // Response Code enum is in Helpers/SubmissionEvaluator
                     const messageArea = document.querySelector("#initialMessage");
                     let updateVal = "";
+                    let responseString = (responseData.completeResponse ?? "") + (responseData.freeformResponse ?? "");
+                    if (responseString.startsWith("Html.Raw")) { 
+                        responseString = responseString.substring(9, responseString.length - 1);
```
digit01Wave, line 703: I assume you want to take the whole response string but remove the first "Html.Raw" bit.

If so, don't you want the below since the second parameter of the substring method is the length of the substring?
int lenToRemove = "Html.Raw".Length;
int substringLen = responseString.Length - lenToRemove;
responseString.substring(lenToRemove, substringLen)

tabascq: HtmlRawHelper.Display() is better, it addresses other cases as well.

jbodner09: @digit01Wave this is Javascript's substring, which uses endIndex instead of length as the second parameter.  substr is the one that uses length, and it's deprecated per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

ServerCore/Pages/Submissions/Index.cshtml
tabascq: Makes sense, but please move this logic to a new helper function in RawHtmlHelper so that we don't lose this location if we ever change/expand the format. RawHtmlHelper.AsString perhaps?

[side note: I don't understand the ASP format at all lol]

jbodner09: Can't do that either.  This data is coming from Javascript after the page is already loaded, so it has no concept of anything ASP.  The only time we can use ASP in Javascript is for creating static strings on page load.

tabascq: Understood, but please at least leave a note in RawHtmlHelper.cs pointing at this other usage.

ServerCore/Pages/Submissions/Index.cshtml
digit01Wave: Ohh. Makes sense. My bad. My mind was stuck in C# land

digit01Wave (approved):
lgtm

tabascq (approved):

PR #897, opened 2023-08-22
Add HasIndividualRegistration

asyasky:
Add concept of HasIndividualRegistration to Event object and adjust navigation flow based on it. Events that have individual registration will push players through that registration flow, events that don't will go back to the old flow of going directly to join/create team. This should also fix the previous event access bug.


Data/DataModel/PuzzleUser.cs
```diff
@@ -123,13 +123,19 @@ public async Task<bool> IsPlayerOnTeam(PuzzleServerContext dbContext, Event this
...
         /// <param name="thisEvent">The event that's being checked</param>
         public async Task<bool> IsRegisteredForEvent(PuzzleServerContext dbContext, Event thisEvent)
         {
-            return await dbContext.PlayerInEvent.Where(p => p.PlayerId == ID && p.EventId == thisEvent.ID).AnyAsync();
+            if(thisEvent.HasIndividualRegistration)
+            {
+                return await dbContext.PlayerInEvent.Where(p => p.PlayerId == ID && p.EventId == thisEvent.ID).AnyAsync();
+            }
+
+            return await dbContext.TeamMembers.Where(tm => tm.Member.ID == ID && tm.Team.Event.ID == thisEvent.ID).AnyAsync();
```
digit01Wave, line 138: [nit]: Can technically do .AnyAsync(tm => tm.Member.ID == ID ...) instead of the Where clause

ServerCore/Pages/Teams/List.cshtml.cs
```diff
@@ -32,7 +32,7 @@ public async Task<IActionResult> OnGetAsync()
                 return Challenge();
             }
 
-            if (!(await LoggedInUser.IsRegisteredForEvent(_context, Event)))
+            if (Event.HasIndividualRegistration && !(await LoggedInUser.IsRegisteredForEvent(_context, Event)))
```
digit01Wave, line 35: Does a similar check need to happen in Team/Apply.cs?
image

Right now, it looks like if the player is not registered, they won't be able to apply for a team and will get redirected to register as a player

asyasky:
Implemented in 2025

Issue #829, opened 2023-06-28
Things still missing from SinglePlayerPuzzles


digit01Wave:
- [DONE] Hints
- [DONE] Feedback
- [DONE] Remove from top if none available
- [DONE] Allow rename (not all SinglePlayerPuzzles are necessarily Pregame puzzles) 
- [DONE] Allow unlocking of puzzle for a specific player (instead of just for yourself)
- [DONE] Fastest solves page should at least show pregame puzzles for total solved (see josh's comment)

jbodner09:
If an event has no single player puzzles, the link in the top bar should be hidden, especially since it causes it to expand to 2 lines before you've registered for a team.

PR #900, opened 2023-08-23
Add SinglePlayerPuzzleHintCountPerPlayer model

digit01Wave:
#829 
We need this model to keep track of hint coins per player.


tabascq:
Why not just add these fields to the PlayerInEvent table?

tabascq:
Hmm, I don't know. TBH I'm not sure how hint coins will be earned for single player puzzles. If they're earned by solving other single player puzzles or set by an event property, then I understand your model. If they're handed out to players via some other mechanism, that mechanism would probably use PlayerInEvent fir the lookup.

Which works better for how you're thinking about it?

digit01Wave:
> Hmm, I don't know. TBH I'm not sure how hint coins will be earned for single player puzzles. If they're earned by solving other single player puzzles or set by an event property, then I understand your model. If they're handed out to players via some other mechanism, that mechanism would probably use PlayerInEvent fir the lookup.
> 
> Which works better for how you're thinking about it?

You are right. I am abandoning this PR and replacing it with https://github.com/PuzzleServer/mainpuzzleserver/pull/905

PR #908, opened 2023-08-26
Allow admins and authors to see fastest solves when not registered

digit01Wave:
Allow admins and authors to see fastest solves even when they are not registered for the event

### DIFF
--- a/ServerCore/Pages/Events/FastestSolves.cshtml.cs
+++ b/ServerCore/Pages/Events/FastestSolves.cshtml.cs
@@ -45,11 +45,12 @@ namespace ServerCore.Pages.Events
             this.StateFilter = stateFilter;
             this.CurrentTeam = (await UserEventHelper.GetTeamForPlayer(_context, Event, LoggedInUser));
 
-            await this.updateTeamPuzzleStats();
-            await this.updateSinglePlayerPuzzleStats();
+            HashSet<int> authorPuzzleIds = this.GetAuthorPuzzleIds();
+            await this.updateTeamPuzzleStats(authorPuzzleIds);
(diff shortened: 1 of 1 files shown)


ServerCore/Pages/Events/FastestSolves.cshtml.cs
```diff
@@ -52,7 +52,7 @@ public async Task OnGetAsync(TeamPuzzleSortOrder? teamPuzzleSort, SinglePlayerPu
         private async Task updateTeamPuzzleStats()
         {
             bool isRegisteredUser = await this.IsRegisteredUser();
-            if (!isRegisteredUser)
+            if (EventRole == EventRole.play && !isRegisteredUser)
```
morganbr, line 55: For authors, this would need to filter the puzzles to the author's puzzles

digit01Wave: Shouldn't authors see the fastest solves even if they didn't write the puzzle?
After all, players can see the fastest solves of all puzzles

tabascq: As a player, the list of puzzles shown is filtered to what you have unlocked, so I agree with Morgan.

digit01Wave: Is there a reason we don't want authors to see the fastest solves page of all the puzzles?

tabascq: Exactly

digit01Wave: Makes sense. I updated it

tabascq (approved):

PR #911, opened 2023-10-03
Single player puzzle hints

digit01Wave:
#829 Add hints to single player puzzles

### DIFF
--- a/ServerCore/Pages/Puzzles/SinglePlayerPuzzleHints.cshtml.cs
+++ b/ServerCore/Pages/Puzzles/SinglePlayerPuzzleHints.cshtml.cs
@@ -91,12 +91,36 @@ namespace ServerCore.Pages.Puzzles
 
         private async Task<IList<HintWithState>> GetHints(int puzzleID)
         {
-            Hints = await (from Hint hint in _context.Hints
+            List<HintWithNullableState> hintsWithNullableStates = await (from Hint hint in _context.Hints
                            where hint.Puzzle.ID == puzzleID
-                           join SinglePlayerPuzzleHintStatePerPlayer state in _context.SinglePlayerPuzzleHintStatePerPlayer on hint.Id equals state.HintID
-                           where state.PlayerID == LoggedInUser.ID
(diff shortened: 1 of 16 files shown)


tabascq:
Looks good, though I worry that we're going to have an explosion of records since the number of records per event is (#players * #hints) even if players don't look at the hints. I think it would be better if the default is to have no records, then add records as players take hints.

digit01Wave:
> Looks good, though I worry that we're going to have an explosion of records since the number of records per event is (#players * #hints) even if players don't look at the hints. I think it would be better if the default is to have no records, then add records as players take hints.

I have updated the code to only create the HintStatePerPlayer when the player needs to actually view the hints

ServerCore/Pages/Puzzles/SinglePlayerPuzzleHints.cshtml.cs
```diff
@@ -91,12 +91,36 @@ public async Task<IActionResult> OnPostUnlockAsync(int hintID, int puzzleId)
 
         private async Task<IList<HintWithState>> GetHints(int puzzleID)
         {
-            Hints = await (from Hint hint in _context.Hints
+            List<HintWithNullableState> hintsWithNullableStates = await (from Hint hint in _context.Hints
                            where hint.Puzzle.ID == puzzleID
-                           join SinglePlayerPuzzleHintStatePerPlayer state in _context.SinglePlayerPuzzleHintStatePerPlayer on hint.Id equals state.HintID
-                           where state.PlayerID == LoggedInUser.ID
+                           join SinglePlayerPuzzleHintStatePerPlayer state in _context.SinglePlayerPuzzleHintStatePerPlayer on hint.Id equals state.HintID into leftJoinedTable
```
tabascq, line 96: I've never written a left join and would be happier if @morganbr signs off on this part. The rest looks good though!

morganbr: This looks ok, but usually we eagerly create state when a puzzle or player gets created rather than needing to lazily fix it up later like this

morganbr (approved):
It would be nice to avoid having a duplicated hints page, but if it's not reasonable reconcile this and the existing hints page, this is otherwise good.

tabascq:
Deployed!

Issue #914, opened 2023-10-04
Event page details items are cut off when too long


digit01Wave:
![image](https://github.com/PuzzleServer/mainpuzzleserver/assets/4121745/e3f32a24-e791-4257-9a38-9e431d834813)

PR #915, opened 2023-10-04
Updated details pages to be tables instead

digit01Wave:
#914

Before:
image

After:
image


digit01Wave:
Abandoning because dd/dt is much more accessible

PR #916, opened 2023-11-07
Remove phone number fields if event is remote

tabascq:
If the event is remote, the phone numbers could be anywhere in the world and we aren't going to call them. Let's not even bother to collect.

Also hiding the team room from all pages when the event is remote, not just the team creation page.


tabascq:
Can I please get a quick PR on this? It's actually blocking the announcement of an event that Dana and I are going to host on the server (and you can play in it!)

morganbr (approved):

PR #929, opened 2023-12-20
Embedded Puzzle Page Tweaks

tabascq:
- When hints are used, show link to Hints page instead of link to email
- Find iframe by ID rather than by tag (important in case a past response has an iframe in it)

### DIFF
--- a/ServerCore/Pages/Submissions/Index.cshtml
+++ b/ServerCore/Pages/Submissions/Index.cshtml
@@ -461,11 +461,11 @@ else
             {
                 if (Model.Puzzle.IsForSinglePlayer)
                 {
-                    <div><b>Need help?</b> <a asp-page="/Puzzles/SinglePlayerPuzzleHints" asp-route-puzzleid="Model.Puzzle.ID">Click this link</a> to access hints for this puzzle.</div>
+                    <div><b>Need help?</b> <a asp-page="/Puzzles/SinglePlayerPuzzleHints" asp-route-puzzleid="Model.Puzzle.ID" target="_blank">Click this link</a> to access hints for this puzzle.</div>
                 }
                 else
                 {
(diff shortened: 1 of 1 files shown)


ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -457,7 +457,18 @@ else
             }
         </div>
         <div class="inlinewrapper">
-            @if (!Model.PuzzleState.IsEmailOnlyMode)
+            @if (!Model.Event.HideHints)
+            {
+                if (Model.Puzzle.IsForSinglePlayer)
+                {
+                    <div><b>Need help?</b> <a asp-page="/Puzzles/SinglePlayerPuzzleHints" asp-route-puzzleid="Model.Puzzle.ID">Click this link</a> to access hints for thsi puzzle.</div>
```
jbodner09, line 464: Typo:  thsi -> this in both links

Also, I'd prefer `target="_blank"` be added to both these links as well so they open in new tabs and don't potentially erase progress.

ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -804,7 +815,7 @@ else
                 // Resize iframe so the frame has no scrollbars
                 // By the time the window has loaded, the iframe contents should have as well, thus no iframe load listener
                 // Note that when the puzzle receives the message, it should remove any extraneous wrapping before reporting its size
-                const iframe = document.querySelector("iframe");
+                const iframe = document.getElementById("puzzle-frame");
```
jbodner09, line 818: While you're making edits, querySelector is the preferred method for selection, so you could just change the `iframe` to `#puzzle-frame`.

jbodner09 (changes requested):

jbodner09 (approved):

PR #932, opened 2024-01-10
More bug fixes

### LINKED ISSUE #275: Clean up encodings page
* Clean up / table of contents
* Decoding shortcuts as opposed to this list
    * Tree for morse
    * Semaphore shortcuts
* Base 3 is more common than base 8
    * 2, 3, and 10 are the common ones
    * In base 3, given three digits

### LINKED ISSUE #715: Feature request:  a "select all" button to check all the boxes for all the material files for a puzzle
When re-uploading puzzles that have a ton of images, it's very tedious to have to check every box on the existing versions of the files to delete them.  I can't just use the "Delete All" button either because other materials (like the solution) are also uploaded and I don't want to get rid of them.  Having a way to check all boxes so I can then just uncheck the few that I don't want to delete would be super helpful.

### LINKED ISSUE #832: Print embedded puzzles directly
Too risky for right now and less necessary for this year's puzzles, but it should probably be possible to print a puzzle without needing to click a new link.

### LINKED ISSUE #858: "Players in Event" admin page improvements
- Column headers are wrong.  "Event" should be "Player" and "Player" should be "Email".  
- "Team" column needs to be added for easy "which team is this player on?" mapping?  
- Default sorting should be by player name
- Even better, columns should be sortable

### LINKED ISSUE #875: LunchReportDate isn't initialized to a proper date

### LINKED ISSUE #876: CSS files are not applied to pages in a testable manner
_Layout.cshtml only adds the link to the minified css files for sites that we manually allow it to.  Unfortunately, our way of allowing that is based on the URL, and the URL of the site is configurable.  This means that we can actually break the site's layout just by changing its URL.  It also means it's impossible to test our CSS changes in production, since we also use the URL for purposes of making the site externally visible.  It would probably be better to apply the CSS file based on the site's homepath, since that's also where it finds the CSS files to begin with.  That would then allow us to see exactly how the site will look without making it visible externally.

jbodner09:
- Fixed the Player in Event page which wasn't displaying the name of certain properties properly
- Fixed the Players in Event page to have proper column headers, include the player's team name, and be sorted by player name (closes #858)
- Added default assets for the Alpha site (including a copy of puzzlejs) for easier testing of future site changes (see next bullet)
- Changed site asset lookup to be based on home path instead of url so changes can be tested in development (closes #876)
- Properly initialized Lunch Report Date when creating an event (closes #875)  
- Adds a Select All button to the materials section of a puzzle's file management page (closes #715)
- Make embedded puzzles printed directly from the submission page display similar to (but not exactly like) the unembedded version (closes #832)  
- Overhauls the encoding page (closes #275)

The encoding page now has lists organized similar to the encoding cards, where each letter has multiple encodings listed beside it, as opposed to each encoding type having its own separate image.  Many new encodings were added, including ASL, Caesar cipher, roman numerals, RGB color mixing, etc.  Layout was changed to list multiple things side-by-side so the page isn't as long.  

image
image
image
image
image
image


ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -181,6 +181,31 @@
     #img-frame {
         width: 100%;
     }
+
+    @@media print {
+        .no-print {
```
tabascq, line 186: Why not use online-only like the other places?

tabascq (approved):

ServerCore/Pages/Submissions/Index.cshtml
jbodner09: I'm trying to create a separation between the CSS that the puzzles use and the CSS that the site uses, since I know that we've been using online-only for the puzzles.  In fact, I probably should change the CSS I added in _Layout to be no-print as well just to keep with that convention, although the semantics really only matter if other people continue to follow them, which luckily doesn't matter TOO much since it's just naming.

tabascq: Of the three options described in this conversation, I think the one currently in the PR is the worst choice. 😃 I have a slight preference for using the same class everywhere, but your alternative of using one class exclusively for puzzle content and another exclusively for server content works too.

tabascq (commented):
I've been thinking about this more and I'm less certain that puzzle.js is ready to include. Rationale is that I expect to be doing further work to enable new puzzle types for a while, often mid-event to support puzzles in that event, and when I do that it'll be unreasonable to continually deploy additions to the server codebase and update links for all puzzles in the event etc.

I mean, if we started to use puzzle.js for example puzzles that are also part of this repo instead of in the events themselves, sure this is absolutely the way to do it. But I don't see that here.

@jbodner09 , thoughts?

jbodner09:
A couple!  

The first is about static versioning vs. evergreen.  This is basically the same issue that Edge's WebView deals with:  some people always want the latest version, others hard-depend on a specific version.  For the specific version, this works totally fine.  An event can just say "I'm using this version", and the only downside is that we start to collect multiple copies of the library of different versions, instead of only having a single copy that's always up to date.  It really just depends on if we NEED to use the later version once we decide what kind of puzzles are being written, which is obviously not a question we can answer until we get there.  

The second is for deploying the library itself.  While it certainly is nice that the relative link to the library that we used last year worked for both local testing and in production without any changes, I really wasn't happy that you had to log into the Azure account itself to update those files.  Deploying the website honestly isn't much better, but it's at least a little more auditable.  My preference would be to have the resources live in a "Resources" puzzle so that _anybody_ could update and access those files as necessary, but I couldn't think of a clean way to do that that wouldn't break links when migrating files between events like we do for alpha -> beta -> etc.  We also ran into this issue with the font file for the cartoon puzzles, so I think this is just a broader issue we need to solve.  

Overall, I actually don't plan to use this copy of the library for any particular event at the moment.  This is more about having it available in case other events CAN use it in this form, or we do want to update the site's example puzzles to use it (which we totally should do btw).

tabascq:
Point 1 reply: Nothing about our current approach uses a single version of puzzle.js for all events; it uses a per-event version.

Point 2 reply: How about we solve this problem, and others, by adding a feature to upload event resources with basically the same UX as puzzle materials, but with the folder obfuscation step removed? People could use it for whatever purpose they desire, be it puzzle.js, font files, event logos for events that don't deploy their Home/FAQ/Rules content to the site, etc. This approach lets an event manage its own needs without exposing its intentions to the tech team, and also lets the webification repo work perfectly well standalone even while those resources are under active development.

jbodner09:
I think this is exactly what we need.  It'll also help better separate resources that puzzles use that are currently being deployed with the site.  Like, if you look in the wwwroot folder, the css files for each event contain both the base- and print- files that the puzzles use, as well as the css (and minified version) that that year's homepage uses.  I've always hated how updating the puzzle template required redeploying the site, so this will definitely help enforce the "only resources for the site are checked into the mainpuzzleserver repo and deployed with the site" rule that I think we should have.  I'll open an issue for it.

tabascq:
OK, so if you pull puzzle.js from this, I'll merge it.

jbodner09:
@tabascq done!

Issue #934, opened 2024-01-16
Feature request:  file upload without obfuscated raw link


jbodner09:
Currently, uploading files to the azure blob storage hides them behind an obfuscated link.  This makes it difficult to re-use files between events, and in the worst case forces us to upload resources for _puzzles_ to the repo that hosts the _website_.  For example, the css for styling puzzleday puzzles lives in the same folder as the css that styles the puzzleday homepage, which requires redeploying the entire website any time we need to make changes to puzzle styling.  

Even worse, when the puzzles are migrated to the nipd event, now they are all pointing to the outdated resources from the intern event, which in the worst case may no longer have permission to be accessed by the new event, causing all puzzles to need to be updated.  One possible solution for this (as discussed in #932) is to create an event-agnostic resource upload page on the website that will not hide uploaded files behind obfuscated links.  This will allow those links to be valid across any and all events.  

Besides the above use case of migrating puzzles between events, this will also help us no longer need to upload puzzle resources to the website repo, which I believe is very important since that means puzzle authors or creators won't need access to the website repo.  It also will help for hosting resources like Kenny's puzzlejs library, which is designed to be used by multiple events (barring versioning requirements).

tabascq:
Handled by #936

PR #936, opened 2024-01-27
Shared Resource UX

tabascq:
Allow an event to manage its shared resources in a folder on the Azure storage explicitly named "resources". This means that an event can drop any number of dependencies in that folder (CSS files, event logos, puzzle.js, etc) while keeping the URLs stable and not needing to use Azure Storage Explorer.

### DIFF
--- a/ServerCore/FileManager.cs
+++ b/ServerCore/FileManager.cs
@@ -24,10 +24,11 @@ namespace ServerCore
         /// <param name="fileName">Name of the file to upload. The caller should ensure it is safe to use in a URL</param>
         /// <param name="eventId">Event the file will be used in</param>
         /// <param name="contents">Contents of the file</param>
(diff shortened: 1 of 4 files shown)


tabascq:
Fixes #934

jbodner09 (approved):
Perfect v1.  I think for the future, we'll need to think about the file management UI here, since in theory, this resources folder will continue to increase in contents, similar to the wwwroot's css folders.  Implementing a collapsable tree view instead of a flat list will likely help once we get multiple years' worth of stuff in here and want to delete an entire folder at a time, although that's all just purely for our convenience.

tabascq:
This storage is per-event, so I don't think the tree view will be necessary. Per-event is very much by design - no fair peeking at what other events are cooking up!

PR #945, opened 2024-02-11
Allow for resizable embedded puzzles (server side)

tabascq:
Just allows for multiple messages to come from the iframe. Client-side support can go in the event webification repo.


tabascq:
I'm thinking we can probably use ResizeObserver from within resize.js, but it'll be easier to experiment once this is deployed.

Issue #913, opened 2023-10-04
Website revamp


jbodner09:
Current design is obsolete and boring.  Can we make the site look more "modern" and "seamless" like GPH?  
Potential problem:  how can we make it look nice, but also generic enough for ALL events?  (We can't just make a sweeping change that breaks everything else from the past...)

jbodner09:
Related:  #869 (for bootstrap update) and #870 (since it may inform the design)

jbodner09:
Checklist for style updates after bootstrap merge:
- printing directly isn't working due to iframe sizing.  Remove height property on https://github.com/PuzzleServer/mainpuzzleserver/blob/main/ServerCore/Pages/Submissions/Index.cshtml#L204
- Title in header doesn't need to be that big (18px was old one's size)

jbodner09:
Relevant info from old issues that still needs to be addressed:
- Site accessibility pass needs to happen to make sure that screen readers work as expected, Accessibility Insights and F12 tool don't flag any issues, etc. 
- Design is still the same as before the bootstrap update and could still be prettier.  Also, work to improve each event's landing page in #870 should be consistent with other work to make the rest of the site look better.  
- Mobile usability pass also needs to happen.  Current mobile bugs include:
 - s

PR #946, opened 2024-02-12
Event customization

jbodner09:
Note: turn off whitespace differences to review this PR

Goal: support custom theming for an event that doesn't require re-deploying the website or changing any files in this repo.  

Associated work items: #954, #913,  [Wiki page with how-to for creating future events](https://github.com/PuzzleServer/mainpuzzleserver/wiki/Event-Customization)

This PR does the following, in order of files changed:

1. Adds a property per-puzzle for the name of a custom css file that will be applied to the answer submission page for that puzzle when puzzle embedding is enabled.  The rules for that file are noted in red above the field, and general styling advice is provided in the wiki linked above.
2. Deprecates the Partial files for customizing event homepages, rules pages, and FAQ pages.  
- Because those are not being used for any future events, the HomePartial field in the event database object is now also deprecated and shouldn't be set for any future events, although old events will still depend on it and the Partials.  
- Now, event customization will use `home-content.html`, `rules-content.html`, and `faq-content.html` files uploaded to the Shared Resources and shown in automatically-sized iframes, similar to embedded puzzles.
- The event database field for showing content is still supported and said content will be shown above the iframes, but it is not recommended to use those (old logic was to show this field, or a Partial, but not both).
3. Deprecates the wwwroot folder use for event-specific style assets.
- Previously, event-specific css files were applied to every page in the event based on their path in the wwwroot folder and a matching event HomePartial field.  Note that this actually used to be based on the UrlString, which was considered a bug in #876 and partially fixed in #932.  This now fully fixes it for old events. 
- Now, events apply a `global-styles.css` file from the Shared Resources, which is intended to be a companion to the puzzle-specific css files.
- Also adds custom classes ending in `-customizable` for targeted changes to things like puzzle titles that were previously pseudo-supported via Html.Raw, all of which are listed on the wiki.  No css files in this repo use those classes for anything.  

Fixes the following bugs:
- Properly initializes puzzle properties in the database that previously weren't initialized due to an oversight
- Fixes a Cancel link on the Player event registration edit page that leads back incorrectly
- Tweaked the footer placement so it's not styled when the submission page is.  
- Fixes more style bugs like incorrect checkboxes from the bootstrap update
- Disables client-side form validation in development environments to work around a bug in the current jquery validation library that doesn't allow localhost urls, which prevents local testing of file uploads for things like testing puzzle embedding


tabascq:
I think custom per-event CSS outside the repo is a great idea - but I also think it should be an event property, not a puzzle property, and apply to all pages rather than just puzzle pages. Thoughts?

jbodner09:
After discussion at the meeting, this looks like a great opportunity to go "yes, and".  

First, we need to really stop using the wwwroot folder, full stop.  Now that we have the per-event puzzle-agnostic resource file repository from #936, everything to style the website should go in there.  So, from this first commit, change the resource location from the wwwroot folder to the Azure storage Resources folder.  That way, css file changes don't need an entire website redeploy!  

Second, we're going to change the structure of the other stuff that's required in order to deploy a new event.  For example, #788 to add a new PD event needs new partial pages in Pages/Resources/<eventUrl>.  Those partials should be changed to just be a barebones wrapper that looks for a file with a specific name in the Azure storage Resources folder, and if not, shows an appropriate blank message.  

Currently, the wrapper looks for content in the event properties.  Those properties can't be removed from the database, but we can remove them from the event management page so future events don't use them, and then we just look for a resource file instead.  That way, changing site content no longer needs site deployment if you don't want to use the event properties, although like the event properties, we lose any asp benefits since we're using plain html.  

On the plus side, this also no longer requires us to remember to add the minified css building to bundleconfig.json either.  Then, any other css files to style the entire event will also be uploaded to that Azure storage Resources folder (alongside the per-puzzle css files already envisioned here, the resource files that the puzzles themselves reference for styling (which could overlap with the files used by the host per-puzzle page), etc.  

That means that the following css file structure should be used moving forward (obviously this is just a recommendation, and naming could be improved):  
- The main wwwroot/css/site.css that affects every page on every event and should be updated rarely (all other css in the wwwroot folder will serve past events only)
- The Azure storage Resources event_name.css that affects every page for a single event
- The Azure storage Resources puzzle-styles.css and puzzle-print-styles.css that all puzzles in an event use for their own internal styling
- The optional named css files added here that can be referenced by individual pages in an event and/or individual puzzles (since both kinds of pages are now served from Azure storage, they could in theory use the same files, like if certain puzzles need a "futuristic" style, there could be a futuristic.css with classes used by both the host submission page and the puzzle itself).

Finally, the rest of the pages on the site should have `-customizable` css classes added so other things can be overridden like the answer submission page added.  A few more bugs can also be fixed, like:
- make navbar text unselectable
- add hover to errata and solved icons to tell users what that means
- make the checkbox and errata a span so they can be customizable
- fix unreadable save button text on individual puzzle management page

ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -459,16 +459,30 @@
...
+    {
+        // Add class to body wrapper here so only this page's content gets it, not every page
+        <script type="text/javascript">
+            document.addEventListener("DOMContentLoaded", () => {
+                let bodyWrapper = document.querySelector(".body-content");
+                bodyWrapper.classList.add("body-content-customizable");
+            });
+        </script>
+        cssFileName += (!cssFileName.EndsWith(".css")) ? ".css" : "";
+        <link rel="stylesheet" href="~/css/@(Model.Event.HomePartial)/@cssFileName" />
```
morganbr, line 473: Is there a requirement to load from this server? If authors could upload css as materials, this could be self-service (particularly useful for module hunt)

ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -459,16 +459,30 @@
...
+    @if (cssFileName.Length > 0)
+    {
+        // Add class to body wrapper here so only this page's content gets it, not every page
+        <script type="text/javascript">
+            document.addEventListener("DOMContentLoaded", () => {
+                let bodyWrapper = document.querySelector(".body-content");
+                bodyWrapper.classList.add("body-content-customizable");
+            });
+        </script>
+        cssFileName += (!cssFileName.EndsWith(".css")) ? ".css" : "";
```
morganbr, line 472: Does this work around something? I think it's reasonable to require correct file extensions

ServerCore/Pages/Submissions/Index.cshtml
jbodner09: As implied by my huge comment above, this is going to change to indeed load from the resources folder for all future events so that absolutely nothing for future events will come from the wwwroot folder.

ServerCore/Pages/Submissions/Index.cshtml
jbodner09: This is just dummy proofing.  All you need to supply is a file name, so forcing users to remember to also add the extension seems unnecessarily complicated.  Plus, the worst that will happen if you get it wrong is that is just won't load because it won't be able to find the file.

ServerCore/Pages/Submissions/Index.cshtml
tabascq: I think it's quite reasonable/desirable to allow for custom CSS per puzzle, so that a puzzle in a module event can change whatever it likes if desired. No problem if the default is to go to the resources folder, so long as any puzzle that wants to do something custom can put that CSS file in its materials and provide a URL of, say, "../mypuzzle/mycustom.css".

jbodner09: I think we're getting our wires crossed here a little bit.  CSS theming for the puzzle itself will ideally always go alongside the puzzle materials that are uploaded per-puzzle and be addressed relative to the puzzle file, assuming it's uploaded correctly.  This PR won't change anything about that since we're already able to do exactly that.  

This PR is specifically for theming the _host_ page so that the entire thing can be themed, and not just inside the puzzle's iframe, in case that's something that's desired.  In this case, you _could_ use the same file to theme both the host page and the puzzle itself, since the puzzle should _also_ have access to everything in the resources folder, whereas the host page _only_ has access to the resources folder.  

The way I'm envisioning this being used this year is that each year's puzzle(s) will have a different look, so each host page will need a different theme depending on the puzzle.  This definitely will be useful for module hunts too; I just want to make sure we're clear on the boundary between files for the website itself, and files for a puzzle.  There's currently nothing _enforcing_ that boundary, which is why if we get it confused now, it will only get worse over time.

tabascq: For module events with no true admin team, a puzzle may want to modify top-level CSS for their puzzle. They shouldn't have to use global resources because they may not have access.

I don't think this is the most common use case, but might as well light it up along with the rest.

jbodner09: Ah, I see what you're saying, you want the host page to be able to load css from more than JUST the resources folder.  That makes sense, I hadn't thought about the fact that only admins (and authors?) can access that.  That seems reasonable, so how about this:  if you ONLY supply a file name, it will be assumed to live in the resources folder, but if you supply a full URL (relative should work too, so basically anything with a / in it will trigger this logic) it will instead link to there.  And of course the tooltip will tell you all of this.

tabascq: If it's evaluated relative to the resources folder, and "../mypuzzle/mycss.css" works, that's sufficient IMO (assuming it's documented).

ServerCore/Pages/EventSpecific/FAQ.cshtml
```diff
@@ -5,6 +5,27 @@
...
+        <partial name="../Resources/@(Event.HomePartial ?? "Default")/FAQPartial.cshtml" />
+    }
+}
+else
+{
+    <environment include="Development">
+        <iframe id="faq-frame" style="width:100%;height:100%;overflow-x:scroll;overflow-y:visible;" title="FAQ Content" sandbox="allow-same-origin allow-scripts allow-popups" src="http://127.0.0.1:10000/devstoreaccount1/evt@(Model.Event.ID)/resources/faq-content.html"></iframe>
+    </environment>
+    <environment exclude="Development">
+        <iframe id="faq-frame" style="width:100%;height:100%;overflow-x:scroll;overflow-y:visible;" title="FAQ Content" sandbox="allow-same-origin allow-scripts allow-popups" src="https://puzzleserverteststore.blob.core.windows.net/evt@(Model.Event.ID)/resources/faq-content.html"></iframe>
```
tabascq, line 29: I don't think I understand the sizing here. Doesn't this need some kind of resize.js logic to size properly?

ServerCore/wwwroot/lib/jquery-validation-1.19.2-dist/jquery.validate.js
```diff
@@ -1426,11 +1426,15 @@ $.extend( $.validator, {
 		// https://jqueryvalidation.org/url-method/
 		url: function( value, element ) {
 
+			/* Included:
```
tabascq, line 1429: ??? Why is this in the PR?

ServerCore/Pages/EventSpecific/FAQ.cshtml
```diff
@@ -5,6 +5,27 @@
...
     ViewData["PlayRoute"] = "/EventSpecific/FAQ";
+    var eventPage = Model as ServerCore.ModelBases.EventSpecificPageModel;
+    var Event = eventPage?.Event;
 }
 
-<partial name="../Resources/@(Model.Event.HomePartial ?? "Default")/FAQPartial.cshtml" />
\ No newline at end of file
+@if ((Event.EventBegin.Year < 2024) || string.IsNullOrEmpty(Event.HomePartial))
+{
+    if (!string.IsNullOrEmpty(Event.FAQContent))
```
tabascq, line 14: Why not support FAQContent whenever it is specified? I happen to know of a 2024 event that is using it today. :-) Ditto the other two pages of course.

ServerCore/Pages/Events/Edit.cshtml
```diff
@@ -240,21 +240,24 @@
...
-                <label asp-for="EditableEvent.FAQContent" class="control-label"></label>
-                <textarea asp-for="EditableEvent.FAQContent" class="form-control"></textarea>
-                <span asp-validation-for="EditableEvent.FAQContent" class="text-danger"></span>
-            </div>
-            <div class="form-group">
-                <label asp-for="EditableEvent.RulesContent" class="control-label"></label>
-                <textarea asp-for="EditableEvent.RulesContent" class="form-control"></textarea>
-                <span asp-validation-for="EditableEvent.RulesContent" class="text-danger"></span>
-            </div>
+            @if (Model.Event.EventBegin.Year < 2024)
```
tabascq, line 243: Can we leave these in for a bit to allow for proper migration to happen?

ServerCore/Pages/Resources/Default/HomePartial.cshtml
```diff
@@ -1,24 +1,10 @@
-﻿@{
-    var eventPage = Model as ServerCore.ModelBases.EventSpecificPageModel;
-    var Event = eventPage?.Event;
+﻿<h2>@Model?.Event?.Name homepage TBD</h2>
+@if (Model?.Event?.EventRole == ModelBases.EventRole.admin)
+{
+    <p><b>Since you're an admin:</b> Upload a file named home-content.html to the <a href="https://puzzlehunt.azurewebsites.net/@(Model.Event.UrlString)/admin/Puzzles/ResourceManagement">shared resources folder</a> for this event. Don't forget the rules and FAQ, and for good measure, set the event's HomePartial field to be the same as its UrlString so this content is permanently hidden.</p>
```
tabascq, line 4: This message should clearly provide all filenames:

- faq-content.html
- home-content-html
- rules-content-html
- global-styles.css

tabascq: Also, new event creation should really provide these files by default instead of relying on this magic string property that acts like a bool. Then we could deprecate the HomePartial property as well.

tabascq (changes requested):

ServerCore/Pages/EventSpecific/FAQ.cshtml
jbodner09: If we want it to look as nice as the embedded puzzle do, yes.  Initially I wasn't sure if we did though, so I can add that.  It does add a slight pre-requisite to the embedded files, which I was trying to avoid, but if you're fine with that, especially since I will cover the need for the pre-requisite in the wiki page I'm writing, then I have no problem doing it. 

The only question I have is:  are you okay with this living in the wwwroot folder for these pages to use since we don't have event-agnostic file management via Azure storage?  We actually already have a size.js file that's being included but doesn't appear to have anything in it that we actually use, so I'd just add the code there.

ServerCore/wwwroot/lib/jquery-validation-1.19.2-dist/jquery.validate.js
jbodner09: See last bullet point in PR description.  I'm surprised you never hit this issue with your local development!

ServerCore/Pages/EventSpecific/FAQ.cshtml
jbodner09: If we still plan on using that field we can; I just thought it was no more or less difficult to add code there than it is to a separate file you'll upload and that we didn't need both ways of doing things.  The question is:  what hierarchy do we want?  I wasn't really a fan of this field completely overwriting the partials to begin with, so we should decide right now:  do we ever only show one set of content, or can we have both the database field AND the custom html file show up at the same time?  

We should also decide how the partials will be handled.  I know you also commented on one of the partial files below, but I'll continue this conversation here in case future changes to the partial files mess with your comment, as I know github sometimes does :)  I believe that we no longer need partials at all and that we should stop using them for all future events immediately, including rubix, ph24, and pd2024.  They require re-deployment of the site for new events, which this PR is explicitly trying to stop.  

Keeping in mind that the partials themselves for old events are the ones that handle including the event database field, and that these default files are only ever shown to admins who will otherwise have access to the wiki I'm writing anyway, I propose the following:
- Remove the check for the HomePartial field.  This will mean old events will always get the partials, and new events won't at all.  
- Move the event database field into the else block so new events get both it and the html file.  If both are missing, you just get no content, which is totally fine since that default content was just a failsafe anyway.  

Note that we're still using the HomePartial field for old events to also apply the style files located in the wwwroot folder, so we'll likely never be able to get rid of that field, but we now no longer will need to set it at all for any future events since no new events will have partial files associated with them.

ServerCore/Pages/Resources/Default/HomePartial.cshtml
jbodner09: Given that I don't believe we need these files at all moving forward, I'm closing this conversation.  See the one above as to why.

ServerCore/wwwroot/lib/jquery-validation-1.19.2-dist/jquery.validate.js
tabascq: I haven't run into this. Can we add this to the top of Puzzles/Edit/OnPostAsync instead? I'm just terrified of changing libraries that we may choose to update in the future.

```
            if (Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT") == Environments.Development)
            {
                ModelState.ClearValidationState("Puzzle.CustomURL");
                ModelState.ClearValidationState("Puzzle.CustomSolutionURL");
            }
```

ServerCore/Startup.cs
```diff
@@ -59,6 +59,12 @@ public void ConfigureServices(IServiceCollection services)
             {
                 options.Conventions.AuthorizeFolder("/Pages");
                 options.Conventions.AuthorizeFolder("/ModelBases");
+            }).AddViewOptions(options =>
+            {
+                if (Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT") == Environments.Development)
```
tabascq, line 64: Please add a comment here documenting exactly why this is in place :-)

tabascq: (to be clear, I fully support this approach over modifying jQuery!!! I just want us to remember why this is needed)

ServerCore/Pages/EventSpecific/FAQ.cshtml
```diff
@@ -5,6 +5,46 @@
...
+        window.addEventListener("message", (e) => {
+            if ((e.data.width > 0) && (e.data.height > 0)) {
+                const iframe = document.querySelector("#faq-frame");
+                iframe.classList.remove("unsized");
+                iframe.width = e.data.width + 30;
+                iframe.height = e.data.height + 30;
+            }
+        });
+    </script>
+    <environment include="Development">
```
tabascq, line 44: Suggest revising this so that it gets the path from FileManager.GetPuzzleDirectoryAsync (which would of course need to become public). This would make it more stable in the case that, say, the local storage path structure changes in the dev environment (since we don't really control that prefix).

ServerCore/Pages/Events/Standings.cshtml
```diff
@@ -11,70 +11,70 @@
 <h2>Standings</h2>
 
 <table class="table table-condensed">
```
tabascq, line 13: Were there any changes in this file?

ServerCore/Pages/Player/Create.cshtml
```diff
@@ -9,15 +9,15 @@
```
tabascq, line 9: Were there any changes here (other than whitespace)?

ServerCore/Pages/Resources/Default/FAQPartial.cshtml
```diff
@@ -1,4 +1,4 @@
-﻿@{
+﻿﻿@{
```
tabascq, line 1: Nit: This file and the next two can be removed

ServerCore/Pages/Shared/_Layout.cshtml
```diff
@@ -66,11 +66,25 @@
...
         if ((ev?.HomePartial != null) && ((ev.HomePartial.StartsWith("pd")) || (ev.HomePartial.StartsWith("nipd"))))
         {
-            <link rel="stylesheet" href="~/css/@(ev.UrlString)/styles.min.css" />
+            if (ev.EventBegin.Year < 2024)
+            {
+                <link rel="stylesheet" href="~/css/@(ev.HomePartial)/styles.min.css" />
+            }
+            else
+            {
+                <environment include="Development">
```
tabascq, line 81: Same FileManager trick here

tabascq: Also, I assume we want this stylesheet set up even if the event does not begin with pd or nipd...

ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -459,16 +459,44 @@
...
+                cssFileName = puzzleFilePath.Substring(0, puzzleFilePath.LastIndexOf('/')) + cssFileName.Substring(1);
+            }
+            <link rel="stylesheet" href="@cssFileName" />
+        }
+        else
+        {
+            // Otherwise, assume that bare file names are in the shared resources folder
+            // Note that the url for development may be different on your machine
+            <environment include="Development">
+                <link rel="stylesheet" href="http://127.0.0.1:10000/devstoreaccount1/evt@(Model.Event.ID)/resources/@cssFileName" />
```
tabascq, line 482: FileManager trick

tabascq (changes requested):

ServerCore/Pages/Events/Standings.cshtml
jbodner09: Just whitespace

ServerCore/Pages/Player/Create.cshtml
jbodner09: Nope

ServerCore/Pages/Resources/Default/FAQPartial.cshtml
jbodner09: Done

ServerCore/Pages/Shared/_Layout.cshtml
jbodner09: To the second question, no, this was the original logic if you look through the file history.  We've only ever applied styles like this to puzzleday and only have style files in wwwroot for pd and nipd, so that is technically not a change in this PR.

ServerCore/Pages/EventSpecific/FAQ.cshtml
jbodner09: Done for all hard-coded urls.

ServerCore/FileManager.cs
```diff
@@ -138,7 +138,7 @@ public static async Task DeleteBlobAsync(Uri fileUri)
         /// </summary>
         /// <param name="eventId">Event the directory should be associated with</param>
         /// <param name="puzzleDirectoryName">Name of a subdirectory to concatenate. Use empty string for none.</param>
-        private static async Task<CloudBlobDirectory> GetPuzzleDirectoryAsync(int eventId, string puzzleDirectoryName)
+        public static async Task<CloudBlobDirectory> GetPuzzleDirectoryAsync(int eventId, string puzzleDirectoryName)
```
tabascq, line 141: OK I know that I suggested that you make this function public, but I forgot that it would leak CloudBlobDirectory out of this file. I think it's better to keep all blob know-how inside FileManager; can you please change this back to private and make a new public method that returns the (absolute) Uri, which is all you care about anyway?

tabascq:
LGTM other than the one new comment, and also please merge with main and rebuild the migration part since Jessica's change also had migrations in it.

tabascq (approved):

Issue #912, opened 2023-10-04
Send hint and help requests in site instead of puzzhunt


digit01Wave:
Right now, help requests gets sent to puzzhunt email.
This makes it difficult for authors who aren't Microsoft employees because they don't have access to the puzzhunt email.
It also makes it difficult for players because now they have to monitor their email as well.

Would be nice if it shows up on the puzzle's page (or at least linked on the puzzle page)

digit01Wave:
GPH site: https://github.com/galacticpuzzlehunt/gph-site

PR #948, opened 2024-02-19
Add help message to database

digit01Wave:
First PR for #912 to add help messages to database so that we can allow users to directly ask for help for puzzles on the site (as opposed to sending an email).

This PR adds the new HelpMessage data model to the database.

Version 2 is at https://github.com/PuzzleServer/mainpuzzleserver/pull/955


Data/DataModel/HelpMessage.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// </summary>
+        public string Text { get; set; }
+
+        /// <summary>
+        /// True if message was from the player (as opposed to a response from puzzle control).
+        /// </summary>
+        public bool IsPlayerMessage { get; set; }
+
+        /// <summary>
+        /// The team ID of the sender (will be empty if not a player of if it is for a SinglePlayerPuzzle).
```
tabascq, line 43: If it could be empty then the value should be nullable (i.e. int?, not int).

Also, I think that (puzzle x team) or (puzzle x individual) are required scopes for a puzzle message, so I think calling these the "sender" is too limiting. For example, when puzzle authors respond to a message on a team puzzle, the team ID needs to be filled out.

Data/DataModel/HelpMessage.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        /// <summary>
+        /// The team of the sender (will be empty if not a player or if it is for a SinglePlayerPuzzle).
+        /// </summary>
+        public virtual Team Team { get; set; }
+
+        /// <summary>
+        /// The id of the sender.
+        /// </summary>
+        public int SenderID { get; set; }
```
tabascq, line 55: Ditto my comment above, and in addition I think that when authors reply, we should not tell the player who has replied. No problem reporting which player has posted a team message though.

Data/DataModel/HelpMessage.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        /// <summary>
+        /// The user that sent the message.
+        /// </summary>
+        public virtual PuzzleUser Sender { get; set; }
+    }
```
tabascq, line 61: We need a way to mark a message as claimed by support staff (and specifically whom, though the player should never see that info). Probably another person field necessary for that.

Data/DataModel/HelpMessage.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// <summary>
+        /// The date time (in UTC) of when the message was sent.
+        /// </summary>
+        public DateTime DateTimeInUtc { get; set; }
+
+        /// <summary>
+        /// The puzzle id that this message is for.
+        /// </summary>
+        [Required]
+        public int PuzzleID { get; set; }
```
tabascq, line 24: Are there use cases where there is no puzzle? E.g. 2-way communications about pizza etc.

morganbr: If so, should that have a different data model?

morganbr (changes requested):
This looks like great progress, added comments on a few issues

Data/DataModel/HelpMessage.cs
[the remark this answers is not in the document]
tabascq: perhaps pizza/help me find the light switch/etc...

digit01Wave: We will likely not tell the player who replied in the UI (will show up just as puzzle control) but it might be nice information to have on our side. It will also make it so that the model is generally the same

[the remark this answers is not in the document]
digit01Wave: GC is also a puzzle user

Data/DataModel/HelpMessage.cs
digit01Wave: I made a PuzzleMessage, which has a Puzzle, which will know if it is a singlePlayerPuzzle or not.
When looking at PuzzleMessages, we could join with team members to find messages viewable to the entire team.
This is so that we can put GameControl replies and player messages in the same database.

Or do you think it would be worth it to put them into three separate databases (PlayerMessage, TeamMessage, and GameControlMessages)?

Data/DataModel/HelpMessage.cs
digit01Wave: Added a IsMarkedAsRead property

Data/DataModel/HelpMessage.cs
digit01Wave: There are two separate sets in the puzzle context now

Data/DataModel/HelpMessage.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System;
+using System.ComponentModel.DataAnnotations;
+using System.ComponentModel.DataAnnotations.Schema;
+
+namespace ServerCore.DataModel
+{
+    /// <summary>
+    /// Message in help thread when a player asks for help.
```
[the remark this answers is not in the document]
digit01Wave: I think it would be a good idea to include the thread id just in case, so I added the property

Data/DataModel/HelpMessage.cs
digit01Wave: Actually, I changed my mind and made a teamMessage

Data/DataModel/HelpMessage.cs
digit01Wave: EventId, PuzzleId, and TeamID is there.
Do we really need staffId or playerId? Wouldn't the PuzzleUserId be enough? SinglePlayerPuzzles can be solved even if you aren't a registered player

Data/DataModel/Message.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// <summary>
+        /// Gets or sets the user that sent the message.
+        /// </summary>
+        public virtual PuzzleUser Sender { get; set; }
+
+        /// <summary>
+        /// Gets or sets the value indicating whether a user has marked the message as "read".
+        /// This will likely be used by game control to "claim" a question so no one else tries to answer it.
+        /// </summary>
+        public bool IsMarkedAsRead { get; set; }
```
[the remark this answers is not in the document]
digit01Wave: The properties have been added

Data/DataModel/Message.cs
```diff
@@ -0,0 +1,78 @@
...
```
morganbr, line 66: Is this only for GC since different members of a team might have read/not read it?

Data/DataModel/TeamMessage.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace ServerCore.DataModel
+{
+    /// <summary>
+    /// A message from a team.
+    /// </summary>
+    public class TeamMessage : Message
+    {
+        public TeamMessage()
+        {
+            this.IsFromGameControl = false;
```
morganbr, line 10: Nit: this shouldn't be necessary since bools default to false, but if you want, you could also set the property in the base class

Data/PuzzleServerContext.cs
```diff
@@ -90,6 +94,9 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             modelBuilder.Entity<Submission>().HasOne(submission => submission.Team).WithMany().OnDelete(DeleteBehavior.Restrict);
             modelBuilder.Entity<SinglePlayerPuzzleSubmission>().HasOne(submission => submission.Submitter).WithMany().OnDelete(DeleteBehavior.Restrict);
             modelBuilder.Entity<Annotation>().HasOne(annotation => annotation.Team).WithMany().OnDelete(DeleteBehavior.Restrict);
+            modelBuilder.Entity<Message>().HasOne(message => message.Event).WithMany().OnDelete(DeleteBehavior.Restrict);
```
morganbr, line 97: To confirm, does deleting a puzzle, team, and event work if there are messages?

morganbr (approved):

Data/DataModel/TeamMessage.cs
digit01Wave: That is true. I just wanted to set this explicitly

tabascq (approved):

PR #955, opened 2024-02-27
Add help message to database version2

digit01Wave:
This is a version where we make one table with nullable properties as the database type.
In code however, we only interact with the model types.

Version 1 is at https://github.com/PuzzleServer/mainpuzzleserver/pull/948


Data/PuzzleServerContext.cs
```diff
@@ -79,6 +80,7 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             modelBuilder.Entity<SinglePlayerPuzzleStatePerPlayer>().HasIndex(pspt => new { pspt.PlayerID, pspt.SolvedTime });
             modelBuilder.Entity<SinglePlayerPuzzleUnlockState>().HasKey(state => new { state.PuzzleID });
             modelBuilder.Entity<Room>().HasIndex(room => new { room.EventID, room.Building, room.Number }).IsUnique();
+            modelBuilder.Entity<Message>().HasIndex(message => new { message.ThreadId });
```
digit01Wave, line 83: @morganbr , I believe this is the spot you talked about if I wanted to add an index to speed up queries.

Are there allowed to be more than one index? I see some of the other items with multiple indexes as one object. Does that mean the index would only work as that kind of composite object? Or is the object considered a list of indexes?

Also, can indexes be null?

Data/DataModel/Message.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// <summary>
+        /// Gets or sets the id.
+        /// </summary>
+        [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
+        public int ID { get; set; }
+
+        /// <summary>
+        /// Gets or sets a thread id.
+        /// </summary>
+        public string ThreadId { get; set; }
```
tabascq, line 20: Do you expect that there will be multiple threads per (team-puzzle x team) or (single-player-puzzle x player) or (event x team) or (event x player)? If not, then you do not need this ThreadID. Personally I don't think it's necessary, but I won't quibble if you want it. If you do still want it, though, then I would suggest that you keep it an int, and have the ID be the ID of the start message (which, among other things, ensures that it's unique).

digit01Wave: I think this thread id will be good to have just in case we do need it.
It will also be useful for general non-puzzle related messages.

It makes sense from a "messaging" object.

Also, I plan to make the thread id for puzzle messages to be of the format "PuzzleMessage_<puzzleId>_<TeamId>"

PR #959, opened 2024-03-12
Add help message version2

digit01Wave:


ServerCore/Helpers/TeamHelper.cs
```diff
@@ -44,7 +45,12 @@ public static async Task DeleteTeamAsync(PuzzleServerContext context, Team team,
             var annotations = from annotation in context.Annotations
                               where annotation.Team == team
                               select annotation;
-            context.Annotations.RemoveRange(annotations);
+            context.Annotations.RemoveRange(annotations);
+
+            List<Message> messages = (from message in context.Messages
+                              where message.Team == team
+                              select message).ToList();
```
morganbr, line 52: Since this doesn't need the content of the messages, skipping ToList here is a rare case of being the right thing so RemoveRange can delete them, I think. It also looks like you might be able to use .ExecuteDelete() (which is newer than the other code here)

morganbr (approved):

PR #960, opened 2024-03-12
Resize iframes upon window resize

jbodner09:
Instead of hard-coding the iframe width, allow it to change as the window width changes, then update the iframe height in response.


tabascq (approved):
Thanks!!!

Issue #909, opened 2023-08-27
Tracking issue for "I'm working on this puzzle!"


jbodner09:
Since the website now has support for realtime updates via Blazor and tracking state for puzzles thanks to embedding, now would be a good time to work on the "I'm working on this puzzle!" button so that remote teams have an easy way to track who's working on what.

vidapour:
- tab is open marks puzzle as being 'worked on' by person X
- should have history of people who have opened tab, maybe have length of time each person was on? (might not be possible to show progress on webpage)

Vroo:
Suggest green circle, yellow triangle, gray square for active, idle for <5 minutes, idle for >5 minutes. Color + shape works well for colorblind people.

vidapour:
- cross btwn team status + google docs indicators

- team status indicator
- could track mouse/page/keyboard activity
- 'last active' state

- google docs contributor indicator
- use contributor initials to note where in the doc/puzzle they are

PR #961, opened 2024-03-12
Presence

### LINKED ISSUE #941: Handle/Prevent Blazor Disconnects
Blazor gets disconnected in background tabs and tells the user to reload. We should do one or both of:
1. Automatically reload like https://alldev.co.uk/reload-blazor-server-on-connection-down-in-net-6/
2. Prevent the background tab from sleeping by using something like a WebLock

morganbr:
Adds infrastructure for servers to push messages to each other and uses it to show "Presence" -- the set of your team members on each puzzle page.

SingalR will enable features where something happening for one user/author/admin should update pages for other users/authors/admins, such as push notifications, presence indications, and puzzle data syncing. This is composed of:

 *   A "Hub" -- the server side of SignalR, responsible for sending messages to clients
 *   A "Listener" -- the client side of SignalR, responsible for receiving messages
 *   Strongly-typed messages
 *   Integration with the Azure SignalR service. This connects each frontend instance's hub to all of the others' listeners so that every message is broadcast to all frontends. For local deployments and to avoid costs in the off-season, this is optional
 * A Presence store pattern that can illustrate how to handle other types of messages
 * A Blazor widget on the puzzle page that updates whenever a user's presence changes

The production instance has been configured to be compatible with this change and the scaling wiki page has been updated with instructions. Normal deployments have no impact from this change.

Fixes #941, #909


ServerCore/ServerMessages/ServerMessageListener.cs
```diff
@@ -0,0 +1,148 @@
...
+        /// Fires when any presence message comes in
+        /// </summary>
+        public event Func<PresenceMessage, Task> OnPresence;
+
+        private async Task OnPresenceMessageAsync(PresenceMessage message)
+        {
+            await OnPresence?.Invoke(message);
+        }
+
+        private async Task OnGetPressenceState(GetPresenceState requestMessage)
```
jbodner09, line 128: nit: typo in name

jbodner09 (changes requested):

ServerCore/Properties/ServiceDependencies/PuzzleServerTestDeploy - Web Deploy/profile.arm.json
```diff
@@ -0,0 +1,113 @@
+{
```
tabascq, line 1: Completely unqualified to review this file.

ServerCore/ServerMessages/PresenceStore.cs
```diff
@@ -0,0 +1,122 @@
...
+                teamPuzzleStore = teamStore.TeamPuzzleStores[puzzleId];
+            }
+
+            return teamPuzzleStore;
+        }
+
+        /// <summary>
+        /// Returns a list of all present users on puzzles
+        /// </summary>
+        public List<PresenceMessage> GetAllPresence()
```
tabascq, line 41: I'm having trouble walking through all the logic here, but I'm wondering if there's a way for there to be more presence messages acquired between the time that a call to GetAllPresence() starts and when the SignalR hub knows to broadcast new presence messages to the new server.

tabascq: Like, is there any sort of lock required, or is the idea that the listener is acquired before the list is sent, so there may just be some duplicates?

ServerCore/ServerMessages/ServerMessageHub.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// </summary>
+    public static class ServerMessageHubExtensions
+    {
+        /// <summary>
+        /// Group to always send messages to.
+        /// Don't send messages to all clients as we don't prevent other clients from connecting, but they can't get added to this group.
+        /// </summary>
+        const string ServersGroup = "servers";
+
+        public static async Task BroadcastExampleMessageAsync(this IHubContext<ServerMessageHub> hub, ExampleMessage message)
```
tabascq, line 24: Maybe remove this now?

tabascq (commented):
Wow, this is a lot! Good stuff though!

ServerCore/Pages/Components/PresenceComponent.razor
```diff
@@ -0,0 +1,20 @@
...
+@inject DataModel.PuzzleServerContext PuzzleServerContext
+@inject IMemoryCache MemoryCache
+@inject PresenceStore PresenceStore
+
+<div>
+    <span>
+        <b>Who&apos;s Here: </b>|
+        @foreach (PresenceModel presenceModel in PresentUsers)
+        {
+            <span>@presenceModel.Name |&nbsp;</span>
```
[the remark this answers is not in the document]
morganbr: That works, switched to it

ServerCore/Properties/ServiceDependencies/PuzzleServerTestDeploy - Web Deploy/profile.arm.json
morganbr: This is an autogenerated file for deployment (confirmed there are no secrets in it). I didn't mean to include it in this review, but if we get these checked in, those of us who deploy could avoid having to keep working around them. Is that ok?

ServerCore/ServerMessages/PresenceStore.cs
morganbr: The ConcurrentDictionaries are threadsafe, so they won't crash without locks. The question is whether a presence change received during this process and missed by enumeration (or later like while the reply is in transit) would cause a problem. I think the answer is mostly no -- by the time an instance can request presence, it's listening for presence messages, so the new instance doesn't need the existing instance to forward them. The case that I'm not sure how to handle (but hope is minimal) is this can cause reordering.

Imagine pageInstance 123 is present when server 2 spins up. It calls GetAllPresence on server 1, which adds pageInstance 123 to the list. Right after that, pageInstance 123 leaves and sends a message that it's gone. Both servers receive the message. Server 1 removes 123 from its PresenceStore. Server 2's PresenceStore is empty and nothing changes. Then Server 2 receives AllPresence from Server 1, which still has 123 on the list. Server 2 puts 123 in the Store and it never goes away.

The only ideas I have to fully prevent that would be either using a real database or figuring out a cross-machine locking protocol, but I'm not sure that's justified for the case above. What do you think?

morganbr:
> Also, did we ever end up adding anything to address the manual reloads needed after the browser sleeps the tabs? If not, https://learn.microsoft.com/en-us/aspnet/core/signalr/javascript-client?view=aspnetcore-6.0&tabs=visual-studio#bsleep would be a good quick addition to _Layout at line 70 where we're already adding Blazor-only things.

@jbodner09  Added the code, but I'm not sure how to test it. Do you know a way to force tab sleep?

ServerCore/Pages/Components/PresenceComponent.razor.cs
```diff
@@ -0,0 +1,123 @@
...
+        private async Task OnPresenceChange(IDictionary<Guid, PresenceModel> presentPages)
+        {
+            await UpdateModelAsync(presentPages);
+        }
+
+        private async Task UpdateModelAsync(IDictionary<Guid, PresenceModel> presentPages)
+        {
+            if (presentPages.Count > 0)
+            {
+                var deduplicatedUsers = (from model in presentPages.Values
```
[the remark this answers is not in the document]
morganbr: Removed the unnecessary ToArray. It's still a bit more overhead than necessary, but helpful if we put idle status back in

jbodner09:
You can force a tab to sleep from the about:discards page, but since we're adding that code specifically to prevent sleeping in the first place, it might be a better test to set the browser to sleep all tabs after 30 seconds, turn on the other aggressive power saving settings on that page (settings/system), put a big load on the browser with other tabs, and then verify on about:discards that it's NOT put to sleep when the rest of the tabs are.

morganbr:
> You can force a tab to sleep from the about:discards page, but since we're adding that code specifically to prevent sleeping in the first place, it might be a better test to set the browser to sleep all tabs after 30 seconds, turn on the other aggressive power saving settings on that page (settings/system), put a big load on the browser with other tabs, and then verify on about:discards that it's NOT put to sleep when the rest of the tabs are.

Thanks! Confirmed the Web Lock adds itself to the list of reasons the tab can't be discarded

tabascq (approved):

jbodner09 (approved):

PR #962, opened 2024-03-17
Add files to new events

jbodner09:
Finalizes new event file creation and structure by doing the following:

- Adds the basic homepage, faq, rules, and style files used in a particular event automatically upon event creation (or according to the provided checkbox for demo event creation).  The contents of these files live in strings in the new helper class, and can be changed or expanded as needed.  Right now, they match the content provided on https://github.com/PuzzleServer/mainpuzzleserver/wiki/Event-Customization, so it's recommended to keep those two in sync if updates are made to one of them.  The files are created via a new helper since there are two different ways to create an event, and creation should fail silently, for example when running locally and no storage emulator exists.  
- Fixes a potential crash in the PuzzleApiController due to improper use of the null conditional (?.) operator.  The operator only works if a property follows the ?.  Since a function is following in the current usage, we must manually check for null before operating on the string. 
- Fixes an incorrect application of a style file from #946.  That PR erroneously applied the global style file based on the presence of the deprecated-in-that-PR HomePartial field.  That's wrong; the global style file should ALWAYS be applied to new events (and consequently never to old ones).  
- Removes a cloud call from the API added in that PR since it turns out the functions it was calling were hitting the server but they didn't need to in order to just determine whether they were running on a local device or not, so a non-service call was substituted.  
- Hides an error introduced in that PR that the Azure Storage APIs return when requesting a file that doesn't exist if that file is for the custom homepage/rules/faq content.  Since that error is in the form of a fully-functional webpage that then becomes user visible in the iframe, the only way to hide it is to remove the iframe entirely if the expected resizing handshake doesn't occur, which it won't since that error page doesn't include the needed javascript file to do the handshake.  
- Refactored page content into a partial so that the logic for displaying all these embedded pages is only written once instead of copied into every file for the faq content, rules content, etc.  
- Cleans up variable usage on the main _Layout page since there were multiple variables doing the same job and nested codepaths that would never be called due to the value from one of those variables.
- Fixed some bugs from the original design in which models are referenced on certain pages.


ServerCore/Pages/Events/Create.cshtml.cs
```diff
@@ -70,6 +72,12 @@ public async Task<IActionResult> OnPostAsync()
 
             await _context.SaveChangesAsync();
 
+            // Create base files for event content and style
```
tabascq, line 75: Rather than have this code twice, can there just be one UploadEventFiles function in NewFileHelper?

ServerCore/Pages/Events/CreateDemo.cshtml.cs
```diff
@@ -482,6 +487,15 @@ public async Task<IActionResult> OnPostAsync()
                 transaction.Commit();
             }
 
+            // Create base files for event content and style
+            if (AddSkeletonContentFiles)
```
tabascq, line 491: If it's efficient to put a try/catch around the NewFileHelper stuff, maybe remove this flag?

ServerCore/Helpers/NewFileHelper.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System.IO;
+using System.Text;
+
+namespace ServerCore.Helpers
+{
+    public static class NewFileHelper
```
tabascq, line 6: How about NewEventFileHelper?

You could also consider making a base class between Create and CreateDemo, which contains this code, and then no helpers would be necessary.

ServerCore/Pages/Events/CreateDemo.cshtml.cs
jbodner09: This is the checkbox on the demo creation page that we said we wanted since demo events sometimes don't need these.

ServerCore/Helpers/NewFileHelper.cs
jbodner09: Can definitely rename.  Won't do the base class though since we decided that a helper was the way to go when we talked about it due to the demo event wanting these files to be optional.

ServerCore/Pages/Events/Create.cshtml.cs
jbodner09: Not sure which twice you're referring to.  If you mean having it duplicated between Create and CreateDemo, that's by design since according to our discussion, we wanted it to be always created for real events but only optionally for demo events.  If you mean having to call this function multiple times (once for each file), I can add a simple wrapper function to do all 4 at once, but that feels unnecessary.

ServerCore/Pages/Events/CreateDemo.cshtml.cs
tabascq: I would argue that demo events on the server always need these, and local events on a dev box don't because some people don't configure local storage. If it gracefully failed instead, that might be enough.

ServerCore/Pages/Events/Create.cshtml.cs
tabascq: I am referring to the fact that there's a lot of code duplicated for all four files, and if a fifth file appears, it'll be easy to forget one of the places. Having a single function that does all the work for both scenarios would be easier to maintain.

ServerCore/Helpers/NewFileHelper.cs
tabascq: Could still be a method on a base class, always called for non demo events, and optionally called for demo events (though I still wonder if exception handling can replace the use of a demo flag, but let's use the other comment for that discussion)

ServerCore/Pages/Events/Create.cshtml.cs
```diff
@@ -70,6 +72,12 @@ public async Task<IActionResult> OnPostAsync()
 
             await _context.SaveChangesAsync();
 
+            // Create base files for event content and style
+            await FileManager.UploadBlobAsync("global-styles.css", Event.ID, NewFileHelper.GetStreamContent(NewFileHelper.ContentType.GlobalCss), SharedResourceDirectoryName);
```
morganbr, line 76: Streams should be disposed after use

ServerCore/Pages/Events/Create.cshtml.cs
jbodner09: That makes sense from a future-proofing standpoint.  I can add a "new event files" wrapper function.

ServerCore/Pages/Events/CreateDemo.cshtml.cs
jbodner09: I feel like we never discussed local/dev events vs. on the server, so we're kind of getting them mixed up with demo events vs. non-demo events.  My understanding was that real/non-demo events will ALWAYS need these new files because on the server we're obviously going to use the event, and locally people would be creating a non-demo event specifically for local customization of its content.  If anything, they would create a demo event just to use the puzzles it creates, so they WOULDN'T need the homepage stuff for anything in a demo event, which is why we decided they should be optional when we talked about it.  

That's a good point though that in either case, if the event is local/dev, file creation could fail due to storage not being set up, so this will need to fail silently if that's the case.

ServerCore/Pages/Events/Create.cshtml.cs
jbodner09: MemoryStream documentation mentions in the Remarks section that it doesn't need explicit Dispose calls:  https://learn.microsoft.com/en-us/dotnet/api/system.io.memorystream?view=net-8.0

ServerCore/Pages/EventSpecific/FAQ.cshtml
```diff
@@ -38,6 +38,14 @@ else
             if ((iframe !== null) && (iframe !== undefined)) {
                 iframe.contentWindow.postMessage("", "*");
             }
+            setTimeout(() => {
+                const iframe = document.querySelector("#faq-frame");
+                if ((iframe !== null) && (iframe !== undefined)) {
+                    if (iframe.classList.contains("unsized")) {
+                        iframe.style.display = "none";
```
tabascq, line 45: Can we also display something explaining what likely happened and how to fix it? This will be very difficult to debug when something is wrong, especially if you are not a member of the tech team.

jbodner09: The old partials only displayed the message for admins and authors.  If we're okay with the message also displaying for players, then sure.

tabascq (approved):

Issue #973, opened 2024-04-16
Changing Puzzle from Single Player to Team (or vice versa) doesn't properly update associated state


jbodner09:
Specifically, `GetVisibleSinglePlayerPuzzleViews` in Play.cshtml.cs (and likely the equivalent `GetVisibleTeamPlayerPuzzleViews`) are returning errors when loading the list of Puzzles for a player after a puzzle's type has been changed.  

The safest thing to do would just be to wipe all the associated state for a puzzle from the `SinglePlayerPuzzleStatePerPlayer` or `PuzzleStatePerTeam` tables as well (depending on the direction of the switch) so that the puzzle no longer appears in that table if it's not that type of puzzle anymore.

PR #976, opened 2024-04-18
Fix single player puzzle update

digit01Wave:
#973 

Fixed issue when switching a puzzle between IsForSingleUser and a regular team puzzle.

### DIFF
--- a/ServerCore/Pages/Puzzles/Edit.cshtml.cs
+++ b/ServerCore/Pages/Puzzles/Edit.cshtml.cs
@@ -127,17 +127,29 @@ namespace ServerCore.Pages.Puzzles
                     if (oldPuzzleView.IsForSinglePlayer)
                     {
                         // If moving from SinglePlayerPuzzle to team puzzle, delete single player puzzle states
-                        await _context.SinglePlayerPuzzleHintStatePerPlayer.Where(hintState => hintState.Hint.PuzzleID == oldPuzzleView.ID).ExecuteDeleteAsync();
-                        await _context.SinglePlayerPuzzleStatePerPlayer.Where(puzzleState => puzzleState.PuzzleID == oldPuzzleView.ID).ExecuteDeleteAsync();
-                        await _context.SinglePlayerPuzzleSubmissions.Where(submissions => submissions.PuzzleID == oldPuzzleView.ID).ExecuteDeleteAsync();
-                        await _context.SinglePlayerPuzzleUnlockStates.Where(unlockState => unlockState.PuzzleID == oldPuzzleView.ID).ExecuteDeleteAsync();
+                        IEnumerable<SinglePlayerPuzzleHintStatePerPlayer> hintStatesToRemove = _context.SinglePlayerPuzzleHintStatePerPlayer.Where(hintState => hintState.Hint.PuzzleID == oldPuzzleView.ID);
+                        _context.SinglePlayerPuzzleHintStatePerPlayer.RemoveRange(hintStatesToRemove);
+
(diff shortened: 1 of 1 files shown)


ServerCore/Pages/Puzzles/Edit.cshtml.cs
```diff
@@ -120,6 +120,41 @@ public async Task<IActionResult> OnPostAsync(int puzzleId)
...
+                        _context.HintStatePerTeam.RemoveRange(hintStatesToRemove);
+
+                        IEnumerable<PuzzleStatePerTeam> puzzleStatesToRemove = _context.PuzzleStatePerTeam.Where(puzzleState => puzzleState.PuzzleID == oldPuzzleView.ID);
+                        _context.PuzzleStatePerTeam.RemoveRange(puzzleStatesToRemove);
+
+                        IEnumerable<Submission> submissionsToRemove = _context.Submissions.Where(submissions => submissions.PuzzleID == oldPuzzleView.ID);
+                        _context.Submissions.RemoveRange(submissionsToRemove);
+                    }
+
+                    await _context.SaveChangesAsync();
```
jbodner09, line 155: Is there any reason this new block shouldn't just come before the existing SaveChangesAsync, removing the need for a second database save?

ServerCore/Pages/Puzzles/Edit.cshtml.cs
```diff
@@ -120,6 +120,41 @@ public async Task<IActionResult> OnPostAsync(int puzzleId)
...
+                        IEnumerable<HintStatePerTeam> hintStatesToRemove = _context.HintStatePerTeam.Where(hintState => hintState.Hint.PuzzleID == oldPuzzleView.ID);
+                        _context.HintStatePerTeam.RemoveRange(hintStatesToRemove);
+
+                        IEnumerable<PuzzleStatePerTeam> puzzleStatesToRemove = _context.PuzzleStatePerTeam.Where(puzzleState => puzzleState.PuzzleID == oldPuzzleView.ID);
+                        _context.PuzzleStatePerTeam.RemoveRange(puzzleStatesToRemove);
+
+                        IEnumerable<Submission> submissionsToRemove = _context.Submissions.Where(submissions => submissions.PuzzleID == oldPuzzleView.ID);
+                        _context.Submissions.RemoveRange(submissionsToRemove);
+                    }
+
```
jbodner09, line 154: We also need to create the appropriate rows in the database for the new puzzle type.  If it's now a single player puzzle, SinglePlayerPuzzleUnlockStates needs to be created, and if it's now a team puzzle, PuzzleStatePerTeam needs to be created for every team.  Mirroring the logic in https://github.com/PuzzleServer/mainpuzzleserver/blob/main/ServerCore/Pages/Puzzles/Create.cshtml.cs#L90-L103 should be sufficient.

jbodner09 (changes requested):

ServerCore/Pages/Puzzles/Edit.cshtml.cs
tabascq: +1

tabascq: Or is this what new lines 218-228 are doing?

ServerCore/Pages/Puzzles/Edit.cshtml.cs
tabascq: Unless there's a reason why the new code should come first (and please document why if so), I'd just use the same if/then for both blocks of code (old line 163, new line 198) for a) maximum clarity and b) full atomicity of the DB change.

digit01Wave: Done!

ServerCore/Pages/Puzzles/Edit.cshtml.cs
digit01Wave: It is what those lines were doing

tabascq (approved):

jbodner09 (approved):

PR #977, opened 2024-04-19
Fix messages not appearing

digit01Wave:
This issue only occurs when the URL string is set.

The fix is to remove the offending eventId from the modelState.


tabascq (approved):
I don't remember why this is the thing to do but I do agree it's used elsewhere!

jbodner09 (approved):
Approving since it's a simple fix, although I still apparently don't know enough about mvc to know how it got there in the first place.

jbodner09:
Oh dang, sniped!

Issue #969, opened 2024-04-09
Message feature requests


digit01Wave:
- [DONE]Allow GC to filter to unclaimed messages only
- [DONE]Allow GC to filter to specific puzzle
- [DONE] Make chat boxes prettier
- [DONE]Respect event level flag that lets us hide help messages from GC if they are not the author
- [DONE] Don't allow users to post more messages after answers are released
- [DONE]Allow filter to only threads the author is explicitly on/supporting
- [DONE] Filter to unclaimed messages only
- [DONE] Live refresh

tabascq:
During an event, Game Control is going to want to easily filter this down to unclaimed messages only.

Also, ideally GC will be able to see this page update live without needing to refresh.

PR #981, opened 2024-04-26
Add refresh

digit01Wave:
#969 

This PR adds the following
- Adds shortcut to Puzzle help thread from "Jump to..."
image
- Make sure you can't edit or post new messages after answers have been released
- Introduce the ability to refresh the threads page for GCs
image

### DIFF
--- a/ServerCore/Pages/Threads/PuzzleThread.cshtml.cs
+++ b/ServerCore/Pages/Threads/PuzzleThread.cshtml.cs
@@ -379,7 +379,7 @@ namespace ServerCore.Pages.Threads
             }
             else if (messageText.Length > MessageCharacterLimit)
             {
-                return ValidationResult.CreateFailure("Your message should not be longer than 3000 characters.");
+                return ValidationResult.CreateFailure($"Your message should not be longer than {MessageCharacterLimit} characters.");
             }
             else
             {
--- a/ServerCore/Pages/Threads/PuzzleThread.cshtml.cs
+++ b/ServerCore/Pages/Threads/PuzzleThread.cshtml.cs
@@ -155,13 +155,15 @@ namespace ServerCore.Pages.Threads
 
         public async Task<IActionResult> OnPostAsync()
         {
-            if (string.IsNullOrWhiteSpace(NewMessage.Text))
+            if (Event.AreAnswersAvailableNow)
             {
-                ModelState.AddModelError("NewMessage.Text", "Your message cannot be empty.");
+                ModelState.AddModelError("NewMessage.Text", "Answers are already available.");
             }
-            else if (NewMessage.Text.Length > MessageCharacterLimit)
+
+            ValidationResult validationResult = IsMessageTextValid(NewMessage.Text);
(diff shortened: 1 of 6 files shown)


ServerCore/Pages/Threads/PuzzleThread.cshtml.cs
```diff
@@ -352,5 +370,38 @@ private async Task SendEmailNotifications(Message newMessage, Puzzle puzzle)
...
+
+        private ValidationResult IsMessageTextValid(string messageText)
+        {
+            if (string.IsNullOrWhiteSpace(messageText))
+            {
+                return ValidationResult.CreateFailure("Your message cannot be empty.");
+            }
+            else if (messageText.Length > MessageCharacterLimit)
+            {
+                return ValidationResult.CreateFailure("Your message should not be longer than 3000 characters.");
```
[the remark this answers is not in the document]
morganbr: If there isn't already one, you can set a limit on the class in Data, which will get enforced by client and server

ServerCore/Pages/Threads/PuzzleThread.cshtml.cs
```diff
@@ -155,13 +155,15 @@ public async Task<IActionResult> OnGetAsync(int? puzzleId, int? teamId, int? pla
 
         public async Task<IActionResult> OnPostAsync()
         {
-            if (string.IsNullOrWhiteSpace(NewMessage.Text))
+            if (Event.AreAnswersAvailableNow)
```
morganbr, line 158: Do we actually need this check? I'm not sure asking for help late is bad per se, it just might not get answered

ServerCore/Pages/Threads/PuzzleThread.cshtml.cs
```diff
@@ -352,5 +370,38 @@ private async Task SendEmailNotifications(Message newMessage, Puzzle puzzle)
                 MailHelper.Singleton.SendPlaintextWithoutBcc(recipients.Select(r => r.Email), emailTitle, emailContent);
             }
         }
+
+        private ValidationResult IsMessageTextValid(string messageText)
+        {
+            if (string.IsNullOrWhiteSpace(messageText))
```
morganbr, line 376: Putting RequiredAttribute on the property would prevent this and add client-side validation

ServerCore/Pages/Threads/PuzzleThread.cshtml.cs
digit01Wave: I just didn't want people continuing to add stuff to the server. 
Also, if the answers are available, would you need help?

morganbr (approved):

ServerCore/Pages/Threads/PuzzleThread.cshtml.cs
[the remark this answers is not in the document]
digit01Wave: the message remains. For example, this is what it looks like when I change the limit to 10
image

I have added maxlength to the client side, but someone can just adjust the html to get around it

PR #982, opened 2024-04-28
Free agent system

### LINKED ISSUE #721: Feature request:  free agent system
This will make the flow from creating an account in our system to signing up for an event to finding a team to play on much smoother.

morganbr:
Implements new registration flows for players who don't know what team they want to be on and unifies it with other kinds of registration. Includes:
* Adding a Blazor signup hub page where you choose between creating a team, joining an open team, and having the system assign you a team.
* Adding the concept of "auto teams" -- teams created by the system on players' behalf, separated by expertise level.
* Blazor-ifying the team creation page to get client-side and server-side validation
* Linking to the new flow from everywhere
* Exposing stats to admins

Screenshots:
image
image
image

Fixes #721


ServerCore/Pages/Teams/Signup.cshtml.cs
```diff
@@ -0,0 +1,51 @@
...
+                return RedirectToPage("/Teams/Details", new { teamId = existingTeamId });
+            }
+
+            if (LoggedInUser.Email.EndsWith("@microsoft.com"))
+            {
+                IsMicrosoft = true;
+            }
+            else if (!string.IsNullOrWhiteSpace(LoggedInUser.EmployeeAlias))
+            {
+                IsMicrosoft = true;
```
tabascq, line 45: Oh this is not even close to true. People put all sorts of things in there.

ServerCore/Pages/Teams/AutoTeam.razor.cs
```diff
@@ -0,0 +1,148 @@
...
+        public async Task OnSubmit()
+        {
+            CantCreateTeam = false;
+
+            int existingAutoTeam = await (from team in _context.Teams
+                                           join tempTeamMember in _context.TeamMembers on team.ID equals tempTeamMember.Team.ID into teamMembers
+                                           from teamMember in teamMembers.DefaultIfEmpty()
+                                           where team.Event == Event && team.AutoTeamType == PlayerType
+                                           group teamMember by team.ID into teamGroup
+                                           where teamGroup.Count() < Event.MaxTeamSize
```
tabascq, line 44: We should leave some extra space here, so that autoplayers can add +1s to their team, and so the last autoteam of like 2 people has somewhere they can be merged. This could either be some hardcoded number like 2, or an event property.

ServerCore/Pages/Teams/Signup.cshtml.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        public async Task<IActionResult> OnGetAsync()
+        {
+            if (LoggedInUser == null)
+            {
+                return Challenge();
+            }
+
+            if ((EventRole != EventRole.play && EventRole != EventRole.admin)
```
tabascq, line 26: Why do admins sign up?

ServerCore/Pages/Teams/Signup.cshtml.cs
tabascq: I know it's beyond the scope of this PR, but I really don't think we can trust that EmployeeAlias is valid without either doing auth or sending a verification email. I wouldn't use random user-typed garbage for automatic validation of team constraints.

tabascq: That last part *is* within the scope of this PR though.

tabascq (changes requested):
Exciting!!!

ServerCore/Pages/Teams/AutoTeam.razor.cs
```diff
@@ -0,0 +1,148 @@
...
+            {
+                CantCreateTeam = true;
+                return;
+            }
+
+            Team newTeam = await UserEventHelper.GetTeamForPlayer(_context, Event, LoggedInUserId);
+            NavigationManager.NavigateTo($"/{Event.EventID}/{EventRole}/Teams/{newTeam.ID}/Details", true);
+        }
+
+        private string[] adjectives = new string[]
```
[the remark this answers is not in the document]
jbodner09: It also is technically an infinite loop since it's not doing selection without replacement.  I'd consider the 3rd level of randomness a requirement.

ServerCore/Pages/Teams/CreateTeam.razor.cs
```diff
@@ -0,0 +1,131 @@
...
+            if (TeamHelper.IsTeamNameTaken(puzzleServerContext, teamModel.EventId, name))
+            {
+                return new ValidationResult("Another team has this name.");
+            }
+
+            return ValidationResult.Success;
+        }
+    }
+
+    public class TeamModel
```
jbodner09, line 55: Assuming this is supposed to mirror the same properties in the server-side Team data model, it would be nice to add a comment here and in that file saying so.  That way people know that if they ever modify one of the fields here, they need to do it there too, and vice versa.

ServerCore/Pages/Teams/Signup.cshtml.cs
jbodner09: Or why NOT authors?

ServerCore/Pages/Teams/CreateTeam.razor.cs
```diff
@@ -0,0 +1,131 @@
...
+        {
+            PuzzleServerContext puzzleServerContext = validationContext.GetRequiredService<PuzzleServerContext>();
+            TeamModel teamModel = (TeamModel)validationContext.ObjectInstance;
+            string name = value as string;
+            if (string.IsNullOrWhiteSpace(name))
+            {
+                return new ValidationResult("Team names cannot be left blank.");
+            }
+
+            if (TeamHelper.IsTeamNameTaken(puzzleServerContext, teamModel.EventId, name))
```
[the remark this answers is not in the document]
morganbr: ```Wait()``` deadlocks here and I couldn't find a way to prevent it. Normally ```.ConfigureAwait(false)``` would work around that kind of deadlock, but somehow Blazor validation gets stuck anyway. We could preload team names, but that risks a team being created with that name sometime after the page loads. I didn't think synchronous call would be too bad here since creating teams is a rare operation, but what do you think?

ServerCore/Pages/Teams/Signup.cshtml.cs
morganbr: Do we have better validation for this elsewhere that I should use? This doesn't exactly validate anything, it just gives the user access to the free agent system (based on what I think was your idea that if we limit it to FTEs/interns, we don't have to worry about auto teams not having enough FTEs/interns).

ServerCore/Helpers/TeamHelper.cs
```diff
@@ -191,7 +249,7 @@ string TruncateName(string name)
 
             string truncatedName = TruncateName(name);
 
-            List<string> existingNames = await (from t in context.Teams where t.EventID == eventObj.ID select t.Name).ToListAsync();
+            List<string> existingNames = (from t in context.Teams where t.EventID == eventId select t.Name).ToList();
```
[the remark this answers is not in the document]
morganbr: As you spotted below, this was to resolve the validation attribute deadlock issue. It shouldn't have a significant performance impact since it only happens when creating a team

ServerCore/Pages/Teams/AutoTeam.razor.cs
tabascq: We also discussed using MaxTeamSize * 0.2, rounded to the nearest integer, which is 2 for 12 person teams and 1 for 4 person teams.

ServerCore/Pages/Teams/Signup.cshtml.cs
tabascq: tldr: I think the else clause at line 43 is going to be abused and cause trouble, and should be removed.

ServerCore/Pages/Teams/CreateTeam.razor
```diff
@@ -0,0 +1,81 @@
...
+                <ValidationSummary />
+                <input type="hidden" @bind="TeamModel.EventId" />
+                <div class="form-group">
+                    <label>
+                        Team name
+                        <InputText @bind-Value="TeamModel.Name" class="form-control" />
+                    </label>
+                </div>
+                @* TODO: Handle puzzleday room assignment and check hunt/safari requirements *@
+                @if (!Event?.IsRemote == true)
```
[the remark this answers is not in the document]
morganbr: This really shouldn't accept a null event and I've removed the null coalescing. For reference though, the result of ```Event?.IsRemote``` is  Nullable<bool>, which isn't compatible with the ! operator (IsRemote could be null, true, or false).

ServerCore/Pages/Teams/Signup.cshtml.cs
morganbr: Admins create test teams or fix signup problems on site. Authors can't currently do those things

ServerCore/Pages/Teams/CreateTeam.razor.cs
morganbr: Done

ServerCore/Pages/Teams/AutoTeam.razor.cs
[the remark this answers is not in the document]
morganbr: Adding a color if there's a collision

ServerCore/Pages/Teams/Signup.cshtml.cs
morganbr: Switched to only allowing ```@microsoft.com```. When MaxExternalsPerTeam is the whole team, anyone is allowed to use this

ServerCore/Pages/Teams/AutoTeam.razor.cs
```diff
@@ -0,0 +1,148 @@
...
+                // Create a new auto team
+                Team team = new()
+                {
+                    Event = Event,
+                    AutoTeamType = PlayerType.Value,
+                    Name = name,
+                    PrimaryContactEmail = captain.Email,
+                };
+
+                await TeamHelper.CreateTeamAsync(_context, team, Event, LoggedInUserId, EventRole == EventRole.play);
```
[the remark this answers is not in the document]
morganbr: Added a player role check

ServerCore/Pages/Teams/AutoTeam.razor
```diff
@@ -0,0 +1,22 @@
+@using ServerCore.DataModel;
+
+<p>I&apos;m a...</p>
+<EditForm Model="this" OnValidSubmit="OnSubmit">
+    <DataAnnotationsValidator />
+    <ValidationSummary />
+
+    <InputRadioGroup Name="PlayerTypeSelector" @bind-Value="PlayerType">
```
tabascq, line 8: There was discussion in a recent PD meeting about perhaps altering this to incorporate how serious about the event the player is. Is that something we should incorporate this year or no?

ServerCore/Pages/Teams/CreateTeam.razor.cs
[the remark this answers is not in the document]
tabascq: Understood - can we have two versions then, one of which is synchronous and is used only in this attribute, and one which is async and used everywhere else? It should be straightforward to put the truncation-check logic into a private helper function used by both.

tabascq (commented):
Two reasonably straightforward comments left - ping when ready for final review!

ServerCore/Pages/Teams/AutoTeam.razor
morganbr: Done

ServerCore/Pages/Player/Create.cshtml.cs
```diff
@@ -28,7 +28,7 @@ public async Task<IActionResult> OnGet()
 
             if(player != null)
             {
-                return RedirectToPage("/Teams/List");
+                return RedirectToPage(Event.AllowBlazor ? "/Teams/Signup" : "/Teams/List");
```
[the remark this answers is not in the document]
morganbr: Done

ServerCore/Helpers/TeamHelper.cs
```diff
@@ -13,6 +14,63 @@ namespace ServerCore.Helpers
     /// </summary>
     static class TeamHelper
     {
+        public static async Task CreateTeamAsync(PuzzleServerContext context, Team team, Event ev, int userId, bool isPlay)
```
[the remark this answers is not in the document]
morganbr: Done

tabascq (approved):

PR #985, opened 2024-04-30
Updated chat UI

digit01Wave:
#969 
This PR does the following:
- Only allow users that made the message to edit or delete it
- Update chat UI
image


tabascq (approved):
Approved but needs to have conflicts resolved

PR #988, opened 2024-05-06
Fix Time 4 Realz

tabascq:
Not sure how exactly this happened (most likely is that I forgot a final build), but site.min.js only picked up half of the changes.


tabascq:
The thrilling sequel to #984 - I already deployed it and it seems to work now!

jbodner09 (approved):
Welp, better approve then!

PR #993, opened 2024-05-16
More site feedback

jbodner09:
- Fixes extra spacing below the puzzle title on the embedded submission page
- Fixes a crash when visiting the fastest solves page when logged out by requiring login
- Fixes a crash when navigating to home/faq/rules for an event that doesn't exist by throwing 404


ServerCore/Pages/Events/FastestSolves.cshtml.cs
```diff
@@ -202,8 +202,10 @@ await PuzzleStateHelper.GetSparseQuery(_context, this.Event, null, null)
 
         private async Task updateSinglePlayerPuzzleStats(HashSet<int> authorPuzzleIds)
         {
-            if (!Event.ShouldShowSinglePlayerPuzzles)
+            bool isRegisteredUser = await this.IsRegisteredUser();
+            if ((!Event.ShouldShowSinglePlayerPuzzles) || (EventRole == EventRole.play && !isRegisteredUser))
```
morganbr, line 206: This should be handled by auth

ServerCore/ModelBases/EventSpecificPageModel.cs
```diff
@@ -214,7 +214,7 @@ public async Task BindModelAsync(ModelBindingContext bindingContext)
         }
 
         public string GetFileStoragePrefix() {
-            return FileManager.GetFileStoragePrefix(Event.ID, "");
+            return FileManager.GetFileStoragePrefix((Event?.ID ?? 0), "");
```
morganbr, line 217: How did you get a null event on event specific pages? That shouldn't happen

ServerCore/Pages/Events/FastestSolves.cshtml.cs
jbodner09: I agree, so I can add an auth redirect.  I'm still going to leave this though since the team puzzle section has it and it seems reasonable that somebody needs to be registered for the event to view this data.

ServerCore/ModelBases/EventSpecificPageModel.cs
jbodner09: If you navigate to a page for an event that doesn't exist, like https://puzzlehunt.azurewebsites.net/pd2025/play due to a typo or something, the website crashes because the event object doesn't exist.  It's not a good look, so this at least lets it finish loading and shows our boilerplate which is actionable.  From a user standpoint, this is a much more friendly error state: 
image

morganbr (approved):
Thanks for helping fix the whole class of navigation bugs!

PR #994, opened 2024-05-22
Make food options an event property

jbodner09:
NOTE: Ignore whitespace for this review; Visual Studio needed to unify whitespace for one of the files

This PR makes it so changing food options no longer requires redeploying the website by doing the following: 

- Adding a "Lunch Options" field to the database that's a string formatted as "Option Name":"Option Description"; pairs.  Since each pair is delimited by semicolons, the event field can be nicely formatted with newlines if necessary, and other punctuation like commas can be used inside the quotes.  This description is also added to the event details editing page so people will know how to use it properly.  
- Fixing the Lunch link in the navbar to both properly know if somebody has already chosen their lunch for single player events (accomplished by fixing the HasSwag function in EventSpecificPageModel), as well as always appearing for events that have lunches regardless of whether it's a team lunch or an individual lunch.  In both cases, the link is formatted to call attention to itself if lunch selection is still open and/or necessary.  
- Fixing a crash when navigating to the Swag Registration page if Swag is enabled but Single Player Lunches are disabled for an event by changing the intentional crash (which is what the Forbid call was doing) to a redirect.  
- Unifying the logic used by the Swag Registration page, player creation page, and player editing page to select an individual lunch, as well as fixing the logic to properly show the custom text field only when a "none of the above" option is selected.  These pages also now pull from the new Event property for their choices instead of using hard-coded radio buttons like the Swag page or plain text boxes like the Player pages used to.  
- Generalizing the wording used on the Team details page for selecting a lunch. This also pulls from the new Event property instead of a hard-coded list of pizza options, although unlike the individual lunch pages, this one only shows the option names, not their descriptions as well.


ServerCore/ModelBases/EventSpecificPageModel.cs
```diff
@@ -129,7 +129,8 @@ public async Task<bool> IsEventAdmin()
         /// </summary>
         public async Task<bool> HasSwag()
         {
-            return Event.HasSwag && await _context.PlayerInEvent.Where(m => m.Event == Event && m.Player == LoggedInUser).AnyAsync();
+            PlayerInEvent playerSwag = await _context.PlayerInEvent.Where(m => m.Event == Event && m.Player == LoggedInUser).FirstOrDefaultAsync();
+            return Event.HasSwag && (playerSwag != null) && !string.IsNullOrWhiteSpace(playerSwag.Lunch);
```
tabascq, line 133: We can early exit w/o hitting the DB if (!Event.HasSwag) - which is arguably the most common case.

ServerCore/Pages/Player/Create.cshtml
```diff
@@ -43,16 +43,66 @@ else
...
+                    <label asp-for="PlayerInEvent.Lunch" class="control-label"></label><br />
+                    @{
+                        var lunchOptions = Model.Event.LunchOptions.Split(";");
+                        bool required = false;
+                        foreach (var lunch in lunchOptions)
+                        {
+                            if (!string.IsNullOrWhiteSpace(lunch))
+                            {
+                                string lunchVal = lunch.Split(":")[0];
+                                string lunchDesc = lunch.Split(":")[1];
```
tabascq, line 56: Micronit: Just split on ":" once

ServerCore/Pages/Player/Create.cshtml
```diff
@@ -43,16 +43,66 @@ else
...
-                        <span asp-validation-for="PlayerInEvent.Lunch" class="text-danger"></span>
-                    </div>
-                    <div class="form-group">
-                        <label asp-for="PlayerInEvent.LunchModifications" class="control-label"></label>
-                        <input asp-for="PlayerInEvent.LunchModifications" class="form-control" />
-                        <span asp-validation-for="PlayerInEvent.LunchModifications" class="text-danger"></span>
-                    </div>
+                <div class="form-group">
+                    <label asp-for="PlayerInEvent.Lunch" class="control-label"></label><br />
+                    @{
```
tabascq, line 48: Future: There should be a common component that is used on both pages here. I'm not super familiar with how to do that, so if you choose not to at least put a big comment at the top of this block and the one in Edit.cshtml so they don't diverge.

ServerCore/Pages/Teams/Details.cshtml
```diff
@@ -204,26 +204,26 @@
             @if (Model.Event.EventHasTeamSwag)
             {
                 <div class="row">
-                    <h2>Pizza</h2>
+                    <h2>Lunch</h2>
                     <p>
-                        Your team can currently have @Model.EditableLunches pizza(s). If your team wants to opt out of a pizza, select &quot;No Pizza&quot;. If you cannot have any of the options offered due to dietary restrictions, please email <a href="mailto:puzzhunt@microsoft.com">puzzhunt@microsoft.com</a> with details and we'll find an alternative.
+                        Your team can currently have @Model.EditableLunches lunch(s). If your team wants to opt out of one of the lunches, select &quot;No Lunch&quot;. If you cannot have any of the options offered due to dietary restrictions, please email <a href="mailto:puzzhunt@microsoft.com">puzzhunt@microsoft.com</a> with details and we'll find an alternative.
```
tabascq, line 209: The use of "lunch" in this section sounds more individual than shared, and I suggest this page use the word "item" to be more generic and avoid potential confusion.

ServerCore/Pages/Teams/Details.cshtml.cs
```diff
@@ -1,328 +1,331 @@
...
+                                 select lunch).ToListAsync();
+
+                double possibleInPersonMembers = Event.MaxTeamSize - remoteMembers;
+                SoftMaxLunches = (int)Math.Ceiling(possibleInPersonMembers / (double)PlayersPerLunch);
+            }
+            LunchOptions = Event.LunchOptions.Split(";");
+            for (int i = 0; i < LunchOptions.Length; i++) 
+            {
+                LunchOptions[i] = LunchOptions[i].Split(":")[0];
+                LunchOptions[i] = LunchOptions[i].Substring(LunchOptions[i].IndexOf("\"") + 1, LunchOptions[i].LastIndexOf("\"") - LunchOptions[i].IndexOf("\"") - 1);
```
tabascq, line 151: So this just throws the details away? If so then the details page should make it clear that the details are for single player events only.

ServerCore/Pages/Events/Edit.cshtml
```diff
@@ -120,6 +120,14 @@
                 <input asp-for="EditableEvent.DefaultLunch" class="form-control" />
                 <span asp-validation-for="EditableEvent.DefaultLunch" class="text-danger"></span>
             </div>
+            <div class="form-group">
+                <label asp-for="EditableEvent.LunchOptions" class="control-label"></label>
+                <textarea asp-for="EditableEvent.LunchOptions" class="form-control"></textarea>
+                <span asp-validation-for="EditableEvent.LunchOptions" class="text-danger"></span>
+                <div style="font-size:11px;">Each option is composed of a "Lunch Name":"Lunch Description"; pair, including the quotes. Each option is delimited by a ; semicolon, so whitespace outside of quotes is ignored. COLON AND SEMICOLON CANNOT BE USED INSIDE THE QUOTES.</div>
+                <div style="font-size:11px;">The Lunch Name is also used as the value that's written to the database. Lunch Name == "noneoftheabove" after removing spaces and converting to lowercase will make the "Custom Order" textbox appear on the Swag Registration page for events with single-player lunches.</div>
+            </div>
```
tabascq, line 129: Can you add a display of the current parse of LunchOptions here, so we can catch errors before players do?

ServerCore/Pages/Player/Create.cshtml
tabascq: Also, am I correct that we won't be exercising this code at all for PD2024 since lunch is a team lunch?

tabascq (changes requested):

ServerCore/Pages/Player/Create.cshtml
jbodner09: I was thinking the same thing since this code, the one on the player edit page, and the swag registration page are identical, so as long as I'm making changes, I'd might as well do this as well.  And while yes, we're not using it this year, the fact that this was dead code in the codebase before this change was something that needed addressed anyway.

jbodner09: Oh wait, partials apparently can't save data according to https://learn.microsoft.com/en-us/aspnet/core/mvc/views/partial?view=aspnetcore-7.0, so this is indeed just going to be relegated to the future.  Comment it is!

ServerCore/ModelBases/EventSpecificPageModel.cs
jbodner09: I just looked, and apparently this is only called once inside an if block that's triggered when the event does have swag, so that Event.HasSwag ends up being unnecessary since it will ALWAYS be true.

ServerCore/Pages/Events/Details.cshtml
```diff
@@ -133,6 +133,31 @@
...
+        </dt>
+        <dd>
+            @Model.Event.LunchOptions
+            <br />
+            <div style="margin-left:20px;font-size:11px">
+                @{
+                    // Parse lunch options the same way they are in Swag\Register.cshtml and Player\Create or Edit.cshtml
+                    // This will help ensure correct formatting without the need to test those pages as a player
+                    // If the parsing method in those 3 files is ever updated, this will need to be updated as well
+                    var lunchOptions = Model.Event.LunchOptions.Split(";");
```
tabascq, line 147: I bet this needs a null check.

tabascq: (for the new-event-without-lunch-options case)

ServerCore/Pages/Events/Details.cshtml
```diff
@@ -133,6 +133,31 @@
...
+                    foreach (var lunch in lunchOptions)
+                    {
+                        if (!string.IsNullOrWhiteSpace(lunch))
+                        {
+                            string[] lunchParts = lunch.Split(":");
+                            string lunchVal = lunchParts[0].Substring(lunchParts[0].IndexOf("\"") + 1, lunchParts[0].LastIndexOf("\"") - lunchParts[0].IndexOf("\"") - 1);
```
tabascq, line 153: Can you please confirm that malformed input doesn't take down the page? Thinking of cases where quotes are missing.

The internet tells me that this might do the trick more safely: `string lunchVal = lunchParts[0].Trim().Trim('"');` - trims whitespace from the edges, then trims quotes from the edges.

tabascq: If this works better, note that the logic would need to be updated in 5 places.

jbodner09: Yeah, that's definitely simpler if nothing else, so I like it!

tabascq (approved):

PR #996, opened 2024-05-25
Partial #993 rollback

tabascq:
This part of #993 broke puzzle details editing and probably several other things. Gotta find another way.


jbodner09 (approved):
The goal of this particular change was to swap from crashing the site when navigating to a non-existent page (like if you make a typo with the event name) to giving a 404.  Obviously that shouldn't affect pages that DO exist, so this is totally harmless to roll back.

PR #998, opened 2024-05-28
hide lunch

jbodner09:
As Morgan pointed out, lunch for teams in the navbar is showing even if they're not on a team yet.

### DIFF
--- a/ServerCore/Pages/Shared/_EventNavigationPartial.cshtml
+++ b/ServerCore/Pages/Shared/_EventNavigationPartial.cshtml
@@ -205,7 +205,7 @@
                             <li class="nav-item"><a style="color:yellow" asp-page="/Swag/Register" title="Lunch">Pick Lunch by @(Event.TeamMiscDataChangeEnd.ToString("M/dd"))!</a></li>
                         }
                     }
-                    else if (Event.EventHasTeamSwag)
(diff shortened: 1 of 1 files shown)


ServerCore/Pages/Shared/_EventNavigationPartial.cshtml
```diff
@@ -205,7 +205,7 @@
                             <li class="nav-item"><a style="color:yellow" asp-page="/Swag/Register" title="Lunch">Pick Lunch by @(Event.TeamMiscDataChangeEnd.ToString("M/dd"))!</a></li>
                         }
                     }
-                    else if (Event.EventHasTeamSwag)
+                    else if (Event.EventHasTeamSwag && isOnTeam)
```
morganbr, line 208: The block above should arguably have this check as well (if you sign up for lunch and aren't on a team, you're not getting lunch)

morganbr (approved):

ServerCore/Pages/Shared/_EventNavigationPartial.cshtml
tabascq: For individual lunches, I think it makes sense this way.

tabascq (approved):

PR #999, opened 2024-05-28
Partial fix for malformed url

jbodner09:
This is a followup to #993 and #996.  Checking for a null event object seems to be safe.  Also removing the role throw since role validation has plenty of other problems that still need followup even when the role is valid, and changing the standings page authorization to match the fastest solves one.


tabascq (approved):

ServerCore/Pages/Events/Standings.cshtml.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace ServerCore.Pages.Events
 {
-    [Authorize(Policy = "IsRegisteredForEvent")]
```
asyasky, line 13: I might vote to change the match to go the other direction, more restrictive instead of less. I don't have a particular reason for it, it just seems like defaulting to more restrictive is good practice.

PR #1000, opened 2024-05-28
Fixes from Bug Bash

### LINKED ISSUE #997: Site crashes instead of giving 404 when mistyping eventid or role part of the url
If the event and/or role parts of the url are mistyped so that they don't correspond to a valid event or role (https://puzzlehunt.azurewebsites.net/event/role/MessingUpAnythingAfterThisJustGivesA404LikeItShould), the site will crash instead of giving a 404 for most pages.  

Some of this was introduced with the new shared-resource-supplied homepages in #946, which now require a valid Event object in order to finish displaying certain content on the site (GetFileStoragePrefix in EventSpecificPageModel will crash for some pages, but the new EventSpecific home/rules/faq pages will also crash).  However, some of these crashes existed before then (puzzles page, team pages, and anything with an incorrect role, since an invalid role is hard-coded to throw an exception).  

At first, we thought we could fix some of these by just validating the event object and role in #993, but it turns out that there are valid cases where the event object is invalid, which was causing new crashes in situations where there shouldn't be any, so that was rolled back in #996.   It's also likely more complicated than that, since there is clearly not a just a single failure point depending on which page you're trying to view.  

Ideally, any url on puzzlehunt.azurewebsite.net that doesn't explicitly work should give a 404 instead of crashing the site.

jbodner09:
- Addresses feedback from #999 that came in after it was merged by making standings pages have a more restrictive auth check
- Closes #997 (in conjunction with #999) by properly checking role and forbidding if it's malformed or the user doesn't have that role for that event.  Note that this required changing the EventRole enum, but that appeared to have no side effects.  
- Cleans up whitespace on a few player pages
- Makes it so links to other websites from the iframe'd home/rules/faq files work properly
- Temporarily add team name length limits


ServerCore/ModelBases/EventSpecificPageModel.cs
```diff
@@ -39,6 +39,17 @@ public override async Task OnPageHandlerExecutionAsync(PageHandlerExecutingConte
                 return;
             }
 
+            // Check permissions now that the role has been binded
+            await IsEventAdmin();
```
tabascq, line 43: I don't think we want to do these checks on every page load even if the requested role is "play". Seems like a perf issue.

ServerCore/Pages/Puzzles/Play.cshtml
```diff
@@ -43,9 +43,18 @@
        asp-route-teamPuzzleSort="@Model.TeamPuzzleSort"
        asp-route-stateFilter=@(unsolvedFilter ? PlayModel.PuzzleStateFilter.All : PlayModel.PuzzleStateFilter.Unsolved)>
         Show @(unsolvedFilter ? "All" : "only unsolved") puzzles
-    </a> |
-    <a asp-page="/Threads/PuzzleThreads">View help threads</a>
+    </a> | 
+    <a asp-page="/Threads/PuzzleThreads">View help threads</a> |
```
tabascq, line 47: What is the importance of this change? Doesn't seem bug bash motivated so I'm wondering if it's an intentional addition.

ServerCore/ModelBases/EventSpecificPageModel.cs
jbodner09: These are cached, so I can make sure we always use the cached version if it exists, but we definitely need to actually check that the role is valid.  We haven't been doing that at all, which is a huge security hole.

ServerCore/Pages/Puzzles/Play.cshtml
jbodner09: This is general page cleanup that falls under the "make the site look better" feedback.

ServerCore/ModelBases/EventSpecificPageModel.cs
jbodner09: Oh actually those functions already don't do the database hit if the cached version already exists, so they're good.

tabascq: OK, agree that the role switcher will cache these values, I forgot about that case.

ServerCore/ModelBases/EventSpecificPageModel.cs
```diff
@@ -39,6 +39,17 @@ public override async Task OnPageHandlerExecutionAsync(PageHandlerExecutingConte
...
+            await IsEventAuthor();
```
tabascq, line 44: Now that I understand why these lines aren't a perf issue: If the LoggedInUser is null, then weird things will happen. Rather than using the static cache, I would just use the return value of these IsEventXxx functions, which will either use the cache or not but definitely be bools.

tabascq (changes requested):
One very simple thing

ServerCore/ModelBases/EventSpecificPageModel.cs
jbodner09: Unfortunately, not possible.  For reasons beyond my understanding, you can't call async functions in if blocks, so I'll just have to add a branch for the logged out case.

tabascq: not even

`bool isAuthor = await IsEventAuthor();`

???

tabascq (approved):

PR #1002, opened 2024-05-29
Auto add remove tweaks

tabascq:
- PrimaryEmail should contain all players for auto teams
- Deleting the last auto player should delete the team

PR #1004, opened 2024-05-29
One More Thing

tabascq:
One line missing from #1002


tabascq:
I test-deployed this and it works. Going to merge w/o review since it's soooo easy

PR #1007, opened 2024-06-03
Fix the freeform checkbox text

morganbr:
The freeform checkbox text got lost for embedded puzzles, put it back, fix the response text to indicate the response will be judged, rather than showing it as incorrect, and fix who gets to see submissions when


jbodner09:
Looks good to me!  For documentation, here's the on-submission view:
image

And here's the view after it's been judged:
image

jbodner09 (approved):

PR #1009, opened 2024-06-14
Sample Puzzle Accessibility Fixes

tabascq:
- Convert The Search for Spock from images to puzzle.js, both for puzzle and hints

- Add a copy of puzzle.js (as of 06-13-24) to the lib folder to facilitate the above - note that it is far better for individual puzzle events to use their own copies of puzzle.js, so we can iterate it faster and also not break prior events etc.

- Add alt text to Crop circles, both for puzzle and hints


jbodner09 (approved):

tabascq:
I would of course like to improve that, but this is certainly a start.
________________________________

PR #1012, opened 2024-06-16
Add database constraints

### LINKED ISSUE #763: Field lengths should be enforced in the database
Followup to #725 since it originally was enforcing field lengths in javascript instead.

jbodner09:
As usual, DISABLE WHITESPACE when reviewing.  Closes #763 by adding string length limits to all player input fields via database annotations.  As mentioned in that issue, these annotations also add client-side enforcement via javascript, so manual server-side checks for the few things that remained in C#, like team name length, were removed.  

Astute viewers will notice that the Regex constraint on team names was removed.  It was completely insufficient for enforcing the ">1 non-whitespace character at the beginning of the name" rule it was intended to since the regex's definition of whitespace was hard-coded to a list of characters that had no support for exotic unicode characters at all.  

Instead, a manual sanitizer method was written to explicitly only allow spaces as the allowable whitespace character. Note that this method still may allow some forms of unicode abuse, like Zalgo text that writes itself over other characters, since the main purpose of this sanitizer is to remove invisible characters that can allow teams to have apparently blank or identical names.  

Also fixes a crash on non-Blazor events where the redirect to the Teams page didn't work, making it impossible for players to join/create a team.  Note that this is just a workaround for the redirect not working; the underlying bug is still there.  This doesn't affect upcoming events like Puzzleday though since they're using Blazor.


Data/DataModel/Feedback.cs
```diff
@@ -43,6 +43,7 @@ public Feedback()
         /// <summary>
         /// The feedback text submitted by the user.
         /// </summary>
+        [MaxLength(500)]
```
morganbr, line 46: This might be short for feedback and it's only in betas, so it might be nice to leave this unlimited or at least put in a very long constraint

Data/DataModel/Message.cs
```diff
@@ -27,6 +28,7 @@ public class Message
         /// <summary>
         /// Gets or sets the subject.
         /// </summary>
+        [MaxLength(50)]
```
morganbr, line 31: Please check these values with @digit01Wave

Data/DataModel/PlayerInEvent.cs
```diff
@@ -20,8 +20,10 @@ public class PlayerInEvent
 
         public string Lunch { get; set; }
 
+        [MaxLength(100)]
         public string LunchModifications { get; set; }
 
+        [MaxLength(5)]
```
morganbr, line 26: This might be a word, maybe a 20 character limit?

Data/DataModel/Response.cs
```diff
@@ -52,6 +52,7 @@ public Response(Response source)
         /// The submitted text
         /// </summary>
         [Required]
+        [MaxLength(100)]
```
morganbr, line 55: Do you have data on this? This doesn't seem like an impossible bound to hit

Data/DataModel/Feedback.cs
jbodner09: This is now 5000

Data/DataModel/Message.cs
jbodner09: She says its good

Data/DataModel/PlayerInEvent.cs
jbodner09: So changed

Data/DataModel/Response.cs
jbodner09: Adjusted per conversation

morganbr (approved):

morganbr (approved):

PR #1014, opened 2024-06-18
Users/zhenyas/ps24

zhenyar:
Add a listing for Puzzle Safari 2024 event


ServerCore/ServerCore.csproj
```diff
@@ -62,6 +62,7 @@
   <ItemGroup>
     <Folder Include="Areas\Identity\Data\" />
     <Folder Include="Content\" />
+    <Folder Include="wwwroot\images\ps24\" />
```
tabascq, line 65: Let's remove this - should be unnecessary. Note that other image folders aren't in here.

tabascq:
Argh I tried to make a one line edit in GitHub to help this get through tonight and it changed all the line endings - we should remove ServerCore.csproj from this PR altogether. Sorry for gumming up the works...

Issue #1021, opened 2024-06-22
New correct response marks newest submission as correct


digit01Wave:
![image](https://github.com/PuzzleServer/mainpuzzleserver/assets/4121745/32d7e64c-0871-4231-84f3-1f17b3ee24ce)

PR #1022, opened 2024-06-22
Fixed submission when mark as correct after the fact

digit01Wave:
#1021 

Before fix
image

After fix
image

### DIFF
--- a/ServerCore/Pages/Submissions/Index.cshtml.cs
+++ b/ServerCore/Pages/Submissions/Index.cshtml.cs
@@ -375,7 +375,7 @@ namespace ServerCore.Pages.Submissions
                     && submissionView.Response.IsSolution 
                     && !Puzzle.IsFreeform)
                 {
-                    AnswerToken = Submissions.Last().SubmissionText;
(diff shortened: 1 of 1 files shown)


ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -370,18 +370,18 @@ orderby submission.TimeSubmitted
...
-            }
 
-            if (PuzzleState.SolvedTime != null)
-            {
-                if (!Puzzle.IsFreeform && Submissions?.Count > 0)
+                if (submissionView.Response != null
+                    && submissionView.Response.IsSolution 
+                    && !Puzzle.IsFreeform)
                 {
                     AnswerToken = Submissions.Last().SubmissionText;
```
tabascq, line 378: Why is this not submissionView.Submission.SubmissionText?

digit01Wave: AH. Good point

tabascq (approved):

Issue #1025, opened 2024-06-22
Feature request: more flexible url structure for customCSSFile


jbodner09:
Currently, there are a few holes in the ability to link to a CSS file from a puzzle depending on where the CSS file lives.  This is the current logic:

```c#
string puzzleFilePath = "";
if (Model.Puzzle.CustomURL != null)
{
    puzzleFilePath = @PuzzleHelper.GetFormattedUrl(Model.Puzzle, Model.Event.ID);
}
else if (Model.Puzzle.PuzzleFile != null)
{
    puzzleFilePath = @Model.Puzzle.PuzzleFile.Url.AbsoluteUri;
}
string cssFileName = String.IsNullOrEmpty(Model.Puzzle.CustomCSSFile) ? "" : Model.Puzzle.CustomCSSFile.Trim();
@if (cssFileName.Length > 0)
{
    cssFileName += (!cssFileName.EndsWith(".css")) ? ".css" : "";
    cssFileName = cssFileName.Replace('\\', '/');
    if (cssFileName.Contains('/'))
    {
        if (cssFileName.StartsWith(".") && (embedType > 0))
        {
            // For relative urls, ASSUME THAT THE RULES DESCRIBED ON THE UPLOAD PAGE HAVE BEEN FOLLOWED
            // This means that the leading . will be replaced with the puzzleFilePath variable above, minus the puzzle file itself
            // If the file doesn't start with a ., it's assumed to be a full url that will be used as-is
            cssFileName = puzzleFilePath.Substring(0, puzzleFilePath.LastIndexOf('/')) + cssFileName.Substring(1);
        }
        <link rel="stylesheet" href="@cssFileName" />
    }
    else
    {
        // Otherwise, assume that bare file names are in the shared resources folder 
        // (FileStoragePrefix used to account for different urls between development and production)
        <link rel="stylesheet" href="@(Model.FileStoragePrefix)/resources/@cssFileName" />
    }
}
```

There are a few problems with this.  First, assuming that a string with a / in it will NOT live in the shared resources presents a problem.  Right now, that means that EVERYTHING in the shared resources folder that is implicitly addressable via this field must live in the root of that shared resources folder.  In other words, there's no way to put these files in a subdirectory in the shared resources folder and have them addressable via this automatic way.  

[...]

PR #1026, opened 2024-06-23
Fix puzzle css path

jbodner09:
Partially addresses #1025. This update at least makes relative links work again, although not in the way they used to.  It also allows subfolders in the shared resources to be used.  

Accessing puzzle material files is now done by preceding the file name with a $

### DIFF
--- a/ServerCore/Pages/Submissions/Index.cshtml.cs
+++ b/ServerCore/Pages/Submissions/Index.cshtml.cs
@@ -396,7 +396,10 @@ namespace ServerCore.Pages.Submissions
                                                  where contentFile.Event == Event &&
                                                  contentFile.ShortName == Puzzle.CustomCSSFile.Substring(1)
                                                  select contentFile).SingleOrDefaultAsync();
-                    PossibleMaterialFile = content.Url.AbsoluteUri;
+                    if (content != null)
+                    {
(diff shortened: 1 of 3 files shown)


ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -384,6 +388,17 @@ orderby submission.TimeSubmitted
...
             FileStoragePrefix = FileManager.GetFileStoragePrefix(Event.ID, "");
+            if (!string.IsNullOrWhiteSpace(Puzzle.CustomCSSFile)) 
+            {
+                if (Puzzle.CustomCSSFile.StartsWith("$"))
+                {
+                    ContentFile content = await (from contentFile in context.ContentFiles
+                                                 where contentFile.Event == Event &&
+                                                 contentFile.ShortName == Puzzle.CustomCSSFile.Substring(1)
+                                                 select contentFile).SingleOrDefaultAsync();
+                    PossibleMaterialFile = content.Url.AbsoluteUri;
```
tabascq, line 399: The general approach makes sense but this will crash if no material is found - we should be gentler than that.

jbodner09: Good catch, fixed!

tabascq (approved):

PR #1027, opened 2024-06-24
Better group sorting

### LINKED ISSUE #1024: Feature request: better group sorting for puzzle state map
Right now, the state map sorts each puzzle within a group by the number of solves.  This is only useful if there are a small number of groups/large number of puzzles per group.  If it's the opposite, it becomes very difficult to tell which puzzles are getting the most/least solves within a group (and therefore overall).  

Possible solution:  In the filter, change the "all other puzzles" star * to lump all puzzles together instead of sorting them by group.  That way, any specific groups that are important like the meta still get shown together if they're mentioned specifically in the filter, and all others are just measured together, as if the group doesn't matter.  

Bonus feature:  filter out, or have a way to optionally filter out, any non-puzzles from the state map, likely based on the isPuzzle flag that's set per puzzle.

tabascq:
Groups at the same sorting level can now be separated by semicolons.

Fixes #1024

### DIFF
--- a/ServerCore/Pages/Events/Map.cshtml.cs
+++ b/ServerCore/Pages/Events/Map.cshtml.cs
@@ -40,7 +40,11 @@ namespace ServerCore.Pages.Events
             string[] groupList = Groups.Split("|");
             for (int i = 0; i < groupList.Length; i++)
             {
-                groupSortIndex[groupList[i]] = i;
+                string[] groupSublist = groupList[i].Split(";");
+                foreach (string sl in groupSublist)
+                {
+                    groupSortIndex[sl] = i;
+                }
             }
 
             // get the puzzles and teams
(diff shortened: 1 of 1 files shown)


jbodner09 (approved):

ServerCore/Pages/Events/Map.cshtml.cs
```diff
@@ -40,7 +40,11 @@ public async Task<IActionResult> OnGetAsync(int? refresh, string? groups)
             string[] groupList = Groups.Split("|");
             for (int i = 0; i < groupList.Length; i++)
             {
-                groupSortIndex[groupList[i]] = i;
+                string[] groupSublist = groupList[i].Split(";");
```
morganbr, line 43: What generates this? Also, what's the problem with pipe separation as above?

morganbr, line 43: ```Groups.Split({'|', ';'})``` would allow splitting by both characters without the nested loop

PR #1028, opened 2024-06-25
ebay print fix

jbodner09:
Instead of opening a new tab to wherever the iframe was initially located when the page was loaded, have the embedded puzzle open its current page in a new tab when a print keyboard command or link click are detected.  This allows puzzles like ebay, which navigate to different pages within the iframe, to correctly print the page they're on.  

Also fixes a bug with PDF printing where ASP incorrectly escapes a + character in the URL for some reason.  

Note 1:  allowing the iframe to open popups requires adding the allow-popups-to-escape-sandbox permission to the iframe so the unembedded puzzle has full permissions to print, etc. as needed.  

Note 2:  Since some browsers aggressively block popups, "please allow popups" was added to the print link since otherwise the print link will appear to do nothing except initiate a tiny notification by the browser that a popup was blocked.  It's currently unknown if this popup blocking was also an issue with the previous/current print implementation.  

Note 3:  Since different types of messages are now being sent to puzzles, an update to the puzzle resize code is also required.  It will be uploaded to the event once this PR is deployed.

### DIFF
--- a/ServerCore/Pages/Submissions/Index.cshtml
+++ b/ServerCore/Pages/Submissions/Index.cshtml
@@ -127,6 +127,14 @@
         background-color: rgba(255,255,255,1);
         overflow: auto;
     }
+    .print-link {
+        text-decoration: underline;
+        color: #0D6EFD;
+        cursor: pointer;
+    }
+    .print-link:hover {
+        color: #0A58CA;
+    }
     iframe {
         border: 0px solid black;
     }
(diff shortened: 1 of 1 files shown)


ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -521,23 +529,17 @@ else
...
+                <div><b>Want to print?</b> <span class="print-link">Click this link</span> for a printable version. Please allow popups!</div>
+            }
+            else if (embedType == 4) // images
+            {
+                <div><b>Want to print?</b> <a href="@puzzleFilePath" target="_blank">Click this link</a> for a printable version. Please allow popups!</div>
             }
             else if (embedType == 1) // PDFs
             {
-                <div><b>Want to print?</b> <a href="@puzzleFilePath" download>Click this link</a> to download the file. Print in color!</div>
+                <div><b>Want to print?</b> <a href="@puzzleFilePath" target="_blank" download>Click this link</a> to download the file. Please allow popups!</div>
```
morganbr, line 542: It would be nice to keep the instruction to print in color

morganbr (approved):

ServerCore/Pages/Submissions/Index.cshtml
jbodner09: Agreed, but given how difficult it is to know how to do that, as well as not having a ton of space horizontally, this seemed more important.

Issue #1020, opened 2024-06-22
Notification about puzzle response update needs to account for Html.Raw()


jbodner09:
As seen in the picture, it does correctly add the link, but it puts the link on more text than expected, and shows the Html.Raw call, which it shouldn't:

![image](https://github.com/PuzzleServer/mainpuzzleserver/assets/30329717/49141903-8a49-49b7-961e-a85b70ba41aa)

PR #1033, opened 2024-06-28
Make plaintext requirement more clear for responses

digit01Wave:
#1020 

The original bug was because "Html.Raw(" was used without a plaintext version at the beginning.
When the Response is just an Html.Raw, we end up returning the entire response text
image

This PR makes it more clear about this formatting requirement.

### Responses page
image

### Bulk creation page
image

### Create page
image

### What player sees
Notification for response "Here is plaintext Html.Raw(<span>Here is a <b>bolded</b> answer. </span>)"
image

Submissions page
image

### DIFF
--- a/ServerCore/Pages/Responses/Create.cshtml
+++ b/ServerCore/Pages/Responses/Create.cshtml
@@ -10,6 +10,8 @@
 }
 
 <h2>@Model.Puzzle.Name: Create response</h2>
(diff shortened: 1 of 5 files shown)


tabascq (approved):
Thanks!!! Was hoping to address this soon, but I wouldn't have done it this well.

Issue #1013, opened 2024-06-18
(nit) Placeholder/default text for ResponseText


sprx97:
When creating answers there are canonical ResponseTexts like "Correct!" and "Keep going..." that are commonly used. It would be nice to have some standardization via either a default if you leave it blank or a placeholder text.

PR #1035, opened 2024-06-29
Add default text for responses

digit01Wave:
#1013 

Adds default text for responses, but allows user to also override the default.
ResponsePill2


tabascq:
If we're going to do this then I think we should also have a brief guide to response writing, to make it clear when these defaults are inappropriate to use.

digit01Wave:
> If we're going to do this then I think we should also have a brief guide to response writing, to make it clear when these defaults are inappropriate to use.

I updated this to use pills instead

tabascq (approved):

Issue #1031, opened 2024-06-25
Feature request: Presence on puzzle list


morganbr:
PD Beta players wanted to see the "who's here" information on the main puzzle list. The presence store should be able to power that. We need a UX design to fit the names or other info with each puzzle

PR #1037, opened 2024-07-01
Add presense to puzzle page

digit01Wave:
#1031

Add Presence on puzzle list

At two players
image

At three players
image

### DIFF
--- a/ServerCore/Pages/Components/PresenceComponent.razor.cs
+++ b/ServerCore/Pages/Components/PresenceComponent.razor.cs
@@ -33,6 +33,9 @@ namespace ServerCore.Pages.Components
         [Parameter]
         public int PuzzleId { get; set; }
 
+        [Parameter]
+        public bool IsReadOnly { get; set; }
+
         Guid pageInstance = Guid.NewGuid();
         TeamPuzzleStore teamPuzzleStore;
 
@@ -106,7 +109,11 @@ namespace ServerCore.Pages.Components
 
             await UpdateModelAsync(teamPuzzleStore.PresentPages);
 
-            await MessageHub.BroadcastPresenceMessageAsync(new PresenceMessage { PageInstance = pageInstance, PuzzleUserId = PuzzleUserId, TeamId = TeamId, PuzzleId = PuzzleId, PresenceType = PresenceType.Active });
+            if (!this.IsReadOnly)
+            {
+                await MessageHub.BroadcastPresenceMessageAsync(new PresenceMessage { PageInstance = pageInstance, PuzzleUserId = PuzzleUserId, TeamId = TeamId, PuzzleId = PuzzleId, PresenceType = PresenceType.Active });
+            }
+
             await base.OnParametersSetAsync();
         }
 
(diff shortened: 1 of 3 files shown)


ServerCore/Pages/Components/PresenceComponent.razor.cs
```diff
@@ -33,6 +33,9 @@ public partial class PresenceComponent : IAsyncDisposable
         [Parameter]
         public int PuzzleId { get; set; }
 
+        [Parameter]
+        public bool IsReadOnly { get; set; }
```
tabascq, line 37: I think this is a somewhat weird name for the property. Wondering if "IsListComponent" is clearer.

tabascq (commented):
I think this is fantastic (modulo the one very minor comment I made), but I want us to think VERY carefully about the fact that we are having scale troubles and whether this change will make us more likely to run into performance issues day-of. @morganbr needs to be in that conversation.

morganbr:
> I think this is fantastic (modulo the one very minor comment I made), but I want us to think VERY carefully about the fact that we are having scale troubles and whether this change will make us more likely to run into performance issues day-of. @morganbr needs to be in that conversation.

If this doesn't create extra Blazor connections, which I think it doesn't and we can confirm, the perf should be ok.

ServerCore/Pages/Components/PresenceComponent.razor.cs
morganbr: Maybe add a comment for this? ReadOnly is a good concept for this (that it shows presence, but doesn't write it), but I didn't immediately get it either

ServerCore/Pages/Components/PresenceComponent.razor.cs
```diff
@@ -106,7 +109,11 @@ protected override async Task OnParametersSetAsync()
 
             await UpdateModelAsync(teamPuzzleStore.PresentPages);
 
-            await MessageHub.BroadcastPresenceMessageAsync(new PresenceMessage { PageInstance = pageInstance, PuzzleUserId = PuzzleUserId, TeamId = TeamId, PuzzleId = PuzzleId, PresenceType = PresenceType.Active });
+            if (!this.IsReadOnly)
```
morganbr, line 112: This check is also needed in DisposeAsync where it shows the player disconnecting from the puzzle

ServerCore/Pages/Puzzles/Play.cshtml
```diff
@@ -178,6 +184,12 @@ else
                             }
                         </td>
                     }
+                    else
+                    {
+                        <td>
+                            <component type="typeof(ServerCore.Pages.Components.PresenceComponent)" render-mode="Server" param-PuzzleUserId="@Model.LoggedInUser.ID" param-TeamId="@Model.Team.ID" param-PuzzleId="@item.ID" param-IsReadOnly=true />
```
morganbr, line 190: The full list of names might get crowded on the puzzle list if several players join a puzzle. Consider shortening it. Maybe something like for 3+ players, show a count?

morganbr (approved):
Thanks for doing this! At the meeting, let's confirm the effect on Blazor/SignalR connections

ServerCore/Pages/Puzzles/Play.cshtml
morganbr, line 190: This needs to check Event.AllowBlazor

morganbr (approved):

PR #1039, opened 2024-07-02
Fewer puzzle list columns

morganbr:
Use fewer columns on the puzzle list to make room for presence in embedded mode. Combines the puzzle name column and the link to the puzzle and removes the hints column (it's already on the puzzle page). See screenshot:
image


tabascq (approved):

morganbr:
This is obsolete

PR #1042, opened 2024-07-04
Revert "Add presense to puzzle page (#1037)"

tabascq:
This reverts commit c20ed268c569b37a7735cdf52c5b2640702ba297.

Just managing risk here.

Trying to ensure that we have a functional site so that our event can properly be validated - right now the site will keel over as soon as we solve !!!Start the Event!!! as solved, and we would have no choice but to immediately turn off ALL Blazor features. Bullet dodged that we discovered this before 10:31AM on event day.

If we have an appropriately LOW RISK approach well before pregame starts on 7/8, we can try again.

### DIFF
--- a/ServerCore/Pages/Components/PresenceComponent.razor
+++ b/ServerCore/Pages/Components/PresenceComponent.razor
@@ -11,19 +11,7 @@
 
 <div>
     <span>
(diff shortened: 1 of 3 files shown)


tabascq:
Revert no longer needed

PR #1043, opened 2024-07-04
Revert the part of #1037 that breaks the puzzle list for events with …

tabascq:
…27+ puzzles unlocked


digit01Wave (approved):

tabascq:
Revert no longer needed

PR #1044, opened 2024-07-05
Add full team presence component

morganbr:
Adds a Blazor component for sending full-team presence information to JavaScript and using that to fill in presence information on the puzzle list to work around the 27-ish Blazor component limit on a page. Also includes factoring shared code out of the presence component and centralizing more of it in the PresenceStore.


ServerCore/ServerMessages/PresenceStore.cs
```diff
@@ -88,10 +88,52 @@ public async Task MergePresenceState(PresenceMessage[] allPresence)
...
+
+            return teamPuzzleStore;
+        }
+
+        private async Task InvokeTeamPresenceChange(int puzzleId, IList<PresenceModel> presentUsers)
+        {
+            var onTeamPuzzlePresenceChange = OnTeamPresenceChange;
+            if (onTeamPuzzlePresenceChange != null)
+            {
+                await onTeamPuzzlePresenceChange?.Invoke(puzzleId, presentUsers);
```
tabascq, line 135: unnecessary ? since 133 already nullchecks and nothing can change the value

ServerCore/ServerMessages/PresenceStore.cs
```diff
@@ -88,10 +88,52 @@ public async Task MergePresenceState(PresenceMessage[] allPresence)
...
         }
+    }
+
+    /// <summary>
+    /// Model for a user's presence on a puzzle page
+    /// </summary>
+    public class PresenceModel
+    {
+        public int UserId { get; set; }
+        public string Name { get; set; }
```
tabascq, line 99: A _little_ weird that this is here even though the PresenceStore never sets it, but in the short term I won't quibble

ServerCore/ServerMessages/PresenceStore.cs
```diff
@@ -100,23 +142,52 @@ private class TeamStore
...
+
         public async Task InvokeTeamPuzzlePresenceChange()
         {
+            IList<PresenceModel> presentUsers = GetPresentUsers();
+
             var onTeamPuzzlePresenceChange = OnTeamPuzzlePresenceChange;
             if (onTeamPuzzlePresenceChange != null)
             {
-                await onTeamPuzzlePresenceChange?.Invoke(PresentPages);
+                await onTeamPuzzlePresenceChange?.Invoke(PuzzleId, presentUsers);
```
tabascq, line 172: same comment as line 135

tabascq (approved):
A few nits but nothing remotely major.

Longer term, there might be some value in making a base class for these two presence components and trying to share more code, but minimizing risk in the short term is certainly preferable.

ServerCore/ServerMessages/PresenceStore.cs
morganbr: Fixed

ServerCore/ServerMessages/PresenceStore.cs
morganbr: Fixed

ServerCore/ServerMessages/PresenceStore.cs
morganbr: Yeah, it's a bit messy. I really wish we had a good way to get the name here in the first place (we could send it over SignalR, but I'm trying to keep presence messages small since there are a ton of them)

PR #1045, opened 2024-07-05
Presence List Perf Fix

tabascq:
Sending an array of {puzzleId, presenceText} objects instead of just one per message. This allows us to batch up all presence on page load, and only send presence data for puzzles that have it.


tabascq:
Can this get a review please? We need this for Monday and need to deploy it before pregame begins.

morganbr (approved):

PR #1046, opened 2024-07-06
New EphemeralHacks property for emergency usage

tabascq:
Whenever we have a property we want to temporarily add to an event, we can use this property on a short-term basis.

Intended usage is a space-delimited set of words that are ephemerally defined in our codebase. Anything with long-term importance should be moved out of this property and placed somewhere else.

In the very short term, we are going to use this to install temporary shutoff valves for some new features that may bog us down on the single front-end server that Puzzleday 2024 is going to have to run on.

Initial set of shutoff valves:
- kill-notifications: turns off toast notifications.
- kill-team-presence: turns off aggregated team presence on the puzzle list page.
- kill-presence: turns off all presence, both on the puzzle list page and on individual puzzles.


tabascq:
Can this get a review please? We need this for Monday and need to deploy it before pregame begins.

Data/DataModel/Event.cs
```diff
@@ -312,5 +312,11 @@ public bool AreStandingsAvailableNow
         /// True if Blazor can run on pages for this event. Should normally be true, but can be disabled in an emergency
         /// </summary>
         public bool AllowBlazor { get; set; }
+
+        /// <summary>
+        /// Short-term hacks for modifying behavior without adding new properties all the time.
+        /// Note: If a hack lasts more than a few months, it should probably be promoted to a real property.
+        /// </summary>
+        public string EphemeralHacks { get; set; }
```
morganbr, line 320: Consider adding ```[NotMapped]``` properties with getters that encapsulate the string checking logic for the strings you're adding so we don't have to spread the logic around and so we have central-ish knowledge of what they are

ServerCore/Pages/Puzzles/Play.cshtml
```diff
@@ -25,7 +26,7 @@
         }
 </style>
 
-@if (Model.Event.AllowBlazor)
+@if (Model.Event.AllowBlazor && Model.Event.EphemeralHacks?.Contains("kill-presence") != true)
```
morganbr, line 29: Should this also check ```canUsePresence```?

morganbr (approved):

Data/DataModel/Event.cs
tabascq: Good idea, will do!

PR #1047, opened 2024-07-08
State map presence icon

morganbr:
Adds an eyes icon to the state map when a team is looking at a puzzle. Note that since the state map isn't Blazor, this still requires a refresh to get new data (like the rest of the state map). See screenshot:

image


tabascq (approved):
I kinda want this to be live, but this is probably the best approach given the timing of when it's going in!

Issue #1076, opened 2024-07-19
During the event could not release all the puzzles for some reason


digit01Wave:

digit01Wave:
When we marked the start the event puzzle, no puzzle was unlocked

PR #1087, opened 2024-07-25
Fix team unlock for single player puzzle

digit01Wave:
#1076

During the event, we had a SinglePlayerPuzzle with a StartTheEvent puzzle prerequisite.
When we solved the StartTheEvent, no team puzzles were unlocked.
This was because it failed when we tried to update the team state for a single player puzzle. All single player puzzles should have been filtered out


tabascq (approved):
Thanks!!!

Issue #1067, opened 2024-07-13
Help Threads: Multiple Clicks submit multiple times


harmon52:
Repro: Write a message in a help thread and click the "submit" button multiple times as quickly as possible. The same message will get submitted multiple times.

PR #1092, opened 2024-08-20
Don't allow messages to be added twice

digit01Wave:
#1067

Multiple clicks submits mutliple times.
This PR fixes that

### DIFF
--- a/ServerCore/Pages/Threads/PuzzleThread.cshtml.cs
+++ b/ServerCore/Pages/Threads/PuzzleThread.cshtml.cs
@@ -210,14 +210,27 @@ namespace ServerCore.Pages.Threads
             m.ClaimerID = NewMessage.ClaimerID;
             m.PlayerID = NewMessage.PlayerID;
 
+            bool isMessageAdded = false;
             using (IDbContextTransaction transaction = _context.Database.BeginTransaction(System.Data.IsolationLevel.Serializable))
             {
-                _context.Messages.Add(m);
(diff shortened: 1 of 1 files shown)


ServerCore/Pages/Threads/PuzzleThread.cshtml.cs
digit01Wave, line 1: Does anyone know a better way to do this so that we don't run server queries at all?

ServerCore/Pages/Threads/PuzzleThread.cshtml.cs
```diff
@@ -210,14 +210,27 @@ public async Task<IActionResult> OnPostAsync()
...
             m.PlayerID = NewMessage.PlayerID;
 
+            bool isMessageAdded = false;
             using (IDbContextTransaction transaction = _context.Database.BeginTransaction(System.Data.IsolationLevel.Serializable))
             {
-                _context.Messages.Add(m);
-                await _context.SaveChangesAsync();
-                transaction.Commit();
+                // Make sure to not add the message if already added previously (sometimes the user double clicks)
+                var newestMessage = _context.Messages
```
tabascq, line 217: we should await this if at all possible - I think LastAsync is a thing that can be used here.

digit01Wave: done!

tabascq (approved):

PR #1095, opened 2024-08-23
Add live event schedule display page

asyasky:
Add a live event schedule page that shows the schedule for the current team and link it from the puzzles list page.

### DIFF
--- a/ServerCore/Helpers/LiveEventHelper.cs
+++ b/ServerCore/Helpers/LiveEventHelper.cs
@@ -20,15 +20,15 @@ namespace ServerCore.Helpers
             await EventClosingReminder(context, currentEvent, hubContext);
         }
 
-        public static List<LiveEventSchedule> GetTeamSchedule(PuzzleServerContext context, Event currentEvent, Team team)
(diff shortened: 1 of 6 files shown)


ServerCore/Helpers/LiveEventHelper.cs
```diff
@@ -22,6 +20,17 @@ public static async Task TriggerNotifications(PuzzleServerContext context, Event
             await EventClosingReminder(context, currentEvent, hubContext);
         }
 
+        public static List<LiveEventSchedule> GetTeamSchedule(PuzzleServerContext context, Event currentEvent, Team team)
```
tabascq, line 23: Should be async, right?

tabascq: Also, since we have like no time to test this, can you please add a ripcord so we can pull it if necessary? See the [NotMapped] EphemeralHack* properties at the bottom of Event.cs

tabascq (changes requested):

ServerCore/Helpers/LiveEventHelper.cs
asyasky: Changes pushed.

tabascq (approved):

PR #1104, opened 2024-08-28
Notification UX Reorg

tabascq:
Revising the UX for notifications to try to make it more palatable.

Changes:
- a new envelope icon has been added to the toolbar that also displays a count of new notifications.
- clicking the envelope icon takes you to a new page that lists all notifications and resets the new-notification count. Notifications can be permanently deleted from this page.
- Notifications are by default ephemeral, meaning they display onscreen for five seconds and then disappear (and be viewed later by clicking the envelope).
- Notifications can also be marked as critical, meaning that they do always stay up until manually dismissed.
- Two types of notifications have been marked as critical: direct emails from staff (bulk mail, not help thread responses), and upcoming live event warnings.

Other than the marking of notifications as critical, no other changes have been made to the model; other improvements might be possible by moving the storage location for notifications to the server.

### DIFF
--- a/ServerCore/ServerMessages/ServerMessageTypes.cs
+++ b/ServerCore/ServerMessages/ServerMessageTypes.cs
@@ -107,4 +107,9 @@ public class Notification
     /// The link that should be followed if the notification is clicked.
     /// </summary>
     public string LinkUrl { get; set; }
+
+    /// <summary>
+    /// Says whether the notification is so critical that it should not disappear automatically
+    /// </summary>
+    public bool IsCritical { get; set; }
 }
--- a/ServerCore/Helpers/LiveEventHelper.cs
+++ b/ServerCore/Helpers/LiveEventHelper.cs
@@ -243,11 +243,11 @@ namespace ServerCore.Helpers
         {
             if (team == null)
             {
-                await hubContext.SendNotification(e, title, message, puzzleUrl);
+                await hubContext.SendNotification(e, title, message, puzzleUrl, true);
             }
             else
             {
-                await hubContext.SendNotification(team, title, message, puzzleUrl);
+                await hubContext.SendNotification(team, title, message, puzzleUrl, true);
             }
         }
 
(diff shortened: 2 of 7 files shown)


digit01Wave:
For the ephemeral notifications, can we make it so that only 3 ephemeral notification can ever appear at once like how phones do it (e.g. if 3 are already on the screen, a new one will push two down and the last one will dissappear)

ServerCore/ServerMessages/ServerMessageTypes.cs
```diff
@@ -107,4 +107,9 @@ public class Notification
     /// The link that should be followed if the notification is clicked.
     /// </summary>
     public string LinkUrl { get; set; }
+
+    /// <summary>
+    /// Says whether the notification is so critical that it should not disappear automatically
+    /// </summary>
+    public bool IsCritical { get; set; }
```
digit01Wave, line 114: Do we want to additionally add a parameter for players to suppress non-critical notifications by type?

digit01Wave (approved):

ServerCore/ServerMessages/ServerMessageTypes.cs
tabascq: I don't understand what you mean by this, can you please elaborate?

tabascq:
> For the ephemeral notifications, can we make it so that only 3 ephemeral notification can ever appear at once like how phones do it (e.g. if 3 are already on the screen, a new one will push two down and the last one will dissappear)

Done, but I set it to 5 instead of 3 for the time being.

ServerCore/ServerMessages/ServerMessageTypes.cs
digit01Wave: I mean if we want to give every notification a "category" (e.g. PuzzleThread, GameControlMessage, PuzzleSolves). Then players can be given an option of which ones they want to leave on and which ones they want to leave off

tabascq: Makes sense, thanks! I will make a separate issue for this rather than do it now, as I don't feel like doing the client UX at present.

PR #1105, opened 2024-09-06
Alpha Event Playtesting

tabascq:
Events can be marked as alpha testing events.
image

In an alpha testing event, authors can mark a puzzle with a count of how many alpha tests it needs.
image

As an alpha tester, the puzzle list page shows all puzzles that need alpha tests. 
image

When one of these puzzles is clicked, the puzzle content is grayed out unless/until the puzzle is claimed by the tester. Claiming a puzzle unlocks it for the tester and decrements the count.
image


ServerCore/Areas/Identity/UserAuthorizationPolicy/AuthorizationHelper.cs
```diff
@@ -200,7 +200,7 @@ public async Task PlayerCanSeePuzzleCheck(AuthorizationHandlerContext authContex
 
             if (thisEvent != null && puzzle != null)
             {
-                if (thisEvent.AreAnswersAvailableNow)
+                if (thisEvent.AreAnswersAvailableNow || thisEvent.IsAlphaTestingEvent)
```
digit01Wave, line 203: What happens if a player happens to find the alpha site?

ServerCore/Pages/Puzzles/Play.cshtml
```diff
@@ -49,6 +49,34 @@
     </div>
 }
 
+@if (Model.Event.IsAlphaTestingEvent && Model.Event.EmbedPuzzles && Model.VisibleTeamPuzzleViews.Any())
```
digit01Wave, line 52: I keep forgetting, but what happens when it is not an embeded puzzle? Will alpha testing not work?

digit01Wave (approved):

ServerCore/Areas/Identity/UserAuthorizationPolicy/AuthorizationHelper.cs
tabascq: Very good concern - basically, they get in. I'll add a reminder to set the max number of teams low (like 0) when this is set, so that admins have to create player teams (and I just enforced this for the alpha site as well).

ServerCore/Pages/Puzzles/Play.cshtml
tabascq: Alpha testing currently won't work without embedding because without embedding there's no way to grey out the puzzle. If someone calls for a non-embedded event with alpha testing, we'll decide then what that really means.

Issue #974, opened 2024-04-16
Allow hints+help threads


morganbr:
We would like to allow hints for simple questions while still allowing help threads (at a cost of some hint tokens) at the same time

digit01Wave:
@morganbr , do we want the number of hints to unlock messaging to be the same for each puzzle?

digit01Wave:
The plan we're voting for is to have an event default value on the event object. Then a nullable override value on the puzzle object. That way the code can check for the override first & fall back to the event value if there isn't one.

PR #1106, opened 2024-09-10
Add datamodel for hint coin cost for help thread

digit01Wave:
#974 

We want the ability to unlock help threads via hints

UPDATE: Replaced with https://github.com/PuzzleServer/mainpuzzleserver/pull/1107


Data/DataModel/Event.cs
```diff
@@ -233,6 +233,12 @@ public bool AreStandingsAvailableNow
         /// </summary>
         public bool HideHints { get; set; }
 
+        /// <summary>
+        /// The number of hint coins a team must have to unlock help threads for any puzzle.
+        /// Note that this can be overridden by the puzzle itself as well.
+        /// </summary>
+        public uint DefaultCostForHelpThread { get; set; }
```
tabascq, line 240: It doesn't *really* matter, but I am wondering why this is a uint whereas the puzzle property is an int. They actually get stored in the DB as differently-sized fields (and uint is apparently larger FWIW). If you're fine with things as they are, I'm happy to go with it because the event table is exceedingly small, but I will hold merging until you ring in.

tabascq (approved):

Data/DataModel/Event.cs
digit01Wave: Thanks for the catch! I updated event to be int to match Hint and Puzzle

tabascq:
I hate to be annoying but we shouldn't need both migrations here. Is it possible to merge the migrations like how #959 became #963?

PR #1107, opened 2024-09-14
Added hint coin cost for help threads

digit01Wave:
#974

We want the ability to unlock help threads via hints


tabascq (approved):
Thanks for making the changes!

Issue #1062, opened 2024-07-13
Feature request: ability to see puzzle authors from help thread list.


JasonRajtar:
I see help threads coming in and I want to let the author know that they have a help request, but I don't know who the authors are.  When I click on the puzzle link, I just get an access denied since I'm not the author.  It would be nice if I could easily see who to notify.

jbodner09:
Sortable headers would also be super helpful.

vidapour:
done with #1108

PR #1108, opened 2024-09-18
Author Column in Help Thread List

mijon-work:
This adds a concatenated list of puzzle authors to the page of help threads.  

Context on the request here: [Feature request: ability to see puzzle authors from help thread list. #1062](https://github.com/PuzzleServer/mainpuzzleserver/issues/1062)

I chose the list of Authors instead of the CustomAuthorText field on the puzzle, since that seemed to be the standard for the admin views.  Please let me know if that was the wrong call (it's much less code to use the CustomAuthorText field).  I also chose to put it as the middle column to allow the most urgent information to remain easily scannable (though I am far from an expert at what the users of this page need!)

Example below, including testing that it still works if no Authors are returned:
image

### DIFF
--- a/ServerCore/Pages/Threads/PuzzleThreads.cshtml
+++ b/ServerCore/Pages/Threads/PuzzleThreads.cshtml
@@ -142,9 +142,9 @@
                     <a asp-page="/Threads/PuzzleThread" asp-route-puzzleId="@item.PuzzleID" asp-route-teamId="@item.TeamID" asp-route-playerId="@item.PlayerID">@item.Subject</a>
                 </td>
                 <td>
-                    @if (@item.PuzzleID != null)
+                    @if (@item.PuzzleID.HasValue)
                     {
-                        @Model.AuthorsForPuzzleID[item.PuzzleID ?? 0]
+                        @Model.AuthorsForPuzzleID[item.PuzzleID.Value]
                     }
                 </td>
                 <td>
--- a/ServerCore/Pages/Threads/PuzzleThreads.cshtml
+++ b/ServerCore/Pages/Threads/PuzzleThreads.cshtml
@@ -115,6 +115,9 @@
                 Thread
             </th>
             <th>
+                Author(s)
+            </th>
+            <th>
                 Timestamp
             </th>
             <th>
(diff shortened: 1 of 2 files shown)


ServerCore/Pages/Threads/PuzzleThreads.cshtml.cs
```diff
@@ -163,6 +168,19 @@ public async Task<IActionResult> OnGetAsync(
...
 
+            ILookup<int, string> puzzleAuthors = (await (from author in _context.PuzzleAuthors
+                                                         where author.Puzzle.Event == Event
+                                                         select author).ToListAsync()).ToLookup(author => author.PuzzleID, author => author.Author.Name);
+
+            AuthorsForPuzzleID = new Dictionary<int, string>();
+
+            foreach (var message in LatestMessagesFromEachThread)
+            { 
+                IEnumerable<string> authorsForPuzzle = puzzleAuthors[message.PuzzleID ?? 0];
```
tabascq, line 179: Two things about the safety-checking here:

1. Safety checking is great!
2. Better to just skip messages that have no PuzzleID than to assign them to the wrong puzzle.

ServerCore/Pages/Threads/PuzzleThreads.cshtml
```diff
@@ -138,6 +141,12 @@
                 <td>
                     <a asp-page="/Threads/PuzzleThread" asp-route-puzzleId="@item.PuzzleID" asp-route-teamId="@item.TeamID" asp-route-playerId="@item.PlayerID">@item.Subject</a>
                 </td>
+                <td>
+                    @if (@item.PuzzleID != null)
+                    {
+                        @Model.AuthorsForPuzzleID[item.PuzzleID ?? 0]
```
tabascq, line 147: Don't need to double-protect here; you know it's not null at this point.

tabascq (changes requested):
Good stuff! A couple extremely minor issues, easily addressed. Thanks!

ServerCore/Pages/Threads/PuzzleThreads.cshtml
mijon-work: I thought the same, but even with the null check in the if, there's an error in converting from int? to int for PuzzleID.  `@Model.AuthorsForPuzzleID[(int)item.PuzzleID]` would also solve the error.  Is casting preferable when we know it's safe?  Removing the if could also work, but defaulting to 0 could technically give incorrect data if something goes wrong in creating the dictionary.

tabascq: item.PuzzleID.Value would be fine and we use that all over (and HasValue for the null check if you feel like it, though we're less consistent with that)

ServerCore/Pages/Threads/PuzzleThreads.cshtml.cs
mijon-work: Great point, I like that!  Skipping definitely seems like the better choice.  I'll make this update pending the int? / int conversation below.

ServerCore/Pages/Threads/PuzzleThreads.cshtml
mijon-work: Ooh even better, thanks!

tabascq (approved):
Thanks very much!!!

Issue #1094, opened 2024-08-22
[UI] Registration page not have spaces and other UI issues


digit01Wave:
![image](https://github.com/user-attachments/assets/18186ef9-94bd-4007-8e07-90971c1a8896)
- No spaces in labels
- Visible to others doesn't show
- Register doesn't look like a button

vidapour:
shipped with #1109

PR #972, opened 2024-04-12
Tools overhaul

jbodner09:
Also changes the link in the navbar from "Tools" to "Getting Started" since the info is now much more relevant for new solvers.

PR #1109, opened 2024-09-18
Registration Page Spaces and Button Styling

mijon-work:
[[UI] Registration page not have spaces and other UI issues #1094](https://github.com/PuzzleServer/mainpuzzleserver/issues/1094)

This issue mentions 3 separate things to fix:

1. **No spaces in labels** 
For this one, I added DisplayName to the fields which should have spaces in their names.  I didn't add it to the ones where the DisplayName field would be identical to the variable name, but I can see value in doing that for consistency if preferred!

2. **Visible to others doesn't show** 
It appears to me that this is supposed to be a checkbox type.

3. **Register doesn't look like a button** 
Removed specific button class (same fix was made [in this PR](https://github.com/PuzzleServer/mainpuzzleserver/pull/972/files) for Login.cshtml)

Screenshots of before/after (apologies for the slightly differing window sizes)

Before:
image

After:
image

### DIFF
--- a/Data/DataModel/PuzzleUser.cs
+++ b/Data/DataModel/PuzzleUser.cs
@@ -1,4 +1,5 @@
 using System;
+using System.ComponentModel;
 using System.ComponentModel.DataAnnotations;
 using System.ComponentModel.DataAnnotations.Schema;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


tabascq (approved):
Looks like we missed this page when we did an update to Bootstrap this spring. Your changes look great, thanks!

PR #1117, opened 2024-11-24
Allow threads to be shown for hints

digit01Wave:
#974 

When we have hints, allow users to unlock help threads

### Before unlock:
image
image

### After unlock
image

### When try to access thread that has not been unlocked
image


ServerCore/Pages/Threads/PuzzleThread.cshtml.cs
```diff
@@ -74,7 +74,27 @@ public async Task<IActionResult> OnGetAsync(int? puzzleId, int? teamId, int? pla
...
             }
 
-            Team team = null;
+            // If we are showing hints, only allow the thread to be seen if the user has unlocked it
+            if (!Event.HideHints)
+            {
+                string returnUrl = HttpContext.Request.Path.HasValue ? HttpContext.Request.Path.Value : null;
+                if (Event.DefaultCostForHelpThread < 0)
+                {
+                    return RedirectToPage("/ErrorPage", new { ErrorMessage = "Help threads are not allowed for this event" });
```
tabascq, line 83: Is this something that can happen without the player customizing the URL? If not, then I think `return Forbid();` is enough.

tabascq: Same goes for the redirect on line 93.

digit01Wave: done

ServerCore/Pages/Teams/Hints.cshtml
```diff
@@ -90,16 +91,25 @@ You have <b>
...
                     @if (hint.IsUnlocked)
                     {
-                        @hint.Hint.Content
+                        if (hintViewState != null)
+                        {
+                            @hint.Content
+                        }
+                        else if (puzzleThreadViewState != null)
+                        {
+                            <a asp-Page="/Threads/PuzzleThread" asp-route-puzzleid="@Model.PuzzleID" asp-route-teamId="@Model.Team?.ID" asp-route-playerId="@Model.LoggedInUser.ID">Use this link to message author!</a>
```
tabascq, line 111: If there's room, "Use this link to message **the** author!" would be clearer.

digit01Wave: done!

tabascq (approved):

Issue #1070, opened 2024-07-14
Make only authors or admins able to see help threads unless author allows everyone to see puzzle


digit01Wave:
By default, we only want authors, admins, and support to see help threads.
Authors can choose to release their help thread to other authors as well, however.

PR #1119, opened 2024-11-26
Add ShowHelpThreadsToAllGameControl

digit01Wave:
#1070

First PR to hide help threads from non-author/non-support


tabascq (commented):
Should the default value be true?

tabascq:
Or change the property to Hide instead of Show?

tabascq:
As per discussion in the room, leaving it as-is for now. Idea is that admins see everything and authors see their stuff only.

tabascq (approved):

PR #1120, opened 2024-11-26
Remove show help thread to all

digit01Wave:
#1070

Actually didn't need to have this property at all because another property already existed
image


tabascq (commented):
I'd prefer to simply revert commit 42266a0 (from PR #1119) since we never deployed it, and this will keep us from thrashing the DB more than necessary. Does that work for you?

digit01Wave:
> I'd prefer to simply revert commit [42266a0](https://github.com/PuzzleServer/mainpuzzleserver/commit/42266a0152516681305913880edd0052b043f3a2) (from PR #1119) since we never deployed it, and this will keep us from thrashing the DB more than necessary. Does that work for you?

Here is the revert
https://github.com/PuzzleServer/mainpuzzleserver/pull/1122/files#diff-8bfb7af99fb51026969e3293c0e21455e293952224faa6e255145745798ebcf3

tabascq:
Closing in favor of #1122

PR #1121, opened 2024-11-26
Fixed help thread description for author

digit01Wave:
#1070 
- Made description for ShouldShowHelpMessageOnlyToAuthor more clear
- Also fixed issue if there was a filter, would still respect the roles


tabascq (approved):
Looks good, I'll approve but not merge until we make a decision on whether to merge #1120 or revert #1119.

PR #1124, opened 2024-12-01
Update Puzzle constructor to ShowHelpThreadsToAllGameControl

digit01Wave:
Forgot to add new property to the constructor.


tabascq (commented):
Shouldn't the one change to puzzle.cs be enough?

digit01Wave:
> Shouldn't the one change to puzzle.cs be enough?

Oh. I had thought i had to make a migration if i changed the models. If not then I'll just abandon this one and replace with the implementation PR
https://github.com/PuzzleServer/mainpuzzleserver/pull/1125

PR #1128, opened 2024-12-06
Notification updates

### LINKED ISSUE #1059: Send a notification if TeamAnnouncement is triggered/updated or if Announcement is updated

### LINKED ISSUE #1078: Team management difficult as an admin
It's hard to create teams, move players, etc. in the new team UX as an admin. Date problems, you're already on a team problems, other constraints. Admins should have no restrictions.

tabascq:
Send notifications when TeamAnnouncement is triggered, or when Announcement is changed. This way, players who are sticking on one page for a while (e.g. solving a puzzle) will still have an opportunity to see notifications.

Bonus fixes:
- A JS notification exception with unregistered players has been fixed.
- Admins can add new (empty) teams even if they are already on another team.

Fixes #1059.
Fixes #1078.

### DIFF
--- a/ServerCore/Pages/Shared/_EventNavigationPartial.cshtml
+++ b/ServerCore/Pages/Shared/_EventNavigationPartial.cshtml
@@ -399,6 +399,9 @@
         }
 
         function updateToastCount() {
+            var unreadCounter = document.querySelector("#unreadCounter");
+            if (!unreadCounter) return;
+
             if (newToastCount) { unreadCounter.classList.remove("hidden"); unreadCounter.innerHTML = newToastCount; }
             else { unreadCounter.classList.add("hidden"); }
         }
--- a/ServerCore/Pages/Teams/Create.cshtml.cs
+++ b/ServerCore/Pages/Teams/Create.cshtml.cs
@@ -70,7 +70,7 @@ namespace ServerCore.Pages.Teams
                 return NotFound();
             }
 
-            if (await (from member in _context.TeamMembers
+            if (EventRole != EventRole.admin && await (from member in _context.TeamMembers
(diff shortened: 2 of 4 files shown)


ServerCore/Pages/Teams/Create.cshtml.cs
```diff
@@ -70,7 +70,7 @@ public async Task<IActionResult> OnPostAsync()
                 return NotFound();
             }
 
-            if (await (from member in _context.TeamMembers
+            if (EventRole != EventRole.admin && await (from member in _context.TeamMembers
```
digit01Wave, line 73: Does this mean that admins can create and become part of multiple teams?

ServerCore/Pages/Shared/_EventNavigationPartial.cshtml
```diff
@@ -399,6 +399,9 @@
         }
 
         function updateToastCount() {
+            var unreadCounter = document.querySelector("#unreadCounter");
+            if (!unreadCounter) return;
```
digit01Wave, line 403: In what cases would this be hit? Should this log an error?

ServerCore/Pages/Teams/Create.cshtml.cs
tabascq: No - admins create blank teams in all cases. The code was still checking to see if admins are on a team, which is pointless.

ServerCore/Pages/Shared/_EventNavigationPartial.cshtml
tabascq: It's hit when a player has not yet registered. I don't think we need an error.

digit01Wave (approved):

PR #1132, opened 2025-01-03
Handle Login Drift

### LINKED ISSUE #1073: Generate recovery link for mangled accounts
Allow global/event admins to access any member of any team and click a "Generate Recovery link" button that makes a URL that can be pasted into an email. The URL should contain the PuzzleUser ID as well as some freshly-generated secret about the PuzzleUser (a GUID, perhaps) to prevent random internet hijacking. The web page corresponding to that URL should authenticate the user and then associate the PuzzleUser for that ID with the IdentityUserID that was just used to log in. 

This can also be used to migrate logins from employee to alumnus, etc.

tabascq:
For years, we've been beset by a small number of users losing access to their accounts, needing manual reset at extremely inopportune times.

After a good deal of code review and web searching, I've finally discovered that a) this is normal and b) there is a simple bit of code that can be used for automatic and instantaneous account recovery.

I've tested this by mangling my local DB ProviderKey values, and observing that it successfully self-heals with the user being none the wiser.

Fixes #1073, though does it in a better way than expected.

### DIFF
--- a/ServerCore/Startup.cs
+++ b/ServerCore/Startup.cs
@@ -3,6 +3,7 @@ using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Identity;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
@@ -149,6 +150,13 @@ namespace ServerCore
             services.AddSingleton<ServerMessageListener>();
             services.AddSingleton<PresenceStore>();
             services.AddSingleton<NotificationHelper>();
+
+            // Require emails to be unique so that we can be resilient to external provider login changes.
+            // Search the codebase for "RequireUniqueEmail" for the code that requires this.
+            services.Configure<IdentityOptions>(options =>
+            {
+                options.User.RequireUniqueEmail = true;
+            });
         }
 
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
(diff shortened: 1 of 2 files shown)


ServerCore/Startup.cs
```diff
@@ -149,6 +150,13 @@ public void ConfigureServices(IServiceCollection services)
             services.AddSingleton<ServerMessageListener>();
             services.AddSingleton<PresenceStore>();
             services.AddSingleton<NotificationHelper>();
+
+            // Require emails to be unique so that we can be resilient to external provider login changes.
+            // Search the codebase for "RequireUniqueEmail" for the code that requires this.
+            services.Configure<IdentityOptions>(options =>
+            {
+                options.User.RequireUniqueEmail = true;
```
tabascq, line 158: Note that I have not verified that emails are actually unique at present; I would not be surprised if some staffers have some duplication generated during server development, or perhaps some migrations have left stale users behind. Verification/cleanup is required before this code can be merged.

digit01Wave (approved):
I am not that well versed in authentication, but the change looks fine to me

tabascq:
I'd like @asyasky or @morganbr to ring in on this as well, and also to help determine how many duplicate emails we currently have in the DB so we can decide how best to remove them before merging.

tabascq:
Actually, if you agree that the query below is the right query, we may be all set as it returns zero results:

```
SELECT Email, COUNT(*)
FROM [dbo].[AspNetUsers]
GROUP BY Email
HAVING COUNT(*) > 1;
```

ServerCore/Startup.cs
asyasky: I've had to delete the existing AspNet user in order to fix this issue in the past, so I suspect there are few to no duplicates.

morganbr:
> Actually, if you agree that the query below is the right query, we may be all set as it returns zero results:
> 
> ```
> SELECT Email, COUNT(*)
> FROM [dbo].[AspNetUsers]
> GROUP BY Email
> HAVING COUNT(*) > 1;
> ```

Yes, this is good

asyasky:
I don't think we have any way to verify that this is what's causing our problem, but I don't see any downsides to trying it and it sets us up for the multiple logins feature, which is a nice bonus. :) This is exciting!

asyasky (approved):
This looks good to me, but I asked Morgan to check me on the duplicates sql query (I think it looks ok but my sql is pretty meh).

PR #1135, opened 2025-01-17
File Upload Enhancements

tabascq:
Removing some user busywork!

- Zip checkbox is now checked by default - when was the last time we wanted to upload a zip as a single file?
- If any file in the upload is named "index.html", fill that in as the Custom[Solution]URL (unless there's already a value there, in case that value has custom args etc)

### DIFF
--- a/ServerCore/Pages/Puzzles/FileManagement.cshtml.cs
+++ b/ServerCore/Pages/Puzzles/FileManagement.cshtml.cs
@@ -288,6 +288,36 @@ namespace ServerCore.Pages.Puzzles
         }
 
         /// <summary>
+        /// Promotes files named "index.html" to the CustomURL or CustomSolutionURL properties of the puzzle.
+        /// </summary>
+        /// <param name="file">The file being uploaded</param>
+        private void UpdateCustomURLs(ContentFile file)
+        {
+            // Skip this if:
+            // - the file is not named index.html
+            // - the file is not a material or solution file
+            // - the appropriate Custom[Solution]URL is non-blank
+            if (!file.ShortName.EndsWith("index.html") ||
+                !(file.FileType == ContentFileType.PuzzleMaterial || file.FileType == ContentFileType.SolveToken) || 
+                (file.FileType == ContentFileType.PuzzleMaterial && !string.IsNullOrEmpty(Puzzle.CustomURL)) ||
+                (file.FileType == ContentFileType.SolveToken && !string.IsNullOrEmpty(Puzzle.CustomSolutionURL)))
+            {
+                return;
+            }
+
+            // lack of translation of {eventId} is very intentional - this is translated at runtime and makes the field value portable
+            string filePath = "/{eventId}/Files/" + file.ShortName;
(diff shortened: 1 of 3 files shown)


ServerCore/Pages/Puzzles/FileManagement.cshtml.cs
```diff
@@ -287,6 +287,36 @@ private bool PuzzleExists(int puzzleId)
...
+        /// Promotes files named "index.html" to the CustomURL or CustomSolutionURL properties of the puzzle.
+        /// </summary>
+        /// <param name="file">The file being uploaded</param>
+        private void UpdateCustomURLs(ContentFile file)
+        {
+            // Skip this if:
+            // - the file is not named index.html
+            // - the file is not a material or solution file
+            // - the appropriate Custom[Solution]URL is non-blank
+            if (!file.ShortName.EndsWith("index.html") ||
```
digit01Wave, line 300: [nit]: These are a lot of "!" which can be a little hard to read.

Perhaps call it
bool shouldUpdateCustomUrl = file.ShortName.EndsWith(index.html) && (file.FileType == solution || fileType == Solve) && etc.

if (!shouldUpdateCustomUrl){ return}

digit01Wave (approved):

ServerCore/Pages/Puzzles/FileManagement.cshtml.cs
morganbr, line 300: I'd suggest using Path.GetFileName for this. I also wonder if there's other relevant info about the path... does anything generate multiple files named index.html?

ServerCore/Pages/Puzzles/FileManagement.cshtml.cs
```diff
@@ -287,6 +287,36 @@ private bool PuzzleExists(int puzzleId)
...
+                !(file.FileType == ContentFileType.PuzzleMaterial || file.FileType == ContentFileType.SolveToken) || 
+                (file.FileType == ContentFileType.PuzzleMaterial && !string.IsNullOrEmpty(Puzzle.CustomURL)) ||
+                (file.FileType == ContentFileType.SolveToken && !string.IsNullOrEmpty(Puzzle.CustomSolutionURL)))
+            {
+                return;
+            }
+
+            // lack of translation of {eventId} is very intentional - this is translated at runtime and makes the field value portable
+            string filePath = "/{eventId}/Files/" + file.ShortName;
```
morganbr, line 309: Path.Combine would be better practice here

morganbr (approved):

ServerCore/Pages/Puzzles/FileManagement.cshtml.cs
tabascq: Except it adds the wrong slash...

ServerCore/Pages/Puzzles/FileManagement.cshtml.cs
tabascq: I don't feel that the alternate version is simpler in this case.

PR #1137, opened 2025-01-23
Fast user list

### LINKED ISSUE #1074: Filter user list
We should write a Blazor-backed user finder that allows us to more easily find users among our increasingly-unwieldy set of 17K+. The user list should have an edit field that lets you search for a user by name or email (both at once is probably best, for simplicity) and an initially blank results list that doesn't begin to search until you have typed 2-3 characters (+2 if the second character is a -, to handle t- and v-).

morganbr:
Replace the pickers for adding admins, authors, and team members with a Blazor user picker with a filter to do quick searching and allow adding multiple users quickly. Fixes #1074


ServerCore/Pages/Teams/AddMemberPicker.razor.cs
```diff
@@ -0,0 +1,53 @@
...
+                                    where member.Team.ID == TeamId
+                                    select member.Member).ToListAsync();
+        }
+
+        protected override async Task OnUserAddedAsync(int addedUserId)
+        {
+            Event ev = await (from evt in _context.Events
+                              where evt.ID == EventId
+                              select evt).SingleAsync();
+            await TeamHelper.AddMemberAsync(_context, ev, EventRole.admin, TeamId, addedUserId);
```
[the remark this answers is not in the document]
morganbr: Not a bug, this is an admin-only scenario and the page has auth.

ServerCore/Pages/Events/AddAdminOrAuthor.cshtml
```diff
@@ -7,48 +7,15 @@
...
-            <td>
-                @item.Name
-            </td>
-            <td>
-                @item.Email
-            </td>
-            <td>
-                @if (Model.addAdmin)
-                {
-                    <a asp-page-handler="AddAdmin" asp-route-userId=@item.ID>Add</a>
```
tabascq, line 44: This handler is dead code now, yes?

ServerCore/Pages/Events/AddAdminOrAuthor.cshtml
```diff
@@ -7,48 +7,15 @@
...
-                }
-                else
-                {
-                    <a asp-page-handler="AddAuthor" asp-route-userId=@item.ID>Add</a>
```
tabascq, line 48: Dead code?

ServerCore/Pages/Events/AddAdminOrAuthor.cshtml
morganbr: Yes -- this whole section is deleted in the diff

ServerCore/Pages/Events/AddAdminOrAuthor.cshtml
morganbr: Ditto -- this was all deleted

ServerCore/Pages/Events/AddAdminOrAuthor.cshtml
tabascq: These appear to be implemented in AddAdminOrAuthor.cshtml.cs lines 54-110, which is **not** in the diff. [The diff does clearly include the deletion of the old code for team members, though.]

morganbr: Actually fixed this time

tabascq (approved):

PR #1139, opened 2025-01-27
Migrate to the Azure.Storage.Blobs API

### LINKED ISSUE #863: Update Azure Storage package
The one we use right now is deprecated. Also the new one would allow bulk clones, which could fix event migration

morganbr:
Migrate to the Azure.Storage.Blobs API, replacing the deprecated WindowsAzure.Storage API. Fixes #863 .

### DIFF
--- a/ServerCore/FileManager.cs
+++ b/ServerCore/FileManager.cs
@@ -1,14 +1,12 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
-using System.Linq;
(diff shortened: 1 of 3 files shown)


tabascq (approved):
Hard to be super sure since I don't know the API myself, but *seems* legit...

tabascq:
Since I don't know if this needs other server magic, I won't merge/deploy unless/until you give an all clear to do so.

PR #1140, opened 2025-01-27
Fix Auto CustomURL Generation

tabascq:
The previously-generated CustomURL didn't parse as a URL, meaning any subsequent edits to the puzzle properties would generate a validation error.


tabascq:
FYI, I deployed this and it works exactly as intended. Works in local deployments too.

morganbr (approved):

Issue #1054, opened 2024-07-09
State map for SinglePlayerPuzzles


digit01Wave:
Really just want a way to see who solved how many puzzles in case we want to give an award to the person who has solved the most pregame puzzles

PR #1145, opened 2025-02-11
Add single player puzzle state map

digit01Wave:
#1054
### What you see when no puzzles are unlocked and no players have seen it
image
- Remember that SinglePlayerPuzzle states are created lazily so if a player has not tried to look at the puzzle page, a SinglePlayerPuzzleStatePerPlayer will not be created for them

### What you see after unlock one of the puzzles for all players
Same as section above

### What you see after player actually looks at the unlocked single player puzzle
image

### What you see when a player solves one of the puzzles
image

### DIFF
--- a/ServerCore/Pages/Events/SinglePlayerPuzzleStateMap.cshtml.cs
+++ b/ServerCore/Pages/Events/SinglePlayerPuzzleStateMap.cshtml.cs
@@ -1,0 +1,159 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Identity;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.EntityFrameworkCore;
+using ServerCore.DataModel;
+using ServerCore.Helpers;
+using ServerCore.ModelBases;
+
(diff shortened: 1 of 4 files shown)


ServerCore/Pages/Events/SinglePlayerPuzzleStateMap.cshtml.cs
```diff
@@ -0,0 +1,159 @@
...
+
+        public class StateStats
+        {
+            public static StateStats Default => new StateStats();
+
+            public DateTime? UnlockedTime { get; set; }
+            public DateTime? SolvedTime { get; set; }
+            public bool LockedOut { get; set; }
+
+            public string Classes
```
tabascq, line 129: I get that this is a simplified version of the function of same name in Map.cshtml - but can it be shared, so that if we make updates we wind up fixing them both at once? Maybe StateStats gets shared and the SinglePlayer version just provides default values for some of the properties it doesn't care about?

tabascq (commented):
Looks great! Just the one comment related to code sharing.

ServerCore/Pages/Events/SinglePlayerPuzzleStateMap.cshtml.cs
digit01Wave: done

tabascq (approved):

PR #1146, opened 2025-02-11
Bulk File Uploader

tabascq:
Uploads one or more puzzles/solutions from a zip file, adding files as necessary. Does updating and adds new files, but can't be used for the first upload of a puzzle/solution since it uses the results of a previous upload to build the proper mapping.

### DIFF
--- a/ServerCore/Pages/Puzzles/BulkFiles.cshtml
+++ b/ServerCore/Pages/Puzzles/BulkFiles.cshtml
@@ -21,8 +21,7 @@
     <hr />
     <h3>What the Bulk File Uploader WILL do</h3>
     <ul>
-        <li>Upload new versions of files for puzzles/solutions, when stored in folders that match the folder names used previously</li>
-        <li>Correctly handle new files added to existing puzzles and solutions</li>
+        <li>Upload and add/replace files for puzzles/solutions, when stored in folders that match the folder names used previously for that puzzle</li>
         <li>Drill into root folders named "puzzles" and/or "solutions"</li>
     </ul>
     <h3>What the Bulk File Uploader WILL NOT do</h3>
--- a/ServerCore/Pages/Puzzles/BulkFiles.cshtml
+++ b/ServerCore/Pages/Puzzles/BulkFiles.cshtml
@@ -1,0 +1,66 @@
+﻿@page  "/{eventId}/{eventRole}/Puzzles/BulkFiles/{handler?}"
+@model ServerCore.Pages.Puzzles.BulkFilesModel
+
+@{
+    ViewData["Title"] = "Bulk File management";
+    ViewData["AdminRoute"] = "../Puzzles/BulkFiles";
+    ViewData["AuthorRoute"] = "../Puzzles/BulkFiles";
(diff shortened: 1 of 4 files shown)


ServerCore/Pages/Puzzles/BulkFiles.cshtml
```diff
@@ -0,0 +1,66 @@
...
+    <a asp-page="/Puzzles/Index">Back to puzzle list</a>
+</p>
+
+<div>
+    <hr />
+    <h3>What the Bulk File Uploader WILL do</h3>
+    <ul>
+        <li>Upload new versions of files for puzzles/solutions, when stored in folders that match the folder names used previously</li>
+        <li>Correctly handle new files added to existing puzzles and solutions</li>
+        <li>Drill into root folders named "puzzles" and/or "solutions"</li>
```
digit01Wave, line 26: Can you say if this will be case sensitive or not case-sensitive?

ServerCore/Pages/Puzzles/BulkFiles.cshtml
```diff
@@ -0,0 +1,66 @@
...
+
+<p>
+    <a asp-page="/Puzzles/Index">Back to puzzle list</a>
+</p>
+
+<div>
+    <hr />
+    <h3>What the Bulk File Uploader WILL do</h3>
+    <ul>
+        <li>Upload new versions of files for puzzles/solutions, when stored in folders that match the folder names used previously</li>
```
digit01Wave, line 24: [nit]: "Upload and override" so it lets people know that the these will replace the old versions

ServerCore/Pages/Puzzles/BulkFiles.cshtml.cs
```diff
@@ -0,0 +1,182 @@
...
+
+            return Page();
+        }
+
+        /// <summary>
+        /// Helper for taking an uploaded form file, uploading it, and tracking it in the database
+        /// </summary>
+        private async Task UploadFileAsync(IFormFile uploadedFile)
+        {
+            if (Path.GetExtension(uploadedFile.FileName).Equals(".zip", StringComparison.OrdinalIgnoreCase))
```
digit01Wave, line 70: [question]: So if the file is not a zip, nothing will happen?

ServerCore/Pages/Puzzles/BulkFiles.cshtml
tabascq: Honestly I'm not sure and don't know how to be sure without deploying it, since the local emulators sometimes have different case sensitivity rules.

ServerCore/Pages/Puzzles/BulkFiles.cshtml
tabascq: Clarifying this thx

ServerCore/Pages/Puzzles/BulkFiles.cshtml.cs
tabascq: Correct - I will make it report this via FilesIgnored, for clarity.

digit01Wave (approved):

PR #1147, opened 2025-02-12
Add allow-modals to iframes

tabascq:
allow-modals allows for confirm() to pop up a modal box from within an <iframe>.

This is used for older puzzles (that didn't save state) when navigating away, and for newer puzzles where the latest puzzle.js pops up a confirm on pressing Reset.


jbodner09 (approved):
Canvas also needs this since I added a roadblock before resetting its state too!

PR #1149, opened 2025-02-19
Alpha Testing Preview Fix

tabascq:
Forgot to push the visibility enhancement into FilesController - before this fix, you had to be an alpha event admin to see puzzle previews.

### DIFF
--- a/ServerCore/Pages/FilesController.cs
+++ b/ServerCore/Pages/FilesController.cs
@@ -47,7 +47,7 @@ namespace ServerCore.Pages
             // * Players can see puzzles and materials in alpha-testing events when the puzzle needs alpha tests
             // * Players can see answers after the event's AnswersAvailable time
             // * Players can see solve tokens on puzzles they've solved
-            if (!(eventObj.IsAlphaTestingEvent && content.Puzzle.AlphaTestsNeeded > 0) && !await IsAuthorized(eventObj.ID, content.Puzzle, content))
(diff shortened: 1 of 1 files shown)


ServerCore/Pages/FilesController.cs
```diff
@@ -44,9 +44,10 @@ public async Task<IActionResult> Index(string eventId, string filename)
             // * Admins have access to all files in their event
             // * Authors have access to all files attached to puzzles they own
             // * Players can see puzzles and materials on puzzles they've unlocked
+            // * Players can see puzzles and materials in alpha-testing events when the puzzle needs alpha tests
             // * Players can see answers after the event's AnswersAvailable time
             // * Players can see solve tokens on puzzles they've solved
-            if (!await IsAuthorized(eventObj.ID, content.Puzzle, content))
+            if (!(eventObj.IsAlphaTestingEvent && content.Puzzle.AlphaTestsNeeded > 0) && !await IsAuthorized(eventObj.ID, content.Puzzle, content))
```
digit01Wave, line 50: Wouldn't this addition be better in the IsAuthorized method?

digit01Wave (approved):

ServerCore/Pages/FilesController.cs
tabascq: I'll have to change the first argument of IsAuthorized from an ID to an Event, but sure that makes sense and the resulting clarity is worth the slightly larger diff.

PR #1168, opened 2025-04-08
Fix Alpha crash

tabascq:


morganbr (changes requested):
I don't think sync will like DateTime.MaxValue or MinValue.I have a fix I previously forgot to push, now up at https://github.com/PuzzleServer/mainpuzzleserver/pull/1169

tabascq:
#1169 is a better fix, thanks

PR #1181, opened 2025-04-29
Change notification icon from envelope to bell

tabascq:


asyasky (approved):

ServerCore/Pages/Shared/_EventNavigationPartial.cshtml
```diff
@@ -127,7 +127,7 @@
             @if (Event?.AllowBlazor == true && Event?.EphemeralHackKillNotifications != true)
             {
                 <a class="nav-link position-relative" asp-page="/Events/Notifications">
-                    <span class="fab text-white">✉️</span>
+                    <span class="fab text-white">✉🔔</span>
```
Vroo, line 130: Why do two of the lines in this PR have ✉🔔 rather than just 🔔?

tabascq: Argh, working too fast will fix

tabascq: Should be fixed with newly-deployed #1183.

Issue #1173, opened 2025-04-21
Multiple pages do not render Html.Raw properly


Vroo:
`https://puzzlehunt.azurewebsites.net/.../play/Teams/.../Answers` does not handle Html.Raw nor does it respect the value of TermForGroup in hunt details.

Vroo:
Raw html is incorrectly displayed in both the group name and the puzzle name.

Vroo:
Notifications do not recognize Html.Raw and worse insert the title as HTML rather than escaped text. So a `<` in a puzzle title will break it.

![Image](https://github.com/user-attachments/assets/89d62b81-1f25-462d-969b-ceaa952186b7)

Vroo:
And feedback and threads pages

```
https://puzzlehunt.azurewebsites.net/.../play/Threads/PuzzleThread/...
https://puzzlehunt.azurewebsites.net/.../play/Puzzles/SubmitFeedback/...
```

bruceleban:
`https://puzzlehunt.azurewebsites.net/ph24beta/admin/Teams/.../Status`

morganbr:
Fixed in #1192

PR #1192, opened 2025-05-12
Html.Raw puzzle title support

tabascq:
Adding support for the raw syntax to a whole mess of scenarios. Should fix most/all issues in #1173 and a whole mess of unreported issues.

Note that a lot of these changes use puzzle.PlaintextName, which is already used in many places, defined as follows, and must be provided for this to work:

        /// The plaintext name of the puzzle, which is a subset of the name when the name is of the form:
        /// {plaintextName}Html.Raw({htmlName})

### DIFF
--- a/ServerCore/Pages/Events/FastestSolves.cshtml.cs
+++ b/ServerCore/Pages/Events/FastestSolves.cshtml.cs
@@ -84,7 +84,7 @@ namespace ServerCore.Pages.Events
                                     !pspt.Team.IsDisqualified &&
                                     pspt.Puzzle.Event == Event &&
                                     (EventRole != EventRole.author || authorPuzzleIds.Contains(pspt.PuzzleID))
-                                    let puzzleToGroup = new { PuzzleID = pspt.Puzzle.ID, PuzzleName = pspt.Puzzle.Name } // Using 'let' to work around EF grouping limitations (https://www.codeproject.com/Questions/5266406/Invalidoperationexception-the-LINQ-expression-for)
+                                    let puzzleToGroup = new { PuzzleID = pspt.Puzzle.ID, PuzzleName = pspt.Puzzle.PlaintextName } // Using 'let' to work around EF grouping limitations (https://www.codeproject.com/Questions/5266406/Invalidoperationexception-the-LINQ-expression-for)
                                     group puzzleToGroup by puzzleToGroup.PuzzleID into puzzleGroup
                                     orderby puzzleGroup.Count() descending, puzzleGroup.Max(puzzleGroup => puzzleGroup.PuzzleName) // Using Max(PuzzleName) because only aggregate operators are allowed
                                     select new { PuzzleID = puzzleGroup.Key, PuzzleName = puzzleGroup.Max(puzzleGroup => puzzleGroup.PuzzleName), SolveCount = puzzleGroup.Count() }).ToListAsync();
@@ -223,7 +223,7 @@ namespace ServerCore.Pages.Events
                                      let puzzleToGroup = new
                                      {
                                          PuzzleID = statePerPlayer.Puzzle.ID,
-                                         PuzzleName = statePerPlayer.Puzzle.Name,
+                                         PuzzleName = statePerPlayer.Puzzle.PlaintextName,
                                          PuzzleUserId = statePerPlayer.PlayerID,
                                          IsUnlocked = statePerPlayer.UnlockedTime.HasValue,
                                          SolveTime = statePerPlayer.SolvedTime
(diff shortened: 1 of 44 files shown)


tabascq:
> One issue which I suspect should be fixed elsewhere separately.

What issue, @bruceleban ?

bruceleban:
The notification text being used as raw html.

tabascq:
> The notification text being used as raw html.

The last two changes in this PR (in PuzzleStateHelper.cs) should fix that.

bruceleban:
The toast code renders notification assuming the sender is sending html, not plain text. So in a TV hunt, if the title of a puzzle is `</scorpion>` for example, it will render incorrectly in notifications.

Took me a while to find it because there's a long chain but look at _EventNavigationPartial.cshtml:popToast. I don't know if the error is that this code should be sending html 
```
... SendNotification(t, "New puzzle!", System.Security.SecurityElement.Escape($"{puzzle.Name} has been unlocked!")), 
```
or popToast should be:
```
<div> ... ${System.Security.SecurityElement.Escape(notification.content)} </div>
```
Are there notifications that need html to render properly? If not, the fix should go in popToast. I also think the code in popToast should definitely be escaping the url.

tabascq:
If the `plaintextName` version contains HTML codes it will break dozens of things, not just notifications. If it's important to PH24 to do this, let's talk, and if not, let's not worry about it.

bruceleban:
We don't need that for ph24. I can't help it. Anytime I see unescaped html I worry. I wasn't sure that user data never appears in notifications.

ServerCore/Pages/Teams/Answers.cshtml
```diff
@@ -22,7 +22,7 @@
             </th>
             <th>
                 <a asp-page="./Answers" asp-route-teamId="@Model.TeamID" asp-route-sort="@(Model.SortForColumnLink(AnswersModel.SortOrder.GroupAscending, AnswersModel.SortOrder.GroupDescending))">
-                    Group
+                    @(Model.Event.TermForGroup ?? "Group")
```
jbodner09, line 25: While you're updating stuff like this, does https://github.com/PuzzleServer/mainpuzzleserver/blob/main/ServerCore/Pages/Puzzles/Delete.cshtml#L67 need this update as well?

tabascq: I decided to leave the CRUD pages unchanged to make absolutely sure that editing the raw HTML syntax worked correctly.

ServerCore/Pages/Teams/Hints.cshtml.cs
```diff
@@ -147,7 +147,7 @@ private async Task PopulateUI(int puzzleId, int teamId)
                 where Puzzle.ID == puzzleId
                 select Puzzle).FirstOrDefaultAsync();
 
-            PuzzleName = puzzle?.Name;
+            PuzzleName = puzzle?.PlaintextName;
```
morganbr, line 150: Since this is player-facing, should it be html?

morganbr (approved):
Thanks for the thorough fix!

Vroo:
One thing I just noticed is that the page title also has raw html in it. I don't know if this is fixed. I searched for "title" and couldn't see any place where that would have been fixed but I don't know where the title is getting set.

tabascq:
> One thing I just noticed is that the page title also has raw html in it. I don't know if this is fixed. I searched for "title" and couldn't see any place where that would have been fixed but I don't know where the title is getting set.

Can you provide an example? There are dozens of pages...

tabascq:
Oh wait - I see the issue in the puzzle sumbissions page, assuming that one

ServerCore/Pages/Events/FastestSolves.cshtml.cs
```diff
@@ -84,7 +84,7 @@ private async Task updateTeamPuzzleStats(HashSet<int> authorPuzzleIds)
                                     !pspt.Team.IsDisqualified &&
                                     pspt.Puzzle.Event == Event &&
                                     (EventRole != EventRole.author || authorPuzzleIds.Contains(pspt.PuzzleID))
-                                    let puzzleToGroup = new { PuzzleID = pspt.Puzzle.ID, PuzzleName = pspt.Puzzle.Name } // Using 'let' to work around EF grouping limitations (https://www.codeproject.com/Questions/5266406/Invalidoperationexception-the-LINQ-expression-for)
+                                    let puzzleToGroup = new { PuzzleID = pspt.Puzzle.ID, PuzzleName = pspt.Puzzle.PlaintextName } // Using 'let' to work around EF grouping limitations (https://www.codeproject.com/Questions/5266406/Invalidoperationexception-the-LINQ-expression-for)
```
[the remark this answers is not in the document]
tabascq: Honestly terrified of breaking your event, being ultra safe.

ServerCore/Pages/Teams/Hints.cshtml.cs
tabascq: Honestly terrified of breaking your event, being ultra safe.

Vroo:
Yup that's it. Here's an example.
https://puzzlehunt.azurewebsites.net/sandbox/play/Submissions/7365
in player mode.

PR #1196, opened 2025-05-20
Filter remote/local puzzles

morganbr:
Optionally filter puzzles for remote and local teams so they don't see puzzles they can't solve


tabascq:
Cases I do not see:
- PlayModel.GetVisibleTeamPlayerPuzzleViews (If you are thinking it's OK to skip this because the puzzle will never be unlocked, clicking "Unlock All" will bite you, and possibly some scenarios through PuzzleStateHelper.UnlockAnyPuzzlesThatThisSolveUnlockedAsync)
- Anything single-player (maybe a note on the Edit page that this only impacts team puzzles, or maybe plumb it through to single-player cases)

asyasky (approved):

PR #1197, opened 2025-05-20
Fix bulk hints crash

tabascq:
The single-player-puzzle check in OnPostAsync was assuming that Puzzle was initialized, but it wasn't.


tabascq:
This has been broken for just shy of 2 years BTW

morganbr (approved):

PR #1204, opened 2025-05-27
Add whether or not a team is remote to the TeamComposition summary page

asyasky:


ServerCore/Pages/Teams/TeamCompositionSummary.cshtml.cs
```diff
@@ -71,7 +72,8 @@ public TeamComposition(int teamId)
         public int TotalNonMicrosoftCount { get; set; }
         public int TotalTotal { get; set; }
         public int TotalPossibleEmployeeAliasesCount { get; set; }
-        
+        public bool TeamIsRemote { get; set; }
```
morganbr, line 75: It doesn't look like this is getting used. It would be helpful to have a count like the other counts here though

morganbr (approved):

PR #1206, opened 2025-05-27
Adjust time on server, proof of concept

tabascq:
This version has no expiry, so if a user changes time zones without the server front end restarting, they will get the old time zone.

Uses the previous client-based mechanism for the first render, as well as for non-Blazor scenarios.

### DIFF
--- a/ServerCore/Components/NotificationComponent.razor.cs
+++ b/ServerCore/Components/NotificationComponent.razor.cs
@@ -2,6 +2,7 @@
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Components;
 using Microsoft.JSInterop;
+using ServerCore.Helpers;
 
 namespace ServerCore.Components
 {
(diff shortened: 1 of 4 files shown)


tabascq:
Our existing client code for updating time is this code in site.js:

```
document.addEventListener("DOMContentLoaded", e => {
    document.querySelectorAll("time").forEach(t => {
        var utctimeval = t.innerText;
        var date = new Date(utctimeval);
        t.innerText = date.toLocaleString();
    });
});
```

which seems awfully similar to your suggestion yet doesn't seem to perform well.

Vroo:
It is although that code uses the _**text**_ inside the `<time>` element instead of the defined `datetime` attribute which means it's not idempotent like my code is. Doing a little research it appears the actual issue is we should be setting `t.textContent` instead of `t.innerText`. When we're all done, we just do `document.body.offsetHeight;` to trigger a reflow.

tabascq:
Fixed this using the textContent suggestion above

PR #1208, opened 2025-05-28
LocalTime Perf Fix

tabascq:
Short-term fix to improve the performance of our client-side LocalTime implementation. Doesn't improve the story for future Blazor scenarios, but it's a start.


asyasky (approved):
Tested these locally with the large size demo event, looks great!

PR #1209, opened 2025-05-29
Remote vs local team caps and various bug fixes

asyasky:
-Add remote/local team max number and supporting UI/logic
-Fix missing player registration when admin adds player to team
-Prioritize unassigned PlayerClasses when admin adds a player to team
-Fix broken dropdown for Temp PlayerClass if TemporaryPlayerClass is null
-Use PlayerClass when TempPlayerClass is hidden even if TempPlayerClass is set


ServerCore/Helpers/TeamHelper.cs
```diff
@@ -35,6 +36,16 @@ public static async Task CreateTeamAsync(PuzzleServerContext context, Team team,
...
+                    {
+                        throw new Exception("Remote team registration is full. No further remote teams may be created at the present time.");
+                    }
+
+                    if (!team.IsRemoteTeam && await context.Teams.Where((t) => t.Event == ev && !t.IsRemoteTeam).CountAsync() >= ev.MaxNumberOfLocalTeams)
+                    {
+                        throw new Exception("Local team registration is full. No further local teams may be created at the present time.");
+                    }
+
                     if (await context.Teams.Where((t) => t.Event == ev).CountAsync() >= ev.MaxNumberOfTeams)
```
tabascq, line 49: Unfortunate that all three of these max numbers exist, and all three need to be set, and if you forget one then you get weird results. Can we just have two properties plus math to get the third, perhaps by deleting MaxNumberOfLocalTeams and computing it as MaxNumberOfTeams - MaxNumberOfRemoteTeams?

ServerCore/Pages/Teams/CreateTeam.razor
```diff
@@ -14,21 +14,42 @@
...
+                            This event allows both local and remote teams.
+                            @if(Event.LockChangesToRemoteStatus)
+                            {
+                                <div> Once the team is created this can only be changed by contacting the organizers.</div>
+                            }
+                            <table>
+                                <tr>
+                                    <InputSelect @bind-Value="TeamModel.IsRemoteTeam">
+                                        <option value=null>Select An Option</option>
+                                        @if (LocalTeamsFull)
```
tabascq, line 29: Does this work?

`<option @(LocalTeamsFull ? "disabled" : "") value=false>Local</option>`

ServerCore/Pages/Teams/Edit.cshtml
```diff
@@ -65,12 +65,15 @@
             </div>
             @if(Model.Event.AllowsRemoteTeams)
             {
-                <div class="form-group">
-                    <label asp-for="Team.IsRemoteTeam" class="control-label">Team is playing remotely</label>
+                if(!Model.Event.LockChangesToRemoteStatus)
```
tabascq, line 68: Since there are no changes to the Edit.cshtml.cs file, I assume that I could get around remote/local limits by declaring my team in the bucket that has room and then editing the team to toggle this flag.

asyasky: This hides the UI for making that edit and then if the player isn't an admin they'll bounce off of the change even if they do it through F12 or something since only admins can ignore the team caps.

ServerCore/Pages/Teams/CreateTeam.razor
```diff
@@ -1,6 +1,6 @@
 <div class="row">
     <div class="col-md-4">
-        @if (!EventFull)
+        @if (!EventFull || !(LocalTeamsFull && RemoteTeamsFull))
```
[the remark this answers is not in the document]
asyasky: It does not, but it'll work for Hunt and will prevent any existing events from having trouble so I added this bug to my list and will handle it before wrapping up tech work on the feature.

ServerCore/Helpers/TeamHelper.cs
tabascq: (they'll break because the local and remote maxes will be zero, effectively locking registration)

asyasky: right you are, sorry, I caught that in the other check but not in this one, I'll tweak things.

ServerCore/Pages/Teams/Edit.cshtml
tabascq: If you're resolving this because for PH24 the flag is true thus nothing can happen in the next week, fine, but unless I'm fundamentally misunderstanding things, this needs to be fixed eventually.

ServerCore/Pages/Teams/CreateTeam.razor.cs
```diff
@@ -128,6 +128,11 @@ protected override async Task OnParametersSetAsync()
                 RemoteTeamsFull = true;
             }
 
+            if (!Event.AllowsRemoteTeams)
```
tabascq, line 131: Do you need to do this before the checks above?

ServerCore/Helpers/TeamHelper.cs
```diff
@@ -36,17 +36,17 @@ public static async Task CreateTeamAsync(PuzzleServerContext context, Team team,
...
                     }
 
-                    if (!team.IsRemoteTeam && await context.Teams.Where((t) => t.Event == ev && !t.IsRemoteTeam).CountAsync() >= ev.MaxNumberOfLocalTeams)
+                    if (ev.MaxNumberOfLocalTeams > 0 && !team.IsRemoteTeam && await context.Teams.Where((t) => t.Event == ev && !t.IsRemoteTeam).CountAsync() >= ev.MaxNumberOfLocalTeams)
                     {
                         throw new Exception("Local team registration is full. No further local teams may be created at the present time.");
                     }
 
-                    if (await context.Teams.Where((t) => t.Event == ev).CountAsync() >= ev.MaxNumberOfTeams)
+                    if (ev.MaxNumberOfTeams > 0 && await context.Teams.Where((t) => t.Event == ev).CountAsync() >= ev.MaxNumberOfTeams)
```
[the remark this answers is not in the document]
asyasky: Don't we have a registration setting for that?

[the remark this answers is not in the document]
asyasky: Oh, is this a beta thing?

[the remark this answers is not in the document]
tabascq: You're right and I rescind this comment.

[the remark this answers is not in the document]
asyasky: Are you sure? I don't want to mess with beta set ups or other weird testing scenarios, so I pulled this check for the moment and made a note to check the event history to see if it's come up.

tabascq (approved):

PR #1211, opened 2025-05-30
Fix autoteams

morganbr:
Fix auto teams radio and signup hub radio buttons that didn't show that they were clicked (even though they were), differentiate local and remote auto-teams and allow remote auto-teams for non-MS

### DIFF
--- a/ServerCore/Pages/Teams/AutoTeam.razor.cs
+++ b/ServerCore/Pages/Teams/AutoTeam.razor.cs
@@ -21,6 +21,9 @@ namespace ServerCore.Pages.Teams
         [Parameter]
         public EventRole EventRole { get; set; }
 
+        [Parameter]
+        public bool IsMicrosoft { get; set; }
+
         [Inject]
         public PuzzleServerContext _context { get; set; }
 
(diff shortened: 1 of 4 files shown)


ServerCore/Pages/Teams/AutoTeam.razor.cs
```diff
@@ -89,9 +110,19 @@ where teamGroup.Count() < maxAutoTeamSize
...
                 };
 
-                await TeamHelper.CreateTeamAsync(_context, team, Event, LoggedInUserId);
+                try
+                {
+                    await TeamHelper.CreateTeamAsync(_context, team, Event, LoggedInUserId);
+                }
+                catch
+                {
+                    // Most exceptions from here are because the team type isn't available
```
tabascq, line 122: Do you need to tell the user this?

asyasky (approved):

ServerCore/Pages/Teams/AutoTeam.razor.cs
```diff
@@ -89,9 +110,19 @@ where teamGroup.Count() < maxAutoTeamSize
...
+                    CantCreateTeam = true;
+                    return;
+                }
```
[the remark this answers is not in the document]
morganbr: CreateTeamAsync throws generic exceptions for the cases we care about. It'd be great to fix those another time.

ServerCore/Pages/Teams/AutoTeam.razor.cs
morganbr: It will show "All teams with that choice are full. Please try another or apply to an existing team". Is that good enough?

PR #1212, opened 2025-05-30
Remove large margins on mobile

morganbr:
Remove large margins on mobile to give the puzzle and answer box more space

### DIFF
--- a/ServerCore/Pages/Shared/_Layout.cshtml
+++ b/ServerCore/Pages/Shared/_Layout.cshtml
@@ -144,7 +144,7 @@
         <partial name="/Pages/Shared/_EventNavigationPartial.cshtml" />
     }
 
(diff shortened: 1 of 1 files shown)


bruceleban (approved):
We should also add:
```
#submitBox {
    min-width: 100px;
}
```
or the equivalent somewhere.

ServerCore/Pages/Shared/_Layout.cshtml
```diff
@@ -144,7 +144,7 @@
         <partial name="/Pages/Shared/_EventNavigationPartial.cshtml" />
     }
 
-    <div class="container body-content body-content-customizable">
+    <div class="container-md body-content body-content-customizable">
```
tabascq, line 147: Is this something we really want to do for all mobile pages on all events?

PR #1218, opened 2025-06-04
Admin quality of life updates

asyasky:
- Add Custom Author Text and Prerequisite Weight to the Puzzle Checklist
- Expand non-alias checks on TeamCompositionSummary - filters out things that are obviously not Microsoft aliases and doesn't display them as possible aliases on the page (initial dataset is based on common words input by players in the current event)
- Add remote/local team information to Puzzle Status page
- Add puzzle availability to puzzle list for author/admin

### DIFF
--- a/ServerCore/Pages/Teams/AllTeams.cshtml
+++ b/ServerCore/Pages/Teams/AllTeams.cshtml
@@ -59,7 +59,7 @@ Registered Teams: <b>@Model.Teams.Count/@Model.Event.MaxNumberOfTeams</b>, Regis
                     <td>
                         if (item.IsRemoteTeam)
                         {
-                        <text>Remote</text>
+                            <text>Remote</text>
                         }
                         else
                         {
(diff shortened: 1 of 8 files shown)


ServerCore/Pages/Teams/AllTeams.cshtml
```diff
@@ -53,6 +56,16 @@ Registered Teams: <b>@Model.Teams.Count/@Model.Event.MaxNumberOfTeams</b>, Regis
                         <text>Disqualified</text>
                     }
                     </td>
+                    <td>
+                        if (item.IsRemoteTeam)
```
tabascq, line 60: Is this OK without an @? If not, why not change it to this nicer format from elsewhere in this PR:

```
@item.Team.IsRemoteTeam ? Remote : Local
```

morganbr (approved):

ServerCore/Pages/Teams/AllTeams.cshtml
asyasky: image
Alternatively, it could be that that's not actually the page that any uses and it's been broken for 6 years. Working on editing the correct page now.

asyasky: And also fixing that ternary statement that I could have sworn I fixed already (must have lost a change somewhere)

tabascq (approved):

PR #1222, opened 2025-06-06
Split leaderboard remote vs local

asyasky:
Add a filter on the leaderboard page to show remote/local/all teams. The standings are calculated on the fly so only UI changes are needed.

### DIFF
--- a/ServerCore/Pages/Events/Standings.cshtml
+++ b/ServerCore/Pages/Events/Standings.cshtml
@@ -10,6 +10,19 @@
 
 <h2>Standings</h2>
 
+@if(Model.Event.AllowsRemoteTeams)
+{ 
+    <p>
+        Filter standings to remote/local/all teams
+    </p>
+    <form method="GET" name="locationForm">
+        <select name="locationFilter" asp-items="@Model.TeamLocationFilter" onchange="this.form.submit();">
+            <option value="">Select Filter</option>
+        </select>
+    </form>
+    <div>@Model.TeamLocationFilter Teams</div>
+}
+
 <table class="table table-condensed">
     <thead>
         <tr>
--- a/ServerCore/Pages/Events/Standings.cshtml
+++ b/ServerCore/Pages/Events/Standings.cshtml
@@ -20,7 +20,8 @@
             <option value="">Select Filter</option>
         </select>
     </form>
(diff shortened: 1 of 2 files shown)


ServerCore/Pages/Events/Standings.cshtml
```diff
@@ -10,6 +10,20 @@
 
 <h2>Standings</h2>
 
+@if(Model.Event.AllowsRemoteTeams)
+{ 
+    <p>
+        Filter standings to remote/local/all teams
+    </p>
+    <form method="GET" name="locationForm">
+        <select name="locationFilter" asp-items="@Model.TeamLocationFilter" onchange="this.form.submit();">
```
tabascq, line 19: Does submitting here preserve the sort order?

ServerCore/Pages/Events/Standings.cshtml.cs
```diff
@@ -17,15 +21,29 @@ public class StandingsModel : EventSpecificPageModel
...
+        }
+
         private const SortOrder DefaultSort = SortOrder.RankAscending;
 
         public StandingsModel(PuzzleServerContext serverContext, UserManager<IdentityUser> userManager) : base(serverContext, userManager)
         {
         }
 
-        public async Task OnGetAsync(SortOrder? sort)
+        public async Task OnGetAsync(SortOrder? sort, TeamLocation locationFilter)
```
tabascq, line 43: Is this a required parameter since it is not nullable? Does that mean that links to this page wouldn't work?

asyasky: That is an excellent question that did not occur to me, let me check.

ServerCore/Pages/Events/Standings.cshtml
asyasky: It does not. Do you know of an easy way to do that with the dropdown form submit? I'm happy to make quick changes otherwise I'll file a bug for post-event.

tabascq: Nope, sorry. Does it *have* to be a drop-down? We have lots of examples of hyperlinks that add in all the needed properties, such as the sort links on many pages.

morganbr (approved):

PR #1277, opened 2025-09-06
Allow Puzzle to Submit Answer

tabascq:
Support to allow a puzzle to submit an answer if it so chooses.

[Can be combined with PuzzleJS validation to only submit for a correctly-solved puzzle, using a hash, for puzzles without human-readable answers.]


tabascq:
Will the REST API immediately show the puzzle as solved on the puzzle page itself? Is the URL necessary for the REST API straightforward to derive without modifying the CustomURL? If the answer to both of these is yes then I agree this is superfluous.

tabascq:
I just didn't want to have to put the text into the guess box like the click handler would require. I think it's just a stylistic choice.

tabascq:
Closing in favor of the other approach

Issue #1115, opened 2024-11-12
Remove isBlazorDisabled check


vidapour:

vidapour:
Includes cleaning up duplicate Blazor-disabled pages

PR #1288, opened 2025-11-29
Remove references to allowBlazor from code

digit01Wave:
#1115
This PR 
- defaults embedded and blazor to true on creation
- removes references to Model.Event.AllowBlazor in code


tabascq:
Does this make the page for /Teams/List dead code, or is there another entry point to it?

tabascq:
> > Does this make the page for /Teams/List dead code, or is there another entry point to it?
> 
> It is referred to in some partial cshtmls from past events (nipd2020/homePartial.cshtml, nipd2021/homepartial.cshtml). To not break those i was going to leave it.

I don't think it is critically important to preserve the signup procedure for a five year old event, but why not change these to /Teams/Signup?

I'm just trying to keep us from needing to perpetually maintain code that nobody will ever run.

digit01Wave:
> > > Does this make the page for /Teams/List dead code, or is there another entry point to it?
> > 
> > 
> > It is referred to in some partial cshtmls from past events (nipd2020/homePartial.cshtml, nipd2021/homepartial.cshtml). To not break those i was going to leave it.
> 
> I don't think it is critically important to preserve the signup procedure for a five year old event, but why not change these to /Teams/Signup?
> 
> I'm just trying to keep us from needing to perpetually maintain code that nobody will ever run.

That is fair. I removed all references to /List and replaced it with signup

tabascq:
Thanks! I'd like @morganbr to ring in on this since he did the original signup conversion and may know of a wrinkle we have forgotten, but I don't see any issues. I'm not signing off just so we remember that Morgan has to look.

Issue #1248, opened 2025-06-24
Fix lunch order deadline


tabascq:
The lunch order deadline date evaluates via server time, not local time.

PR #1291, opened 2026-01-18
Fix Lunch Date

tabascq:
Rather than render the lunch deadline server-side, use the existing time-adjustment technique but add an optional format.


tabascq:
Fixes #1248

PR #1307, opened 2026-02-19
Support for hiding standings

tabascq:
Remove 3 nearly-entirely-nonfunctional properties of Event (one property was used on one line of Play.cshtml and was clearly incomplete in its implementation), and add two new properties that are actually functional. This seemed cleaner than actually wiring support for the old properties, and fewer dates is less work for admins.

The old ShowFastestSolves property (which, again, did not actually do what it advertised) is set on 11 old events, and will need to be reset on those events since that property is being recycled to mean the opposite. [I tried to hand-modify the migration code in this PR but that didn't work out, and tweaking these 11 events in SQL will be the least overall work by far.]


ServerCore/Pages/Events/Create.cshtml
```diff
@@ -64,15 +64,15 @@
             </div>
             <div class="description-column">The text shown after the copyright symbol at the bottom of every page on the site.</div>
             <div class="form-group">
-                <label asp-for="Event.ShowFastestSolves"></label>
-                <input asp-for="Event.ShowFastestSolves" />
+                <label asp-for="Event.HideFastestSolves"></label>
+                <input asp-for="Event.HideFastestSolves" />
             </div>
-            <div class="description-column">Shows a second standings page for the top 3 fastest solves per puzzle.</div>
+            <div class="description-column">Shows the number of times each puzzle is solved. If HideStandings is false, also shows the top 3 fastest solves per puzzle.</div>
```
digit01Wave: [nit]: Would it be clearer to just say "If HideFastestSolve is false, shows the top 3 fastest solves per puzzle"

ServerCore/Pages/Events/Edit.cshtml
```diff
@@ -70,15 +70,15 @@
             </div>
             <div class="description-column">The text shown after the copyright symbol at the bottom of every page on the site.</div>
             <div class="form-group">
-                <label asp-for="EditableEvent.ShowFastestSolves"></label>
-                <input asp-for="EditableEvent.ShowFastestSolves" />
+                <label asp-for="EditableEvent.HideFastestSolves"></label>
```
digit01Wave: Same comment as in Create.cshtml

digit01Wave (approved):

ServerCore/Pages/Events/Edit.cshtml
digit01Wave: nevermind

PR #1311, opened 2026-03-11
Mobile Fixes

tabascq:
Fixes several things:
- crash in some pages if you aren't part of a team
- some capitalization tweaks
- tools & techniques list now wraps on narrow screens
- puzzle author aligns better with rest of content at top of page
- content at top of page responds better to narrow widths
- lockout message now has working link
- team registration reminder on team list does not show if registration is closed


tabascq:
Note: There are other mobile fixes that are part of the event-specific repos, and thus not included here.

morganbr (approved):

PR #1317, opened 2026-04-09
Multiple options (toppings) per item (pizza)

tabascq:
Overall UX:
image

First dropdown (the set of options is an event property, not hardcoded):
image

Disabling on decline:
image

Second dropdown not showing duplicates:
image

Appearance of a double option when ordered:
image


ServerCore/Pages/Teams/Details.cshtml
```diff
@@ -250,15 +293,26 @@
...
                                 <option value="">Select an option</option>
-                                <option value="Skip Item">Skip Item</option>
+                                <option value="none">Decline Lunch Item</option>
                                 @foreach (var lunchOption in DetailsModel.LunchOptions)
                                 {
                                     <option value="@lunchOption">@lunchOption</option>
                                 }
                             </select>
-                            <button type="submit" asp-page-handler="AddLunch" asp-route-teamId="@Model.Team.ID" asp-route-newLunch="@Model.NewLunch">Add</button>
+                            @if (Model.Event.LunchOptionsPerLunch > 1)
```
morganbr: Only supporting one more item is a bit misleading here. I'd suggest binding to an array of items.

ServerCore/Pages/Teams/Details.cshtml
```diff
@@ -57,7 +57,50 @@
...
+
+        isUpdatingOptions = true;
+        let noneSeen = false;
+
+        document.querySelectorAll(".lunch-select option").forEach(option => {
+            option.classList.remove("hidden");
+        });
+
+        document.querySelectorAll(".lunch-select").forEach(select => {
+            if (noneSeen) { select.disabled = true; select.value = ""; return; }
```
morganbr, line 77: Does this prevent unselecting "Decline" if you ever select it?

ServerCore/Pages/Teams/Details.cshtml
```diff
@@ -57,7 +57,50 @@
         display: table;
         clear: both;
     }
+
+    .hidden { display: none; }
 </style>
+<script>
+    let isUpdatingOptions = false;
+
+    function updateOptions() {
```
morganbr, line 66: I imagine this worked for you, but it is the kind of pattern that tends to conflict with asp.net rendering. It might be a better idea to use Blazor if possible

ServerCore/Pages/Teams/Details.cshtml
```diff
@@ -57,7 +57,50 @@
...
+                });
+            }
+        });
+
+        isUpdatingOptions = false;
+    }
+
+    document.addEventListener("DOMContentLoaded", function () {
+        document.querySelectorAll(".lunch-select").forEach(select => {
+            select.addEventListener("change", function () { updateOptions(); });
```
morganbr, line 100: Does this behave correctly if you reload the page?

morganbr (changes requested):

ServerCore/Pages/Teams/Details.cshtml
tabascq: specifically check that graying etc works on load

ServerCore/Pages/Teams/Details.cshtml
tabascq: try this:
        `public string[] NewLunch { get; set; }`

and then `NewLunch[i]` in the `asp-for`

ServerCore/Pages/Teams/Details.cshtml
tabascq: No because at the beginning of the loop, noneSeen is false. It only gets set to true once it finds a "none", a couple of lines below this - so it never disables itself.

ServerCore/Pages/Teams/Details.cshtml
tabascq: Perhaps but this is what I have time for.

ServerCore/Pages/Teams/Details.cshtml
tabascq: On reload this clears out the uncommitted pizza choices, which is fine.

ServerCore/Pages/Teams/Details.cshtml
tabascq: This seemed to work! I had to use Request.Form to access the values though.

ServerCore/Pages/Teams/Details.cshtml
```diff
@@ -237,7 +280,7 @@
                         <p>@Model.Event.LunchDescription</p>
                     }
                     <p>
-                        Your team can currently have @Model.EditableLunches item(s). If your team wants to opt out of one of the items, select &quot;Skip Item&quot;. If you cannot have any of the options offered due to dietary restrictions, please email <a href="mailto:puzzhunt@microsoft.com">puzzhunt@microsoft.com</a> with details and we'll find an alternative.
+                        Your team can currently have @Model.EditableLunches item(s). If your team wants to opt out of one of the items, select &quot;Decline Lunch Item&quot;. If you cannot have any of the options offered due to dietary restrictions, please email <a href="mailto:puzzhunt@microsoft.com">puzzhunt@microsoft.com</a> with details and we'll find an alternative.
```
digit01Wave: [non-blocker]: I think we want this to be sentence cased like the other options (i.e. "Decline lunch item")

digit01Wave: Would it be clearer to say "...can currently have EditableLunchs lunch item(s)?" instead of just "items?"

ServerCore/Pages/Teams/Details.cshtml
```diff
@@ -250,15 +293,18 @@
...
-                                {
-                                    <option value="@lunchOption">@lunchOption</option>
-                                }
-                            </select>
-                            <button type="submit" asp-page-handler="AddLunch" asp-route-teamId="@Model.Team.ID" asp-route-newLunch="@Model.NewLunch">Add</button>
+                            @for (int i = 0; i < Model.Event.LunchOptionsPerLunch; i++)
+                            {
+                                <select class="lunch-select" asp-for="@Model.NewLunch[i]">
+                                    <option value="">Select an option</option>
+                                    <option value="none">@(i==0 ? "Decline Lunch Item" : "None")</option>
```
digit01Wave: [non-blocker]: Sentence case

digit01Wave (approved):

digit01Wave (approved):

PR #1318, opened 2026-05-12
Add database indexes for perf improvements

roger-wolfson:


Data/PuzzleServerContext.cs
```diff
@@ -67,12 +68,17 @@ public static void UpdateDatabase(IApplicationBuilder app)
         protected override void OnModelCreating(ModelBuilder modelBuilder)
         {
             modelBuilder.Entity<ContentFile>().HasIndex(contentFile => new { contentFile.EventID, contentFile.ShortName }).IsUnique();
+            modelBuilder.Entity<ContentFile>().HasIndex(contentFile => new { contentFile.EventID, contentFile.FileType}).IncludeProperties(contentFile => new { contentFile.ShortName, contentFile.UrlString, contentFile.PuzzleID}).IsUnique();
```
morganbr: This should be non-unique since we can have multiple puzzles/solutions in an event

Data/Migrations/PuzzleServerContextModelSnapshot.cs
```diff
@@ -285,6 +285,11 @@ protected override void BuildModel(ModelBuilder modelBuilder)
 
                     b.HasIndex("PuzzleID");
 
+                    b.HasIndex("EventID", "FileType")
```
morganbr: You'll need to regenerate the 3 migration files after your fixes to the model

Data/PuzzleServerContext.cs
```diff
@@ -82,6 +88,9 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             modelBuilder.Entity<SinglePlayerPuzzleStatePerPlayer>().HasIndex(pspt => new { pspt.PlayerID, pspt.SolvedTime });
             modelBuilder.Entity<SinglePlayerPuzzleUnlockState>().HasKey(state => new { state.PuzzleID });
             modelBuilder.Entity<Room>().HasIndex(room => new { room.EventID, room.Building, room.Number }).IsUnique();
+            modelBuilder.Entity<PlayerInEvent>().HasIndex(pie => new { pie.PlayerId, pie.EventId});
+            modelBuilder.Entity<PuzzleUser>().HasIndex(pu => new { pu.IdentityUserId});
+            modelBuilder.Entity<TeamMembers>().HasIndex(tm => new { tm.ID }); //EF can't parse this for nested types, so this is a placeholder to manually edit the generated model to INDEX Team.ID, User.ID, INCLUDE Class.ID, ID, TemporaryClassID
```
morganbr: It looks like the existing clustered index should be on this column. Does this have an effect?

morganbr (changes requested):

Data/PuzzleServerContext.cs
tabascq: I'd be lying if I said I truly knew how to read this line of code, but short names and URL strings are unique per event, correct?

morganbr (approved):

PR #1345, opened 2026-06-02
Fix help thread refresh bug

digit01Wave:
#1340

During the bugbash on June 1, a subset of players suddenly kept seeing a blank help thread page because the server was returning an error.

**Client side error:**
[2026-06-02T03:21:58.966Z] Error: Connection disconnected with error 'Error: Server returned an error on close: Connection closed with an error.'.
image

**Error from azure:**
A second operation was started on this context instance before a previous operation completed. This is usually caused by different threads concurrently using the same instance of DbContext. For more information on how to avoid threading issues with DbContext, see https://go.microsoft.com/fwlink/?linkid=2097913.
unhelpful, but:
System.InvalidOperationException:
   at Microsoft.EntityFrameworkCore.Infrastructure.Internal.ConcurrencyDetector.EnterCriticalSection (Microsoft.EntityFrameworkCore, Version=10.0.5.0, Culture=neutral, PublicKeyToken=adb9793829ddae60)
   at Microsoft.EntityFrameworkCore.Query.Internal.SingleQueryingEnumerable`1+AsyncEnumerator+<MoveNextAsync>d__20.MoveNext (Microsoft.EntityFrameworkCore.Relational, Version=10.0.5.0, Culture=neutral, PublicKeyToken=adb9793829ddae60)

Here is the explanation and fix from AI

The deployed failure is consistent with Blazor Server dbContext lifetime, not the SignalR payload shape anymore.
•	ThreadMessageDTO is already the full payload.
•	PuzzleThreadComponent already updates from the incoming DTO and no longer fetches the message from the database on receive.
•	PuzzleThreadService was still holding an injected scoped PuzzleServerContext.
•	In Blazor Server, scoped services live for the circuit, so overlapping callbacks can share the same dbContext.
•	Under load, one circuit can start another component/service operation before the prior EF query finishes, causing the ConcurrencyDetector.EnterCriticalSection exception.
•	An unhandled server-side exception can close the Blazor circuit, which matches the browser console disconnect.

What was changed:
•	PuzzleThreadService now injects IDbContextFactory<PuzzleServerContext> instead of a scoped PuzzleServerContext.
•	Each public service operation creates its own short-lived context with CreateDbContextAsync.
•	Helper methods now receive that context explicitly.
•	The transaction creation is now async via BeginTransactionAsync.
•	PuzzleThreadComponent now has a simple isMutationRunning guard to ignore overlapping send/edit/delete/claim clicks from the same component instance.
This is better than a lock as the long-term Blazor Server fix because:
•	It follows the recommended Blazor EF pattern.
•	It avoids serializing unrelated operations through a shared context.
•	It keeps each operation as its own EF unit of work.
•	It removes the underlying shared-dbContext concurrency hazard instead of masking it.
Validation:
•	Build completed successfully.

### DIFF
--- a/ServerCore/Pages/Threads/PuzzleThreadComponent.razor
+++ b/ServerCore/Pages/Threads/PuzzleThreadComponent.razor
@@ -15,10 +15,10 @@
                     {
                         <p>The more details you add here, the more helpful the responses you'll get will be!</p>
                     }
-                    <textarea @bind="NewMessageText" class="form-control message-subtitle" rows="4" style="width:75%" placeholder="Type message here" autocomplete="off"></textarea>
+                    <textarea @bind="NewMessageText" class="form-control message-subtitle" rows="4" style="width:75%" placeholder="Type message here" autocomplete="off" disabled="@isMutationRunning"></textarea>
                 </div>
                 <div class="form-group message-subtitle">
-                    <input type="button" value="Send" class="btn btn-default" @onclick="SendMessageAsync" />
+                    <input type="button" value="Send" class="btn btn-default" @onclick="SendMessageAsync" disabled="@isMutationRunning" />
                 </div>
             </div>
         </div>
@@ -31,13 +31,13 @@
     @if (ShouldShowClaimAction())
     {
         <br />
-        <button type="button" @onclick="ClaimLatestMessageAsync">Mark as read</button>
+        <button type="button" @onclick="ClaimLatestMessageAsync" disabled="@isMutationRunning">Mark as read</button>
     }
     else if (ShouldShowUnclaimAction())
     {
         <br />
         <p>Question has already been marked as read by @LatestMessage.ClaimerName</p>
-        <button type="button" @onclick="UnclaimLatestMessageAsync">Undo mark as read</button>
+        <button type="button" @onclick="UnclaimLatestMessageAsync" disabled="@isMutationRunning">Undo mark as read</button>
     }
 
     @if (!string.IsNullOrEmpty(errorMessage))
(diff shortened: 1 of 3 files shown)


ServerCore/Pages/Threads/PuzzleThreadComponent.razor.cs
```diff
@@ -250,15 +252,26 @@ await RunMutationAsync(async () =>
 
         async Task RunMutationAsync(Func<Task> mutation)
         {
+            if (isMutationRunning)
```
tabascq: Since this isMutationRunning code lacks a critical section or equivalent, it only works 100% of the time if it can be guaranteed to run in a single threaded environment. Do we have any assurance of that?

ServerCore/Pages/Threads/PuzzleThreadComponent.razor
```diff
@@ -15,10 +15,10 @@
                     {
                         <p>The more details you add here, the more helpful the responses you'll get will be!</p>
                     }
-                    <textarea @bind="NewMessageText" class="form-control message-subtitle" rows="4" style="width:75%" placeholder="Type message here" autocomplete="off"></textarea>
+                    <textarea @bind="NewMessageText" class="form-control message-subtitle" rows="4" style="width:75%" placeholder="Type message here" autocomplete="off" disabled="@isMutationRunning"></textarea>
```
tabascq, line 18: My understanding is that the C# code is running on the server since it's part of ServerCore, so updating based on these bools would require a message back to the client. Is a mutation such a long-running operation that it's worth sending a message back to the client saying to wait?

ServerCore/Pages/Threads/PuzzleThreadComponent.razor.cs
tabascq: Would it be better to just wait for the previous mutation to complete and then perform the action?

ServerCore/Pages/Threads/PuzzleThreadComponent.razor.cs
```diff
@@ -250,15 +252,26 @@ await RunMutationAsync(async () =>
...
+            {
+                errorMessage = "Please wait for the current action to finish.";
+                return;
+            }
+
             try
             {
+                isMutationRunning = true;
                 errorMessage = null;
                 await mutation();
```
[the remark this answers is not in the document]
digit01Wave: This is fine. It's not worth a second rerender

ServerCore/Pages/Threads/PuzzleThreadComponent.razor.cs
digit01Wave: fixed

ServerCore/Pages/Threads/PuzzleThreadComponent.razor
digit01Wave: fixed

tabascq: I don't know what "fixed" means here. If the operation is relatively short, then the isMutationRunning bool should be removed. If it's relatively long, please explain.

digit01Wave: Oh. I mean that on the razor.cs side, there is not longer a return from the server. Once someone pushes the send button it will await till it is done. This bool is there just in case the blazor does take a long time and yields the thread to other operations which will likely rerender the page with the new bool

tabascq (approved):

PR #1351, opened 2026-06-06
Archive Prototype

tabascq:
Creates a view of an event where all puzzles can be seen without needing to register.

This view is only available when the event is over.

The view offers a restricted set of features and links.

The hope is that a website downloader can be configured to quickly turn this view into a static event archive, but we'll need to deploy in order to test that and polish.

Future work in this area can provide unlocking and answer submission driven by localStorage, so that static archived events are actually playable.

### DIFF
--- a/ServerCore/Areas/Identity/UserAuthorizationPolicy/AuthorizationHelper.cs
+++ b/ServerCore/Areas/Identity/UserAuthorizationPolicy/AuthorizationHelper.cs
@@ -154,13 +154,23 @@ namespace ServerCore.Areas.Identity
         public async Task IsPlayerRegisteredForEvent(AuthorizationHandlerContext authContext, IAuthorizationRequirement requirement)
         {
             EventRole role = GetEventRoleFromRoute();
+            Event thisEvent = await GetEventFromRoute();
+
+            if (role == EventRole.archive)
+            {
+                if (thisEvent.CanArchive)
+                {
+                    authContext.Succeed(requirement);
+                }
+                return;
+            }
+
             if (role != EventRole.play)
             {
                 return;
             }
 
             PuzzleUser puzzleUser = await PuzzleUser.GetPuzzleUserForCurrentUser(dbContext, authContext.User, userManager);
-            Event thisEvent = await GetEventFromRoute();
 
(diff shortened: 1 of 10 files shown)


ServerCore/Pages/Submissions/Index.cshtml.cs
```diff
@@ -345,7 +345,25 @@ private async Task SetupContext(int puzzleId)
 
             IsPuzzleForSinglePlayer = Puzzle.IsForSinglePlayer;
             List<SubmissionView> submissionViews = new List<SubmissionView>();
-            if (Puzzle.IsForSinglePlayer)
+            if (EventRole == EventRole.archive)
+            {
+                PuzzleState = new PuzzleStatePerTeam();
```
digit01Wave, line 352: This only gets the team submissions. Do we care about the singlePlayerPuzzleSubmissions?

ServerCore/Areas/Identity/UserAuthorizationPolicy/AuthorizationHelper.cs
```diff
@@ -220,11 +230,22 @@ public async Task IsPlayerOnTeamCheck(AuthorizationHandlerContext authContext, I
 
         public async Task PlayerCanSeePuzzleCheck(AuthorizationHandlerContext authContext, IAuthorizationRequirement requirement)
         {
+            EventRole role = GetEventRoleFromRoute();
+            Event thisEvent = await GetEventFromRoute();
```
digit01Wave, line 234: [nit]: Can this be lazily evaluated so that we only actually await this when we need it?

digit01Wave (approved):

ServerCore/Areas/Identity/UserAuthorizationPolicy/AuthorizationHelper.cs
tabascq: The only case that does not need both of these is the non-Archive non-PuzzleUser case that returns on line 246, which is rare enough that I don't think we need to optimize for it.

ServerCore/Pages/Submissions/Index.cshtml.cs
tabascq: This actually works for both; I am using this because the base class has no public constructor. I'll add a comment to state this, for clarity.

digit01Wave (approved):

PR #1365, opened 2026-06-22
Fix #1360

tabascq:
PresenceComponent has a ReadOnly flag that nobody is using, but looks safe. But since it isn't being called yet, please give that a review too.

### DIFF
--- a/ServerCore/Pages/Submissions/Index.cshtml
+++ b/ServerCore/Pages/Submissions/Index.cshtml
@@ -423,7 +423,7 @@
                         }
                         @if (!Model.IsPuzzleForSinglePlayer && !Model.Event.EphemeralHackKillPresence)
                         {
-                            <component type="typeof(ServerCore.Pages.Components.PresenceComponent)" render-mode="Server" param-PuzzleUserId="@Model.LoggedInUser.ID" param-TeamId="@Model.Team.ID" param-PuzzleId="@Model.Puzzle.ID" />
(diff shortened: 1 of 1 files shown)


ServerCore/Pages/Submissions/Index.cshtml
```diff
@@ -423,7 +423,7 @@
                         }
                         @if (!Model.IsPuzzleForSinglePlayer && !Model.Event.EphemeralHackKillPresence)
                         {
-                            <component type="typeof(ServerCore.Pages.Components.PresenceComponent)" render-mode="Server" param-PuzzleUserId="@Model.LoggedInUser.ID" param-TeamId="@Model.Team.ID" param-PuzzleId="@Model.Puzzle.ID" />
+                        <component type="typeof(ServerCore.Pages.Components.PresenceComponent)" render-mode="Server" param-PuzzleUserId="@Model.LoggedInUser.ID" param-TeamId="@Model.Team.ID" param-PuzzleId="@Model.Puzzle.ID" param-ReadOnly="@(Model.EventRole != ModelBases.EventRole.play)"  />
```
morganbr: The parameter is IsReadOnly (not ReadOnly). The code in the presence component looks correct, but you should be able to verify by checking the presence view in the corner.

morganbr (changes requested):

ServerCore/Pages/Submissions/Index.cshtml
tabascq: Oops :-) Good catch, thanks! I actually ran it this time and it appears to work.

digit01Wave (approved):

digit01Wave (approved):

morganbr (approved):

